PR #3, opened 2023-09-20
add_effect refactor

ncontinanza:
This PR is a part of https://github.com/lambdaclass/curse_of_myrra/issues/802
Only resets effect countdown if both effects were caused by the same attacking player


AminArria (approved):
LGTM

AminArria (approved):
LGTM, remember to first merge the PR in the engine

PR #4, opened 2023-09-20
Uma Xandas Mark

martinacantaro:
closes https://github.com/lambdaclass/curse_of_myrra/issues/804
along with https://github.com/lambdaclass/curse_of_myrra/pull/913


AminArria (approved):
LGTM

PR #7, opened 2023-09-21
A player can't die because of the Yugen's Mark

ncontinanza:
Ditto.

Partially closes https://github.com/lambdaclass/curse_of_myrra/issues/803


AminArria (approved):
LGTM, but better if you rebase from `main` so CI runs

AminArria (approved):

PR #10, opened 2023-09-27
Wrap myrra state in the engine state and allow for adding and moving players

AminArria:
- In the engine `GameState` we provide `myrra_state` to contain the state that was currently produced by the Myrra part of the code
- Added `add_player/2` function to allow adding a player to the game state by providing the character name. If an inexistent character is sent the function does nothing
- Added `move_player/3` function which will move a player using their speed and direction (angle in degrees) 
- **Important:** this PR starts the foundation for `Position` to be based on coordinates in a 2D plane with (0, 0) as the center


ncontinanza (approved):
LGTM!

PR #13, opened 2023-10-05
Fix Uma marks bug

ncontinanza:
This PR fixes the bug due to which, if a character had Yugen's and/or Xanda's marks applied, no damage greater than its current health was received. This meant that not only Yugen's and Xanda's mark couldn't kill the character (which is correct), but any other form of attack couldn't either (which was the bug).


AminArria (approved):
Code looks good, but didn't playtest the changes

PR #14, opened 2023-10-05
H4ck basic multishot

ncontinanza:
Things to test:

- H4ck's basic now shoots 3 projectiles
- Multishot is no longer available
- Skills reorder


AminArria (approved):
LGTM

PR #20, opened 2023-10-20
Change spawn bot communication

agustinesco:
The bot where spawning with a wrong character, it should be a random character from the defined in the characters


AminArria (approved):
LGTM

PR #21, opened 2023-10-23
Keep track of a player exiting a match

ncontinanza:
Ditto.
State tracking related to https://github.com/lambdaclass/curse_of_myrra/pull/1003


AminArria (changes requested):
This should not go in the engine, the engine has no notion of connection and exiting a game

PR #24, opened 2023-10-25
Add attacked player attribute to projectiles and add piercing feature

manucamejo:
- Projectiles now pierce.
- Projectiles store player ids who have attacked to prevent them from attacking again.

### DIFF
--- a/docs/configuration/projectiles.md
+++ b/docs/configuration/projectiles.md
@@ -7,7 +7,7 @@ Configurable fields:
 - `base_damage`: Damage done by the projectile on collision
 - `base_speed`: Travel speed of the projectile
 - `base_size`: Size of the projectile for collision math
-- `pierce`: Determines if the projectile is removed from game after colliding with a player, default is `true`
+- `remove_on_collision`: Determines if the projectile is removed from game after colliding with a player, default is `true`
 - `on_hit_effects`: Effects given to target on collision
 - `duration_ms`: Defines how long in milliseconds the projectile can exist
 - `max_distance`: Defines the maximum distance the projectile can travel
@@ -25,7 +25,7 @@ Some example configurations
     "base_damage": 10,
     "base_speed": 123,
     "base_size": 50,
-    "pierce": false,
+    "remove_on_collision": true,
     "on_hit_effects": []
   },
   {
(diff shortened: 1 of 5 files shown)


docs/configuration/projectiles.md
```diff
@@ -7,7 +7,7 @@ Configurable fields:
 - `base_damage`: Damage done by the projectile on collision
 - `base_speed`: Travel speed of the projectile
 - `base_size`: Size of the projectile for collision math
-- `player_collision`: Determines if the projectile is removed from game after colliding with a player, default is `true`
+- `pierce`: Determines if the projectile is removed from game after colliding with a player, default is `true`
```
AminArria, line 10: I have issues with this naming

1. `pierce` in a context of games can mean a lot of different things (e.g. pierce damage), I feel `collision` or something like it is immediately clear
2. The explanation alone is not clear, what happens if I set this to `true`?

native/lambda_game_engine/src/game.rs
```diff
@@ -337,24 +337,19 @@ fn apply_projectiles_collisions(
...
                 if player.id == projectile.player_id {
                     continue;
                 }
 
                 player.decrease_health(projectile.damage);
                 player.apply_effects(&projectile.on_hit_effects);
+                projectile.attacked_player_ids.push(player.id);
 
-                projectile.active = false;
-                break;
```
AminArria, line 349: I think we need this `break` otherwise the projectile will collide with multiple players if they are in collision range at the same time

(PS: this is more a guess on my side, we should test it out, I just didn't have the time right now)

AminArria (changes requested):

native/lambda_game_engine/src/game.rs
```diff
@@ -337,24 +337,19 @@ fn apply_projectiles_collisions(
                     projectile.size,
                     player.size,
                 )
+                && !projectile.attacked_player_ids.contains(&player.id)
```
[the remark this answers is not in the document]
manucamejo: Yeah sure!

native/lambda_game_engine/src/game.rs
manucamejo: You're right!

docs/configuration/projectiles.md
manucamejo: Renamed to `remove_on_collision`

AminArria (approved):
LGTM

PR #31, opened 2023-11-03
Zone improvements

AminArria:
- Zone effect is only applied once, before it was applied on each zone tick (😱 )
- Zone center is selected at random, before it was hardcoded to `(0, 0)`

### DIFF
--- a/native/lambda_game_engine/src/game.rs
+++ b/native/lambda_game_engine/src/game.rs
@@ -130,6 +130,8 @@ impl GameState {
     pub fn new(config: Config) -> Self {
         let zone_radius = config.game.zone_starting_radius;
         let zone_modifications = config.game.zone_modifications.clone();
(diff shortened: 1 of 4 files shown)


ncontinanza:
Looking good! Just a detail I found out while testing on the client, depending on the value of the random center it may happen that when the match starts there's a small region of the map where the damage zone is active.

image

ncontinanza (changes requested):
Looking good! Just a detail I found out while testing on the client, depending on the value of the random center it may happen that when the match starts there's a small region of the map where the damage zone is active. We should change the initial values or the radius in order to make sure that the whole map is always clear at the beginning.

image

AminArria:
@ncontinanza I've made the corresponding change in CoM due to config coming from there now

ncontinanza (approved):

PR #36, opened 2023-11-13
Set player as dead when killed by the zone

ncontinanza:
We weren't changing the status from Alive to Dead when a player died because of the zone damage.

### DIFF
--- a/native/lambda_game_engine/src/game.rs
+++ b/native/lambda_game_engine/src/game.rs
@@ -530,6 +530,7 @@ fn apply_zone_effects(
         });
 
         outside_players.into_iter().for_each(|player| {
+            // If the player is alive in this tick, but dies as an effect of the zone, we push it to the killfeed.
             if player.status == PlayerStatus::Alive {
                 player.apply_effects_if_not_present(
                     &current_modification.outside_radius_effects,
--- a/native/lambda_game_engine/src/game.rs
+++ b/native/lambda_game_engine/src/game.rs
@@ -304,7 +304,7 @@ impl GameState {
         remove_expired_effects(&mut self.players);
         run_effects(&mut self.players, time_diff, &mut self.next_killfeed);
         modify_zone(&mut self.zone, time_diff);
-        apply_zone_effects(&mut self.players, &self.zone);
+        apply_zone_effects(&mut self.players, &self.zone, &mut self.next_killfeed);
 
         self.killfeed = self.next_killfeed.clone();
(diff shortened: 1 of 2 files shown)


native/lambda_game_engine/src/game.rs
```diff
@@ -525,10 +530,18 @@ fn apply_zone_effects(players: &mut HashMap<u64, Player>, zone: &Zone) {
         });
 
         outside_players.into_iter().for_each(|player| {
-            player.apply_effects_if_not_present(
-                &current_modification.outside_radius_effects,
-                EntityOwner::Zone,
-            );
+            if player.status == PlayerStatus::Alive {
```
Nico-Sanchez, line 533: Is this if necessary? We weren't using it before. I think you can just use the condition where the player died after this update (just leave the next if instead of these two).

native/lambda_game_engine/src/game.rs
```diff
@@ -525,10 +530,18 @@ fn apply_zone_effects(players: &mut HashMap<u64, Player>, zone: &Zone) {
...
+                player.apply_effects_if_not_present(
```
Nico-Sanchez, line 534: I saw you added a `update_status` function to change the status of a player as dead.
However that logic were being used in `decrease_health`. Couldn't we just call to that function to decrease the player HP?

I think updating a player status everywhere could lead to problems in the future where you have to maintain that behavior in every line of code where you touch the HP.
If that's not possible, maybe we can just update its status here, right before the `status == Death` if.

Nico-Sanchez (changes requested):

native/lambda_game_engine/src/game.rs
ncontinanza: It's necessary because we want to push to the killfeed only in the case in which the player died because of the effect of the zone. So we need to check if it's alive first and only push to the killfeed if the player was killed right then. Otherwise, we would be pushing to the killfeed a lot of times unnecessarily.

native/lambda_game_engine/src/game.rs
ncontinanza: Yeah, we noticed that with @manucamejo while doing this, and agreed to leave it as part of a refactor in the near future.

native/lambda_game_engine/src/game.rs
ncontinanza: I just pushed a comment to clarify this logic

Nico-Sanchez (approved):

PR #37, opened 2023-11-14
Auto aim attacks

### LINKED ISSUE #596: [FEATURE] Match quests
- [ ]  Same as daily quests but specific for match
- [ ]  User chooses only 1 match quest per match

manucamejo:
Closes #596 

## Motivation

We need to auto-attack with all skills if the attack button is tapped.

## Summary of changes

- Added a boolean that helps us to know if you are targeting or auto-attacking.
- Look for the nearest players to attack, and attack in that direction.

## How has this been tested?

Use some bots, and you will auto-attack in the direction of your opponents.

### DIFF
--- a/native/lambda_game_engine/src/game.rs
+++ b/native/lambda_game_engine/src/game.rs
@@ -544,3 +544,29 @@ fn apply_zone_effects(players: &mut HashMap<u64, Player>, zone: &Zone) {
         });
     }
 }
+
+fn nearest_player(
+    players: &Vec<&mut Player>,
+    position: &Position,
+) -> Option<(u64, Position)> {
+    let mut nearest_player = None;
+    let mut nearest_distance = 3000.0;
+
+    for player in players {
+        if matches!(player.status, PlayerStatus::Alive) {
+            let distance = distance_to_center(player, position);
+            if distance < nearest_distance
+            {
+                nearest_player = Some((player.id, player.position));
+                nearest_distance = distance;
+            }
+        }
+    }
+    nearest_player
+}
+
+fn distance_to_center(player: &Player, center: &Position) -> f32 {
+    let distance_squared =
+        (player.position.x - center.x).pow(2) + (player.position.y - center.y).pow(2);
+    (distance_squared as f32).sqrt()
+}
--- a/native/lambda_game_engine/src/game.rs
+++ b/native/lambda_game_engine/src/game.rs
@@ -191,10 +191,28 @@ impl GameState {
             if let Some(skill) = player.character.clone().skills.get(&skill_key) {
                 player.add_action(Action::UsingSkill(skill_key), skill.execution_duration_ms);
 
-                let direction_angle = skill_params
-                    .get("direction_angle")
-                    .map(|angle_str| angle_str.parse::<f32>().unwrap())
+                let auto_aim = skill_params
+                    .get("auto_aim")
+                    .map(|auto_aim_str| auto_aim_str.parse::<bool>().unwrap())
                     .unwrap();
+
+                let direction_angle = if auto_aim {
+                    let nearest_player: Option<(u64, Position)> = nearest_player(
+                        &other_players,
+                        &player.position
+                    );
+
+                    if let Some((_target_player_id, target_player_position)) = nearest_player {
+                        map::angle_between_positions(&player.position, &target_player_position)
+                    } else {
+                        player.direction
+                    }
+                } else {
+                    skill_params
+                        .get("direction_angle")
+                        .map(|angle_str| angle_str.parse::<f32>().unwrap())
+                        .unwrap()
+                };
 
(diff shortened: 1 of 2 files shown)


native/lambda_game_engine/src/game.rs
```diff
@@ -540,3 +558,29 @@ fn apply_zone_effects(
         });
     }
 }
+
+fn nearest_player(
+    players: &Vec<&mut Player>,
+    position: &Position,
+) -> Option<(u64, Position)> {
+    let mut nearest_player = None;
+    let mut nearest_distance = 3000.0;
```
mFragaBA, line 567: can we tweak this down a bit? When playtesting, I noticed that it attacks in a direction even when there are no enemies within the newport

native/lambda_game_engine/src/game.rs
```diff
@@ -540,3 +558,29 @@ fn apply_zone_effects(
...
+        }
+    }
+    nearest_player
+}
+
+fn distance_to_center(player: &Player, center: &Position) -> f32 {
+    let distance_squared =
+        (player.position.x - center.x).pow(2) + (player.position.y - center.y).pow(2);
+    (distance_squared as f32).sqrt()
+}
```
mFragaBA, line 586: perhaps we should move this to `map.rs` since that file contains many of the collision functions already. What do you think?

mFragaBA (commented):
Tested locally and worked as expected. I left two small comments but we should be good to merge

native/lambda_game_engine/src/game.rs
manucamejo: This attribute is now configurable

native/lambda_game_engine/src/game.rs
manucamejo: Sure! Done

native/lambda_game_engine/src/game.rs
```diff
@@ -562,25 +566,19 @@ fn apply_zone_effects(
 fn nearest_player(
     players: &Vec<&mut Player>,
     position: &Position,
+    auto_aim_max_distance: f32,
```
mFragaBA, line 569: I think we can reuse the function for other purposes so `max_search_distance` or `max_lookup_distance` might be a better name

```suggestion
    max_search_distance: f32,
```

native/lambda_game_engine/src/game.rs
```diff
@@ -562,25 +566,19 @@ fn apply_zone_effects(
...
 ) -> Option<(u64, Position)> {
     let mut nearest_player = None;
-    let mut nearest_distance = 3000.0;
+    let mut nearest_distance = auto_aim_max_distance;
```
mFragaBA, line 572:
```suggestion
    let mut nearest_distance = max_search_distance;
```

mFragaBA (dismissed):
LGTM!

docs/configuration/game.md
```diff
@@ -9,6 +9,7 @@ This is the meta configuration for the game, not for a particular thing in the g
 - `zone_starting_radius`: Radius of the playable zone. The zone is an area in the map where effects are applied to players not in it
 - `zone_modifications`: This is attribute is a list of modifications to perform to the playable zone in the map. For the specifics see below ["Zone modification"](#zone-modification)
 - `loot_interval_ms`: If present, interval in milliseconds for spawning loot crates
+- `auto_aim_max_distance`: Determines the max distance to auto aim
```
AminArria, line 12: Shouldn't this be a per character config? Even more, this should be a per skill config

AminArria, line 12: Shouldn't this be a per character config? Even more, this should be a per skill config

AminArria (changes requested):

native/lambda_game_engine/src/game.rs
```diff
@@ -540,3 +562,23 @@ fn apply_zone_effects(
         });
     }
 }
+
+fn nearest_player(
+    players: &Vec<&mut Player>,
+    position: &Position,
+    max_search_distance: f32,
+) -> Option<(u64, Position)> {
```
AminArria, line 570: What's the point of returning the `id`? I didn't see it used anywhere

native/lambda_game_engine/src/game.rs
```diff
@@ -540,3 +562,23 @@ fn apply_zone_effects(
         });
     }
 }
+
+fn nearest_player(
```
AminArria, line 566: nit: This is returning a `Position` so might make more sense to name `nearest_player_position`

AminArria (changes requested):

docs/configuration/game.md
manucamejo: We discuss with @saugon about this, and we think that it should be a global config.

AminArria (approved):

PR #38, opened 2023-11-15
Player item inventory

AminArria:
- Adds the concept of an inventory to a player
- Adds a config for `max_inventory_size` of a Character
- Adds `pickup_mechanic` for Loot to define how the pickup should work
- Adds function to use an item in the inventory

### DIFF
--- a/native/lambda_game_engine/src/game.rs
+++ b/native/lambda_game_engine/src/game.rs
@@ -9,6 +9,7 @@ use crate::config::Config;
 use crate::effect;
 use crate::effect::Effect;
 use crate::loot::Loot;
+use crate::loot::PickupMechanic;
 use crate::map;
 use crate::map::Position;
 use crate::player::Action;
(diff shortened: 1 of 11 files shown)


native/lambda_game_engine/src/player.rs
```diff
@@ -272,6 +276,24 @@ impl Player {
...
+                self.inventory[slot_at] = Some(loot.clone());
+                true
+            }
+            None => false,
+        }
+    }
+
+    pub fn inventory_get_at(&mut self, inventory_at: usize) -> Option<Loot> {
+        self.inventory[inventory_at].take()
+    }
```
mFragaBA, line 296: I would expect a `X_get_at` function to get the value and do nothing with self. Maybe `inventory_take_at` is less confusing, what do you think?

native/lambda_game_engine/src/game.rs
```diff
@@ -338,10 +347,15 @@ fn get_next_id(next_id: &mut u64) -> u64 {
...
-            loot.effects
-                .iter()
-                .for_each(|effect| player.apply_effect(effect, EntityOwner::Loot));
-            false
+            match loot.pickup_mechanic {
+                PickupMechanic::CollisionToInventory => player.put_in_inventory(loot),
+                PickupMechanic::CollisionUse => {
+                    loot.effects
+                        .iter()
+                        .for_each(|effect| player.apply_effect(effect, EntityOwner::Loot));
```
mFragaBA, line 355: there's an existing function named `apply_effects` in `player.rs` that does this same thing. Can we re-use it? (i can't add the proper indentation here)
```suggestion
                    player.apply_effects(&loot.effects, EntityOwner::Loot);
```

mFragaBA (changes requested):
Nice work on the inventory. If we eventually add equipable items I think we can extend the current implementation without greater issues. I left two small comments. If it's not too much work can we add some tests for the inventory functionallity?

native/lambda_game_engine/src/player.rs
AminArria: That makes a lot of sense, renamed in e2ca46f

native/lambda_game_engine/src/game.rs
AminArria: Missed that when moving this, fixed in [83220fb](https://github.com/lambdaclass/game_backend/pull/38/commits/83220fbf56cb7ca2365e3180f75a915afe0c858c)

mFragaBA (approved):
LGTM!

PR #39, opened 2023-11-16
Get fixed spawn positions from config and assign them to players

tkz00:
- Adds `initial_positons` to the config.json file.
- In Player constructor added `initial_position` parameter to specify where to spawn.
- Adds `initial_positons` to the GameConfig structs.
  - Had to make Position derive from Deserialize macro.
- In the `add_player` method get the corresponding position from the configuration depending on the number of players already in the game, and then pass that position to the constructor of Player as a parameter.


AminArria (changes requested):
Pre requesting changes as I don't want this merged right now, but don't have time to write full feedback until later

native/lambda_game_engine/src/game.rs
```diff
@@ -31,6 +31,7 @@ pub struct GameConfigFile {
     loot_interval_ms: u64,
     zone_starting_radius: u64,
     zone_modifications: Vec<ZoneModificationConfigFile>,
+    initial_positions: Vec<Position>,
```
AminArria, line 34: If you are adding a config field you need to add it to the documantion, probaby this one https://github.com/lambdaclass/lambda_backend_game_engine/blob/main/docs/configuration/game.md

native/lambda_game_engine/src/lib.rs
```diff
@@ -30,7 +30,8 @@ fn add_player(game: GameState, character_name: String) -> (GameState, Option<u64
     match game.config.find_character(character_name) {
         None => (game, None),
         Some(character_config) => {
-            let player = Player::new(player_id, character_config);
+            let initial_position = game.config.game.initial_positions[game.players.len()].clone();
```
AminArria, line 33: Two problems here
1. This will panic and crash the game if for some reason there are more players than whatever this config says, a crash should never happen 
2. This makes the initial positions static, no randomness on how they are assigned beyond luck on the order you join the game


Elaborating on point 1, I would rather see a `initial_positions` to be a list of points in the map with a radius, this would allow us to spawn in the same point, but slightly different, when repeating an initial position

AminArria (changes requested):
This PR should be branching from `main`, any reason not to?

mFragaBA:
> To work [thses changes](https://github.com/lambdaclass/curse_of_myrra/pull/1186) are also needed.

I'm guessing these does't count anymore

mFragaBA (approved):
LGTM!

agustinesco (approved):
LGTM!

PR #40, opened 2023-11-16
remove dirty scheduler

mFragaBA:
Based on load testing results from https://github.com/lambdaclass/curse_of_myrra/pull/1132, we found out that removing the dirty scheduler greatly improved times.


fkrause98:
Let's test the UX with these changes, also, we can use this a stop-gap on how to write benchmark and load-testing comparisons.

fkrause98 (dismissed):

manucamejo (approved):

fkrause98 (approved):

PR #42, opened 2023-11-21
Add give effect behavior to execute skills

Nico-Sanchez:
GiveEffect mechanic skill behavior was not implemented and it's a good way to implement the h4ck ultimate skill.
This PR adds a new parameter so an effect can trigger a skill as its effect.

## Changes

- [Add new parameter skills_to_execute to store all the skills that should be executed due to an effect](https://github.com/lambdaclass/game_backend/pull/42/commits/4bb43a7c6baa96240aa5f2bedd4a6ee5637c2ff5)
- [Add new activate_skills function to consume all skills that need to be executed due to an effect](https://github.com/lambdaclass/game_backend/pull/42/commits/968d9efa4bbe3d2be81a67733b1d90b4fd602fec)
- [Add new skill mechanic GiveEffect behavior](https://github.com/lambdaclass/game_backend/pull/42/commits/b1482b0c6a500445e576b80c7c6e4e2e57c1a606)

## How to test it?

- To test this PR you have to use the server (elixir) branch: gh-794-add-h4ck-ultimate-skill and use the h4ck 1st skill.


native/lambda_game_engine/src/effect.rs
```diff
@@ -8,6 +8,7 @@ pub struct Effect {
     pub effect_time_type: TimeType,
     pub player_attributes: Vec<AttributeChange>,
     pub projectile_attributes: Vec<AttributeChange>,
+    pub skills_to_execute: Vec<String>,
```
AminArria, line 11: This is a complicated one, doing this causes a circular dependency between `Effect` and `SkillConfig`

This was something that was heavily avoided to prevent invalid configs (e.g. skill uses effect that does not exist)

I'm not opposed to this because the idea makes sense, but I think we should rethink a bit on the config parsing to prevent this `String` usage so there are no dynamic finds at runtime

native/lambda_game_engine/src/skill.rs
```diff
@@ -46,7 +48,9 @@ pub enum SkillMechanicConfigFile {
 
 #[derive(NifTaggedEnum, Clone)]
 pub enum SkillMechanic {
-    GiveEffect(Vec<Effect>),
+    GiveEffect {
+        effects_to_give: Vec<Effect>,
```
AminArria, line 52: Why this change?

priv/test_config.json
AminArria, line 1: Please add an example using your new things

native/lambda_game_engine/src/player.rs
```diff
@@ -28,6 +28,7 @@ pub struct Player {
     pub size: u64,
     pub speed: u64,
     pub action_duration_ms: u64,
+    pub skills_to_execute: Vec<String>,
```
AminArria, line 31: Don't see why the player should have this, `Effect` already holds this information and when that effect triggers based on its `Periodic` setting they should be executed

native/lambda_game_engine/src/effect.rs
AminArria, line 11: Missing documentation update for this change

AminArria (changes requested):

native/lambda_game_engine/src/effect.rs
AminArria: Discussed this offline, this `String` is actually the skill_key used to identify the skill when activating, not the `Skill` unique name like in other part of the config so it was decided that all good due to being the same behaviour as activating a skill

native/lambda_game_engine/src/player.rs
AminArria: Discussed offline, this is an insane can of worms and no clear solution due to the need to have the `Game` modify itself when activating a skill, but when doing an effect on a player you don't have that information and passing it would break a lot of responsibility boundaries in the code

Long story short, we decided that this is a good middle ground until we figure a better approach

docs/configuration/effects.md
mFragaBA, line 1: can we add in the `examples` subsection one use case of using skill_keys_to_execute?

native/lambda_game_engine/src/game.rs
```diff
@@ -175,6 +175,40 @@ impl GameState {
...
+                        match mechanic {
+                            SkillMechanic::SimpleShoot {
+                                projectile: projectile_config,
+                            } => {
+                                let id = get_next_id(&mut self.next_id);
+
+                                let projectile = Projectile::new(
+                                    id,
+                                    player.position.clone(),
+                                    player.direction,
```
mFragaBA, line 198: how does this interact with the auto-aim feature?

native/lambda_game_engine/src/game.rs
```diff
@@ -175,6 +175,40 @@ impl GameState {
         }
     }
 
+    fn activate_skills(&mut self) {
```
mFragaBA, line 178: can't we reuse `activate_skill` for this function?

mFragaBA (dismissed):
remove the target directory

native/lambda_game_engine/src/game.rs
Nico-Sanchez: Hmmm I think we want it to aim to where the player is looking at. We could ask @manucamejo about this.

docs/configuration/effects.md
Nico-Sanchez: Sure, done!

native/lambda_game_engine/src/game.rs
Nico-Sanchez: I couldn't find a way to reuse that function. That function is called from the user's input (client) and I don't have some parameters. However, I agree we could do a refactor to reutilize some code of it. We can do it in a future PR. What do you think?

Nico-Sanchez:
> remove the target directory

Oops! Done!

native/lambda_game_engine/src/game.rs
mFragaBA: I talked with manu and as for now we decided to keep it like that and iterate in the future if we need it

native/lambda_game_engine/src/game.rs
mFragaBA: same here, it's better to do the refactor afterwards since we need this to implement uma

docs/configuration/effects.md
mFragaBA: Nice!

AminArria (approved):
LGTM, but unable to test in Unity at the moment

AminArria (approved):
Once again, LGTM, and I actually tested in Unity this time :D

PR #51, opened 2023-11-27
Fix non existent characters selections

tvillegas98:
This PR addresses this [issue#1201](https://github.com/lambdaclass/curse_of_myrra/issues/1201)

### DIFF
--- a/lib/dark_worlds_server/runner_supervisor/runner.ex
+++ b/lib/dark_worlds_server/runner_supervisor/runner.ex
@@ -100,13 +100,17 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
 
   @impl true
   def handle_call({:join, user_id, character_name}, _from, state) do
-    {game_state, player_id} = GameBackend.add_player(state.game_state, character_name)
+    case GameBackend.add_player(state.game_state, character_name) do
+      {:ok, {game_state, player_id}} ->
+        state =
+          Map.put(state, :game_state, game_state)
+          |> put_in([:user_to_player, user_id], player_id)
 
-    state =
-      Map.put(state, :game_state, game_state)
-      |> put_in([:user_to_player, user_id], player_id)
+        {:reply, {:ok, player_id}, state}
 
-    {:reply, {:ok, player_id}, state}
+      {:error, reason} ->
+        {:reply, {:error, reason}, state}
+    end
   end
 
   @impl true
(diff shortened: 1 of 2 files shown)


native/game_backend/src/lib.rs
```diff
@@ -22,17 +22,16 @@ fn parse_config(data: String) -> Config {
...
 }
-
 #[rustler::nif(schedule = "DirtyCpu")]
-fn add_player(game: GameState, character_name: String) -> (GameState, Option<u64>) {
+fn add_player(game: GameState, character_name: String) -> Result<(GameState, Option<u64>), String> {
     let mut game = game;
     let player_id = game.next_id();
     match game.config.find_character(character_name) {
-        None => (game, None),
+        None => Err("Character doesn't exists".to_string()),
```
AminArria, line 30: The returned error reason should be an atom, an enum in Rust land probably

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -102,14 +102,18 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
...
-    state =
-      Map.put(state, :game_state, game_state)
-      |> put_in([:user_to_player, user_id], player_id)
+        NewRelic.increment_custom_metric("GameBackend/TotalPlayers", 1)
+        {:reply, {:ok, player_id}, state}
 
-    NewRelic.increment_custom_metric("GameBackend/TotalPlayers", 1)
-    {:reply, {:ok, player_id}, state}
+      {:error, reason} ->
+        {:reply, {:error, reason}, state}
```
AminArria, line 115: What happens client side with this change?

AminArria (changes requested):

lib/dark_worlds_server/runner_supervisor/runner.ex
tvillegas98: Mmmm, if the character doesn't exists, the client loses the WebSocket connection and freezes on the "Prepare for Battle" screen 🤔 .

native/game_backend/src/lib.rs
tvillegas98: Sounds good, do we have an enum for it yet?

tvillegas98:
Moved to draft because we're missing the client fixes

AminArria (approved):

test/game_backend_test.exs
```diff
@@ -47,11 +47,11 @@ defmodule GameBackendTest do
     game = GameBackend.new_game(config)
 
     ## A character that does not exist does nothing and returns `nil`
```
agustinesco, line 49:
```suggestion
    ## A character that does not exist does nothing and returns `:character_not_found`
```

agustinesco (approved):
LGTM! tested on local and it's working as expected

PR #53, opened 2023-11-27
Add flame_on

fkrause98:
## Changes:
- Add flamegraphs from the phoenix dashboard.

## How to test:
1. Set the env variable:
   ```bash 
      export FLAMEGRAPH="true"
   ```
2. Recompile the project:
   ```bash 
    mix compile --force
   ```
3. Start the server:
  ```bash 
   iex -S mix phx.server
   ```
4. Go to: 
```
http://localhost:4000/dev/dashboard/flame_on
```
5. Choose a module + function with arity and start a game or load test or even call it from the elixir repl, you should see something like this:
image (2)


mix.exs
```diff
@@ -64,7 +64,9 @@ defmodule DarkWorldsServer.MixProject do
       {:cors_plug, "~> 3.0"},
       {:tesla, "~> 1.4", override: true},
       {:toxiproxy_ex, "~> 1.0.0"},
-      {:uuid, "~> 1.1"}
+      {:uuid, "~> 1.1"},
+      {:eflambe, "~> 0.3.0"},
```
AminArria, line 68: Why add `eflambe`?

AminArria:
Has this been tested as a mix release?

config/prod.exs
```diff
@@ -23,4 +23,5 @@ config :logger, level: :info
 # Newrelic agent
 config :new_relic_agent,
   app_name: System.get_env("NEWRELIC_APP_NAME"),
-  license_key: System.get_env("NEWRELIC_KEY")
+  license_key: System.get_env("NEWRELIC_KEY"),
+  disable_erlang_trace: if(System.get_env("FLAMEGRAPH"), do: true, else: false)
```
AminArria, line 27: What does this do?

mix.exs
fkrause98: I did not find a way to save the raw results of flame_on, but eflambe does have that option if we ever want it, i thought it would be useful, should I remove it?

fkrause98:
> Has this been tested as a mix release?

No, but does it make sense to do so so? After all, this flamegraph would be used for load testing, I don't think we want this functionality when actually releasing the server, right?

config/prod.exs
fkrause98: flame_on uses tracing to generate its output, new relic also uses tracing for some of its metrics, but there can only be one running tracing process, see this [issue](https://github.com/newrelic/elixir_agent/issues/137).

mix.exs
AminArria: No idea about the need to save them, but if you think is useful good by me

AminArria:
> No, but does it make sense to do so? After all, this flamegraph would be used for load testing, I don't think we want this functionality when actually releasing the server, right?

No idea how this is released to the load testing server, but if it is working then all good :D

AminArria (approved):

manucamejo:
> Has this been tested as a mix release?

@fkrause98 can you answer this question from @AminArria?

fkrause98:
@AminArria 

I just tested this with a release, you can double check it with:
```zsh
export FLAMEGRAPH="true"
export PHOENIX_HOST="localhost:4000"
rm -rf _build
mix release
 _build/dev/rel/dark_worlds_server/bin/dark_worlds_server start
```

PR #54, opened 2023-11-27
Bot server

AminArria:
- Introduce endpoint to initiate bots`POST /api/bot` which expects `game_id, bot_count, config`
- Create a `BotClientSupervisor` to dynamically create websocket clients from the bot server towards the game server
- Make `Runner` call the new bot endpoint to initiate bots instead of doing it locally
- Minor changes to `BotPlayer` to adapt to how the game state is received now and to going through `BotClient` instead of calling directly `Runner`
- Config to enable/disable bot endpoint and to configure the URLs for the game and bot servers


agustinesco (changes requested):
When i try to tests this pr i'm getting an error message `(MatchError) no match of right hand side value: {:error, :invalid_request}` this is being triggered when we try to make a post request to the new bot endpoint here `{:ok, %{status: 201}} = Tesla.post("#{bot_url}/api/bot", payload, headers: headers)`

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -265,7 +248,7 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
 
   @impl true
   def terminate(_reason, state) do
-    player_count = length(state.game_state.players) - state.bot_count
+    player_count = length(Map.values(state.game_state.players)) - state.bot_count
```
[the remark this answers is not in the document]
AminArria: Made the change

lib/dark_worlds_server/bot/bot_client.ex
```diff
@@ -0,0 +1,65 @@
...
+  alias DarkWorldsServer.Communication
+  alias DarkWorldsServer.Communication.Proto.GameEvent
+  alias DarkWorldsServer.RunnerSupervisor.BotPlayer
+  use WebSockex, restart: :temporary
+  require Logger
+
+  def start_link(%{game_id: game_id, config: config}) do
+    client_id = UUID.uuid4()
+    ws_url = Application.fetch_env!(:dark_worlds_server, DarkWorldsServer.Bot) |> Keyword.get(:game_server_url)
+    WebSockex.start_link("#{ws_url}/play/#{game_id}/#{client_id}/2", __MODULE__, %{config: config}, [])
```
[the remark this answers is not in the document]
AminArria: Why you say that? There should be one ws per bot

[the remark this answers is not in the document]
agustinesco: Oh sorry mb, i didn't noticed that you were creating a new `client_id` for every bot that you spawn

lib/dark_worlds_server_web/controllers/bot_controller.ex
```diff
@@ -0,0 +1,23 @@
+defmodule DarkWorldsServerWeb.BotController do
+  use DarkWorldsServerWeb, :controller
+
+  alias DarkWorldsServer.Bot.BotClientSupervisor
+
+  plug :check_bot_server_enabled
+
+  def create(conn, %{"game_id" => game_id, "bot_count" => bot_count, "config" => config}) do
```
fkrause98, line 8: I'm not quite sure how to test this, can I get some help?
For example, what is the config parameter, can it be an empty object?
Game id is an existing game?
For example, should this curl work?

```sh
curl
     --request POST 
    --url http://localhost:4000/ 
    --header 'content-type: application/json' 
    --data '{ game_id: 1, bot_count: 1, config: 1  }'
```

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -266,7 +248,7 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
 
   @impl true
   def terminate(_reason, state) do
-    player_count = length(state.game_state.players) - state.bot_count
+    player_count = Enum.count(state.game_state.players) - state.bot_count
```
fkrause98, line 251: nit: Why this change?

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -229,29 +228,12 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
   end
 
   def handle_info({:spawn_bots, bot_count}, state) when bot_count > 0 do
```
fkrause98, line 230: I think the idea is that bots are not some kind of special player but should behave like a regular player, do we need this handle_info for some special reason? 🤔

lib/dark_worlds_server/runner_supervisor/runner.ex
fkrause98: nvm, I just saw the previous comment.

lib/dark_worlds_server/runner_supervisor/runner.ex
AminArria: Now that you say it, this could go in `matching_coordinator.ex` so I'll move it

AminArria: Done in bb8dee8

lib/dark_worlds_server/runner_supervisor/runner.ex
AminArria: You are probably right, made the change as part of bb8dee8

lib/dark_worlds_server/matchmaking/matching_coordinator.ex
```diff
@@ -54,6 +55,7 @@ defmodule DarkWorldsServer.Matchmaking.MatchingCoordinator do
     bot_count = @session_player_amount - length(state.players)
     {:ok, game_pid, game_config} = start_game(bot_count)
```
agustinesco, line 56: Since this change will remove the responsibility of starting the bots from the runner i think we should remove this "bot_count" parameter from the runner initialization. 
This change implies changing:
`start_game/1` 
`RunnerSupervisor.start_child/1`
`Runner.init/1`

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -228,47 +223,14 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
...
-        {new_game_state, [player_id | bots]}
-      end)
-
-    Process.send_after(self(), {:activate_bots, bots_ids}, 10_000)
-
-    state =
-      Map.put(state, :game_state, game_state)
-      |> Map.put(:bot_handler_pid, bot_handler_pid)
-
-    NewRelic.increment_custom_metric("GameBackend/TotalBots", bot_count)
```
agustinesco, line 249: Don't we need this metric? maybe we could add it to the bot player init function

lib/dark_worlds_server/bot/bot_client.ex
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  @impl true
+  def handle_connect(_conn, state) do
+    Logger.info("Connected #{inspect(self())}")
```
agustinesco, line 16: Should we keep this?

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -228,47 +223,14 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
...
-    {game_state, bots_ids} =
-      Enum.reduce(0..(bot_count - 1), {state.game_state, []}, fn _, {acc_game_state, bots} ->
-        # character = Enum.random(["h4ck", "muflus"])
-        character = "muflus"
-        {new_game_state, player_id} = GameBackend.add_player(acc_game_state, character)
-
-        {new_game_state, [player_id | bots]}
-      end)
-
-    Process.send_after(self(), {:activate_bots, bots_ids}, 10_000)
```
agustinesco, line 243: Bots had this delay on the bots activation since the player won't be able to move for 10 seconds after we enter the match since we have the "Prepare for battle" screen + the camera cinematic, with this change we're activating the bots instantly, do you think we should fix this on this pr?

lib/dark_worlds_server/bot/bot_client.ex
```diff
@@ -0,0 +1,65 @@
...
+    {:reply, {:binary, Communication.player_move(angle)}, state}
+  end
+
+  def handle_info({:use_skill, angle, skill}, state) do
+    {:reply, {:binary, Communication.player_use_skill(skill, angle)}, state}
+  end
+
+  @impl true
+  def handle_disconnect(%{reason: {:local, reason}}, state) do
+    Logger.info("Local close with reason: #{inspect(reason)}")
```
agustinesco, line 37: Same, should we keep this?

agustinesco (changes requested):

lib/dark_worlds_server/bot/bot_client_supervisor.ex
```diff
@@ -0,0 +1,21 @@
...
+    DynamicSupervisor.start_link(__MODULE__, args, name: __MODULE__)
+  end
+
+  def start_bot_clients(game_id, bot_count, config) do
+    Enum.each(1..bot_count//1, fn _ ->
+      {:ok, _child_pid} = DynamicSupervisor.start_child(__MODULE__, {BotClient, %{game_id: game_id, config: config}})
+    end)
+
+    :ok
+  end
```
mFragaBA, line 15: some things about this:

- bots would now be counted as regular players (in newrelic metrics). Are we ok with that? I think it would be nice to be able to differentiate them
- we're never checking the capacity on the game (neither here nor in the runner).

lib/dark_worlds_server/bot/bot_client.ex
AminArria: Like the others, don't see any harm. Is not he most useful log, but is not useless and we should start thinking about logs in the app

lib/dark_worlds_server/runner_supervisor/runner.ex
AminArria: Need might be a strong word :rolf:, but I've added it back in `bot_client`

lib/dark_worlds_server/bot/bot_client_supervisor.ex
AminArria: > bots would now be counted as regular players (in newrelic metrics). Are we ok with that? I think it would be nice to be able to differentiate them

For good or bad this has been the intent from all this, and from a game perspective they are players. We will have a bot metric to have an idea, but maybe as a follow up we can add a param in the request to differentiate for telemetry purposes

AminArria: > we're never checking the capacity on the game (neither here nor in the runner).

In the good actors only case, `matching_coordinator` removed the need for that, it handles the capacity. 

In the bad actors case, in the future we should have a "secret" that only matched users know and thus will be the only ones allowed to connect to a game

lib/dark_worlds_server/bot/bot_client.ex
```diff
@@ -0,0 +1,72 @@
+defmodule DarkWorldsServer.Bot.BotClient do
+  alias DarkWorldsServer.Communication
+  alias DarkWorldsServer.Communication.Proto.GameEvent
+  alias DarkWorldsServer.RunnerSupervisor.BotPlayer
+  use WebSockex, restart: :temporary
+  require Logger
+
+  def start_link(%{game_id: game_id, config: config}) do
+    client_id = UUID.uuid4()
+    ws_url = Application.fetch_env!(:dark_worlds_server, DarkWorldsServer.Bot) |> Keyword.get(:game_server_url)
```
mFragaBA, line 10: question,can't the request itself contain the url of the game server to play games with?

mFragaBA: idk if that supposes a security issue

AminArria: This is a config thing, this way its not hardcoded and we have an env var to change it

lib/dark_worlds_server/bot/bot_client.ex
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  defp handle_msg(%{type: :GAME_FINISHED}, state) do
+    GenServer.stop(state.bot_pid)
+    {:close, state}
+  end
+
+  defp handle_msg(%{type: :PING_UPDATE}, state) do
+    {:ok, state}
+  end
```
ncontinanza, line 71:
```suggestion
  end
  
    defp handle_msg(%{type: :GAME_STARTED}, state) do
    {:ok, state}
  end

```

ncontinanza: We should handle this recently merged event in order to have the bots moving properly

ncontinanza (changes requested):

lib/dark_worlds_server/runner_supervisor/runner.ex
agustinesco: hahaha i'll rephrase to "should we keep this metric?"

lib/dark_worlds_server/bot/bot_client.ex
agustinesco: Sounds good!

lib/dark_worlds_server/bot/bot_client.ex
AminArria: Done in 73d7622

ncontinanza (approved):

manucamejo (approved):

PR #55, opened 2023-11-27
Space Hash Grid

fkrause98:


native/game_backend/src/player.rs
```diff
@@ -77,19 +80,36 @@ impl Player {
...
         self.add_action(Action::Moving, 0);
-        self.position = map::next_position(
+        let possible_next_position = map::next_position(
             &self.position,
             self.direction,
             self.speed as f32,
             config.game.width as f32,
         );
+
+        let ids_in_grid_for_player = grid.get_nearby(&GameEntity::from(&(self.clone())));
```
Oppen, line 97: Since `GameEntity::from` receives a reference, let's just do:
```suggestion
        let ids_in_grid_for_player = grid.get_nearby(&*self.clone());
```
Cloning `Player` each time may be too costly.

Oppen: (Rationale: there's at least 6 calls to `malloc` per `Player::clone` call)

native/game_backend/src/space_hash_grid.rs
```diff
@@ -0,0 +1,152 @@
+use crate::{map::Position, player::Player, projectile::{Projectile, self}};
+use rustler::NifMap;
+use std::collections::HashMap;
+
+#[derive(Clone, Debug, rustler::NifMap)]
```
Oppen, line 5: For this simple type, `Copy` may be quite handy.
```suggestion
#[derive(Clone, Copy, Debug, rustler::NifMap)]
```

native/game_backend/src/space_hash_grid.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+}
+
+#[derive(NifMap)]
+pub struct SpatialHashGrid {
+    cols: u64,
+    rows: u64,
+    bucket_total: u64,
+    buckets: HashMap<u64, Bucket>,
```
Oppen, line 65: This looks like you really want a `Vec<Bucket>`, from how you initialize it.

native/game_backend/src/space_hash_grid.rs
```diff
@@ -0,0 +1,152 @@
...
+            sceneheight,
+            cellsize,
+            scenewidth,
+        }
+    }
+    pub fn clear_buckets(&mut self) {
+        self.buckets = HashMap::new();
+        for i in 0..self.bucket_total {
+            self.buckets.insert(i, vec![]);
+        }
```
Oppen, line 94:
```suggestion
        self.buckets.values_mut().for_each(|b| b.clear());
```
Calling `clear` on each value saves reallocations later, as it reuses the old buffer.

Oppen: Alternatively, I think the following should work too:
```suggestion
        self.buckets = HashMap::new();
        for i in 0..self.bucket_total {
            self.buckets.values_mut().for_each(HashMap::clear);
        }
```

native/game_backend/src/space_hash_grid.rs
```diff
@@ -0,0 +1,152 @@
...
+            bucket.push(entity.clone());
+        }
+    }
+    pub fn add_walls_at(&mut self, x: i64) {
+        for y in -5000..5000 {
+           let wall_position = Vec2 { x: x as f32, y: y as f32};
+           let wall = GameEntity {is_blocking: true, position: wall_position, radius: 1 as f32};
+           self.register_entity(&wall)
+        }
+    }
```
Oppen, line 112: Maybe just enforce the hardcoded limits as logic rather than creating 20000 entities each tick.

Oppen: NOTE: first check it's relevant, of course.

PR #57, opened 2023-11-27
Adapt backend messages in the client

ncontinanza:


AminArria:
This PR is replaced by #105

PR #59, opened 2023-11-28
Make Muflus the only playable character

BertovDev:
This PR leaves Muflus as the only playable character as well for BOTS


ncontinanza (approved):

tkz00 (approved):
Great!

PR #60, opened 2023-11-28
[GH-1215] change integers to floats for position representations

mFragaBA:
closes [#1215](https://github.com/lambdaclass/curse_of_myrra/issues/1215)

# Motivation

We want to use floats to represent positions in both the frontend and the backend .

# Changes Made

Changed to float all fields related to:

- size (this includes the size of the characters, loots, projectiles, map size, map modifications values, etc.)
- position
- speed (both character and projectile speed)
- attack ranges

Also changed the config to divide by 100 all values related to distances. That is speeds, ranges and zone updates. This change is also replicated in the client. Before we had to do this because the game's board was 100x100 units in the client but 10000x10000 units in the backend. We couldn't use 100x100 in the backend with integers because the tiles would be too big and the movement would end up being too choppy. Now that we have floats we can just use 100x100 map size and have movement speeds with decimal amounts.

There's a small warning with this as using floats might incurr in precision errors or small desyncs between client and server ([not all computers represent and handle floats in a consistent way, and it might also vary based on the compiler used](https://grouper.ieee.org/groups/msc/ANSI_IEEE-Std-754-2019/background/addendum.html)), but for now unless we see playability issues we'll deem it non-relevant

# Steps to test this

You can use the branch `use-floats-for-backend-client-comm` from the [curse_of_myrra repo](https://github.com/lambdaclass/curse_of_myrra)

Check that the game still feels ok, particularly projectiles, hitboxes, movement and the zone


native/game_backend/src/map.rs
```diff
@@ -7,13 +7,13 @@ use crate::player::Player;
...
-    pub x: i64,
-    pub y: i64,
+    pub x: f32,
+    pub y: f32,
 }
 
 pub fn hit_boxes_collide(center1: &Position, center2: &Position, size1: u64, size2: u64) -> bool {
-    let squared_x = (center1.x - center2.x).pow(2) as f64;
-    let squared_y = (center1.y - center2.y).pow(2) as f64;
+    let squared_x = ((center1.x - center2.x) as f64).powi(2);
```
AminArria, line 15: Should this also change to f32?

fkrause98: Since this the exponentiation of an f32, it can turn into a number bigger than f32::max, so the f64 gives it a bit of head room.

lib/dark_worlds_server/runner_supervisor/bot_player.ex
agustinesco, line 1: I think we forgot to fix the numbers in the following functions:
`skill_would_hit/2`
`maybe_put_wandering_position`

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -266,7 +266,7 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
 
   @impl true
   def terminate(_reason, state) do
-    player_count = Enum.count(state.game_state.players) - state.bot_count
+    player_count = map_size(state.game_state.players) - state.bot_count
```
agustinesco, line 269: is map_size mor efficient than Enum.count?

native/game_backend/src/map.rs
```diff
@@ -22,20 +22,20 @@ pub fn hit_boxes_collide(center1: &Position, center2: &Position, size1: u64, siz
...
-    max_distance: u64,
+    max_distance: f32,
     cone_angle: f32,
 ) -> bool {
     // TODO: Take into consideration `size` attribute of Player
-    let squared_x = (center_player.position.x - target_player.position.x).pow(2) as f64;
-    let squared_y = (center_player.position.y - target_player.position.y).pow(2) as f64;
+    let squared_x = ((center_player.position.x - target_player.position.x) as f64).powi(2);
+    let squared_y = ((center_player.position.y - target_player.position.y) as f64).powi(2);
     let target_distance = (squared_x + squared_y).sqrt();
```
agustinesco, line 31: Do we still need this `as f64`?

agustinesco (changes requested):

lib/dark_worlds_server/runner_supervisor/runner.ex
mFragaBA: Yes! Enum.count goes though the whole collection whole map_size Is calculated in constant time

native/game_backend/src/map.rs
mFragaBA: Yes, since squaring the number might not Fit in an f32

ncontinanza (approved):
LGTM! Tested locally and on an Android device, everything worked fine.

PR #62, opened 2023-11-29
Circular map

manucamejo:
Now the map is circular.

### DIFF
--- a/native/game_backend/src/map.rs
+++ b/native/game_backend/src/map.rs
@@ -53,7 +53,9 @@ pub fn next_position(
     let angle_rad = direction_angle * (PI / 180.0);
     let new_x = (current_position.x as f32) + movement_amount * angle_rad.cos();
     let new_y = (current_position.y as f32) + movement_amount * angle_rad.sin();
-    let radius = width / 2.0;
+
+    // This is to avoid  the overflow on the front end
+    let radius = (width - 200.0) / 2.0;
 
     let center = Position { x: 0, y: 0 };
 
@@ -66,6 +68,7 @@ pub fn next_position(
         new_position
     } else {
         let angle = angle_between_positions(&center, &new_position) * (PI / 180.0);
+
         Position {
             x: (radius * angle.cos()) as i64,
             y: (radius * angle.sin()) as i64,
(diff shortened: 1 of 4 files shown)


native/game_backend/src/map.rs
AminArria, line 1: This will slightly break something in the bots

native/game_backend/src/map.rs
```diff
@@ -115,7 +125,11 @@ pub fn angle_between_positions(center: &Position, target: &Position) -> f32 {
 }
 
 pub fn distance_to_center(player: &Player, center: &Position) -> f32 {
+    distance_between_positions(&player.position, center)
+}
```
agustinesco, line 129: Isn't this function redundant?

agustinesco (changes requested):

native/game_backend/src/map.rs
```diff
@@ -96,14 +104,16 @@ pub fn collision_with_edge(center: &Position, size: u64, width: u64, height: u64
...
 }
 
-pub fn random_position(width: u64, height: u64) -> Position {
+pub fn random_position(width: u64, _height: u64) -> Position {
     let rng = &mut rand::thread_rng();
-    let bound_x = (width / 2) as i64;
-    let bound_y = (height / 2) as i64;
+
+    let radius = (width / 2) as i64;
+    let random_radius = (rng.gen_range(0..radius.pow(2)) as f32).sqrt();
```
mFragaBA, line 111: why do we need the pow(2) and sqrt? can't. we just use `rng.gen_range(0..radius)`?

native/game_backend/src/map.rs
manucamejo: Yes, when they are wandering. I think this is not a problem because they are going to find another place. Also, this is a temporary solution, we are implementing a generic collision solution.

mFragaBA (changes requested):
Tested locally and works as expected! There's one think we can avoid doing, and I think that if the map will be circular we wouldn't need a width/height, but we talked about this and we're ok with doing this for now as it will soon be replaced by using a "wall" once we have collisions

native/game_backend/src/map.rs
manucamejo: Because if not, we get a nonuniform distribution:
https://mathworld.wolfram.com/DiskPointPicking.html

agustinesco (approved):
LGTM! tested on local and worked as expected

mFragaBA (approved):
LGTM!

PR #64, opened 2023-11-29
Implement skill Leap

AminArria:
- [x] Automatically move on each game tick
- [x] Prevent other actions (move) while leaping
- [x] Do other mechanics when arriving
- [x] Prevent damage while leaping
- [x] Add targeting for the leap destination
- [x] Update `messages.proto` for new skill params for MoveToTarget
- [x] Update `runner` to handle new skill params for MoveToTarget mechanic

Extra
- [ ] PR to curse_of_myrra to update `messages.proto` and send MoveToTarget params


ncontinanza (dismissed):

lib/dark_worlds_server/runner_supervisor/runner.ex
```diff
@@ -80,7 +78,7 @@ defmodule DarkWorldsServer.RunnerSupervisor.Runner do
 
     state = %{
       game_state: GameBackend.new_game(game_config),
-      game_tick: @game_tick_rate_ms,
+      game_tick: game_config.game.tick_interval_ms,
```
agustinesco, line 81: Nice change!

agustinesco (approved):
Let's wait for the client pr

PR #75, opened 2023-12-01
Hotfix killfeed hit duration

manucamejo:


ncontinanza (approved):
LGTM!

native/game_backend/src/game.rs
```diff
@@ -336,7 +339,7 @@ impl GameState {
                                         target_player,
                                         *range,
                                         *cone_angle as f32,
-                                    )
+                                    ) && target_player.status == PlayerStatus::Alive
```
mFragaBA, line 342: we should do this filter for `other_players` at the beggining of the method. This way we also filter those from the autoaim calculation which does:

```rust
let direction_angle = if auto_aim {
    let nearest_player: Option<Position> = nearest_player_position(
        &other_players,
        &player.position,
        self.config.game.auto_aim_max_distance,
    );

    if let Some(target_player_position) = nearest_player {
        map::angle_between_positions(&player.position, &target_player_position)
    } else {
        player.direction
    }
}
```

mFragaBA: you can do something like:

```rust
let (mut player_in_list, mut other_players): (Vec<_>, Vec<_>) = players
             .values_mut()
             .filter(|player| player.status == PlayerStatus::Alive)
             .partition(|player| player.id == player_id);
```

mFragaBA (changes requested):

mFragaBA (approved):
LGTM! Tested locally and now:

- we don't hit nor kill already dead players
- we don't auto aim towards dead players
- the animations now look good

ncontinanza (approved):

PR #44, opened 2023-11-22
Add elixir code and actions

Nico-Sanchez:
We had our backend splitted in two repositories (one for Elixir, Rust code in the other). That separation isn't correct and leads to many problems that we've been dodging.
We'll move our Elixir code and our deployment actions here.

## Changes made

I strongly recommend reviewing this PR commit by commit because the amount of lines changed is huge.

- [Add Elixir mix project files (mix.exs + lib folder)](https://github.com/lambdaclass/game_backend/pull/44/commits/969bb1b8139a4ae6127f3d069f4399ea3669490f)
- [Add Elixir priv folder and files](https://github.com/lambdaclass/game_backend/pull/44/commits/e141b0c6bdef5879b18d7b5e267e96757114f1b8)
- [Add Elixir assets folder and files](https://github.com/lambdaclass/game_backend/pull/44/commits/72fe24c92a428e5012bc5adcfad19078a4df339e)
- [Add Elixir config folder and files](https://github.com/lambdaclass/game_backend/pull/44/commits/ade3cb1959ed8017e0320ce09749495cb677c9c0)
- [Add elixir make commands to makefile](https://github.com/lambdaclass/game_backend/pull/44/commits/e283ada2286ab9b71512bf503c240d84f54048ab)
- [Add docker compose file](https://github.com/lambdaclass/game_backend/pull/44/commits/d9fe52616952770376e555861bdeffacd08113dd)
- [Update formatter phoenix dependency](https://github.com/lambdaclass/game_backend/pull/44/commits/40dfa701fa2e1257da1820abdd2cb0a1ba4823df)
- [Update credo](https://github.com/lambdaclass/game_backend/pull/44/commits/6b8b770f34efd52153f25fe20f1533e9650abc6f)
- [Update CI](https://github.com/lambdaclass/game_backend/pull/44/commits/7a8ec614784b03f4ccf3cc7e5186f496a31df8a7)
- [Restore make credo statement](https://github.com/lambdaclass/game_backend/pull/44/commits/fa13969845c8012c060e1ed7b0a66f5adb289c0e)
- [Fix CI yaml](https://github.com/lambdaclass/game_backend/pull/44/commits/2217b5af64d10545d6eb6f773a2b989f31371a2c)
- [Add load_test folder and files](https://github.com/lambdaclass/game_backend/pull/44/commits/6c8aa5d0555b8747dc560a9b0d983a49e6658d63)
[...]

PR #80, opened 2023-12-04
add script with changes

mFragaBA:
The script was left out on #44. And since we use toxiproxy along the backend it should belong here


fkrause98 (approved):

agustinesco (approved):
LGTM! let's remember to delete this from the client repo

PR #83, opened 2023-12-05
Muflus hit delay

ncontinanza:
Currently, when a player attacked using Muflus, the opponent's life was immediately affected. 
With these changes, there's a brief delay (300ms) in order to discount life when the animation shows the opponent being hit.


saugon (approved):
Nice job!

PR #84, opened 2023-12-05
Add random modifier to bot behavior

agustinesco:
Now the bot_player will have a random number called "random_factor" this number will be used to determine certain modifiers on the bot behvaior.

The affected actions are:
- Added a little delay in the decision interval to simulate an user delay
- Fixed angle tilt in the attack
- The bot has a chance in four of the random_factor value to change his decision to wander
- The bot has a chance in eight of the random_factor value to change to nothing for a cicle


ncontinanza (approved):
LGTM!

PR #87, opened 2023-12-05
Fix bot attack range and add randomness to it

agustinesco:
The bots are walking on top of the enemies instead of attacking them when they're in range, this is due to two factors:
1. The :attack action is being updated every 500ms so if the bot started walking to an enemy he will continue to do so for 500 ms in the worst case scenario
2. The attack range is hardcoded in the `skill_would_hit/2` function, so we need to get the range value from the skills in the config file


ncontinanza (changes requested):
The code seems fine, but during the game play I felt that the bots are not aggresive at all, ignoring me and the other bots and almost never attacking.

AminArria:
@agustinesco same experience, the bots basically did not attack me

ncontinanza (approved):

manucamejo (approved):

PR #88, opened 2023-12-06
Add 2023-12-06 load testing report

mFragaBA:
## Changes:
- Add a report for today's load testing (Wed, 6 Dec 2023).


load_test/reports/2023_12_06/2023_12_06.md
```diff
@@ -0,0 +1,73 @@
...
+- 20 games of 10 players each (NUMBER_OF_USERS = 200), PLAY_TIME = 5min, updates every 30ms on `main` branch
+- 20 games of 10 players each (NUMBER_OF_USERS = 200), PLAY_TIME = 5min, updates every 120ms on `separate-movement-from-tick` branch
+
+## Test Results
+
+### 20 games of 10 players each, PLAY_TIME = 5min, updates every 30ms on `main` branch 
+
+#### Briefly describe the UX exprerience:
+Tested without any noticeable issues. It was a good exprerience overall.
+
```
agustinesco, line 56: I think a good mention here would also be the amount of fps/ping you experienced in the match

mFragaBA:
> LGTM! i have a doubt tho, why we're only testing with 200 players (the 20 matches) now? and not with the 400 / 500 hundred that we were testing before?

We were using a smaller server and it couldn't handle as many users. I should mention it tough since it's not clear. Nice catch!

PR #89, opened 2023-12-06
simulate player joining on lobby

agustinesco:
If you entered a lobby the player counter would remain in 1 and the jump instantly to 10, this is wrong and we should simulate player joining to the lobby while we wait

### DIFF
--- a/lib/dark_worlds_server/matchmaking/matching_coordinator.ex
+++ b/lib/dark_worlds_server/matchmaking/matching_coordinator.ex
@@ -28,7 +28,8 @@ defmodule DarkWorldsServer.Matchmaking.MatchingCoordinator do
   #######################
   @impl true
   def init(_args) do
-    {:ok, %{players: [], session: :unset}}
+    send(self(), :simulate_player_joined)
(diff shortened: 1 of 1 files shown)


lib/dark_worlds_server/matchmaking/matching_coordinator.ex
```diff
@@ -28,17 +28,19 @@ defmodule DarkWorldsServer.Matchmaking.MatchingCoordinator do
   #######################
   @impl true
   def init(_args) do
-    {:ok, %{players: [], session: :unset}}
+    send(self(), :simulate_player_joined)
+    {:ok, %{players: [], session: :unset, curent_lobby_players: 0}}
```
AminArria, line 32: Typo in `curent_lobby_players` it should be `current_lobby_players`

AminArria (changes requested):
One thing I'm failing to understand is why do we need this in the backend? If this is just to simulate a counter going up why can't the frontend/client do it?

agustinesco:
This is more a client change so we'll do it there

PR #90, opened 2023-12-06
Pick up items to the inventory and use them

ncontinanza:
Receive a new :use_inventory action in order to use an item in the inventory.
Should be tested together with: https://github.com/lambdaclass/curse_of_myrra/pull/1258
Based in the implementation of: https://github.com/lambdaclass/game_backend/pull/38.


agustinesco (approved):
LGTM! while testing this pr i think i've found a bug with the loots, when you heal with max hp you are getting more than your base health, causing the health bar to not be accurate with your current health, i'm not sure if this is intentional or if it's a bug, either way  i think we should fix it in another pr

lib/dark_worlds_server/bot/bot_client.ex
```diff
@@ -33,6 +33,10 @@ defmodule DarkWorldsServer.Bot.BotClient do
     {:reply, {:binary, Communication.player_use_skill(skill, angle)}, state}
   end
 
+  def handle_info({:use_inventory, inventory_at}, state) do
```
AminArria, line 36: Bot player are not sending this so they are not using items

AminArria (approved):
LGTM, just the small thing about bots, but not a blocking issue

lib/dark_worlds_server/bot/bot_client.ex
agustinesco: Yeah the bots are kind of broken since the loots behave differently from before, i think we could introduce another case to their decision list in another pr

agustinesco (approved):

PR #99, opened 2023-12-11
Name all dependencies and minor rust cleanup

edg-l:
I always like to run the following line:

`cargo clippy -- -D clippy::nursery -D clippy::all -D clippy::complexity -A clippy::use_self `

Which tends to find nice stuff, but its better to not add it too CI because some stuff is more opinionated.

- With this i found some places where one can use [fused multiply add](https://doc.rust-lang.org/std/primitive.f64.html#method.mul_add) which has less rounding errors and sometimes faster perf, i also found a method that can be const.
- Added a clippy allow because rustler generates names that don't follow conventions.
- Updated some dependencies.
- Made release use lto and minimum codegen units for maximum perfomance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ format:
 
 check:
 	mix credo --strict
-	cargo clippy --manifest-path native/game_backend/Cargo.toml -- -D warnings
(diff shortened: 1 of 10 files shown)


mFragaBA (approved):
LGTM! I also tested the game with these changes and worked without any issues

mFragaBA (approved):

AminArria (approved):
LGTM

PR #102, opened 2023-12-12
Add new damage tracker list

agustinesco:
We were having some borrowing issues with the damage application and the player list, so we'll split the damage on 2 steps

1. If a player is victim of a damaging effect (can be an effect or a hit) we'll store this in a list called `pending_damages` 
2. Every tick we traverse this list and apply the pending damages poping them outside of the list, we also apply the on hit effect here

To test this pr you should test it with a client in the main branch and check that the damage applications feels right

### DIFF
--- a/native/game_backend/src/effect.rs
+++ b/native/game_backend/src/effect.rs
@@ -1,7 +1,7 @@
 use rustler::{NifMap, NifTaggedEnum};
 use serde::Deserialize;
 
-#[derive(Deserialize, NifMap, Clone)]
+#[derive(Deserialize, NifMap, Clone, Debug)]
 pub struct Effect {
     pub name: String,
     pub is_reversable: bool,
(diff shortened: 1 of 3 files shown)


native/game_backend/src/effect.rs
```diff
@@ -1,7 +1,7 @@
 use rustler::{NifMap, NifTaggedEnum};
 use serde::Deserialize;
 
-#[derive(Deserialize, NifMap, Clone)]
+#[derive(Deserialize, NifMap, Clone, Debug)]
```
ncontinanza, line 4: Do we want to keep all these Debug derives, or were them just to be able to print these struct during the debugging process?

agustinesco: These Debug trait were added just for debugging while developing the list, i'll remove them since they're not necessary

ncontinanza (approved):

AminArria (approved):
LGTM

PR #108, opened 2023-12-14
Check if action is already in list before adding

AminArria:
This fixes a memory leak introduced in #97, it was causing the `next_actions` list to keep adding `Action::Moving` growing infinitely

### DIFF
--- a/native/game_backend/src/player.rs
+++ b/native/game_backend/src/player.rs
@@ -41,7 +41,7 @@ pub enum PlayerStatus {
     Death,
 }
 
-#[derive(NifTaggedEnum, Clone)]
+#[derive(NifTaggedEnum, Clone, PartialEq, Eq)]
 pub enum Action {
     Nothing,
     Moving,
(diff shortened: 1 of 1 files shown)


native/game_backend/src/player.rs
```diff
@@ -92,8 +92,10 @@ impl Player {
     }
 
     pub fn add_action(&mut self, action: Action, duration_ms: u64) {
-        self.next_actions.push(action);
-        self.action_duration_ms += duration_ms;
+        if !self.actions.contains(&action) {
+            self.next_actions.push(action);
+            self.action_duration_ms += duration_ms;
+        }
```
Oppen, line 98: This looks like we really want some kind of set instead.

AminArria: Interesting enough, due to the small size of this list a Vec might be more efficient

ncontinanza (approved):
LGTM!

native/game_backend/src/player.rs
AminArria: I'll work with @agustinesco to look into this in #64 as his reworking the actions in there

PR #109, opened 2023-12-14
Add Cargo workspace to root project

AminArria:
By having a Cargo workspace we can easily open the project with `VSCode` and rust-analyzer will work and find the Rust code


agustinesco (approved):
Nice!

PR #110, opened 2023-12-15
Burst mechanic

ncontinanza:


native/game_backend/src/player.rs
```diff
@@ -121,13 +126,25 @@ impl Player {
 
     pub fn add_cooldown(&mut self, skill_key: &String, cooldown_ms: u64) {
         self.cooldowns.insert(skill_key.to_string(), cooldown_ms);
+        self.available_burst_loads -= 1;
```
agustinesco, line 129: The available burst loads are something that only affect the basic attacks right? if that's the case maybe we should check that the skill key is the same as the `BASIC_SKILL_KEY` variable

agustinesco (changes requested):

native/game_backend/src/player.rs
```diff
@@ -342,6 +359,20 @@ impl Player {
...
+            && (can_perform_basic_attack(self) || can_perform_skill_attack(self, skill_key))
+    }
+}
+
+fn can_perform_basic_attack(player: &Player) -> bool {
+    player.cooldowns.contains_key(BASIC_SKILL_KEY) && player.available_burst_loads > 0
+}
+
+fn can_perform_skill_attack(player: &Player, skill_key: &String) -> bool {
+    !player.cooldowns.contains_key(BASIC_SKILL_KEY) && !player.cooldowns.contains_key(skill_key)
```
agustinesco, line 375: Why are we checking that the player doesn't have a cooldown for the basic_skill_key here?

native/game_backend/src/player.rs
ncontinanza: Right! I added a validation for this

native/game_backend/src/player.rs
ncontinanza: Actually that's unnecessary, I changed it to a check against the BASIC_SKILL_KEY constant

native/game_backend/src/player.rs
```diff
@@ -120,14 +125,30 @@ impl Player {
     }
 
     pub fn add_cooldown(&mut self, skill_key: &String, cooldown_ms: u64) {
-        self.cooldowns.insert(skill_key.to_string(), cooldown_ms);
+        self.cooldowns
+            .entry(skill_key.to_string())
+            .or_insert(cooldown_ms);
+        if self.cooldowns.contains_key(BASIC_SKILL_KEY) {
```
agustinesco, line 131: I think this won't stop us of depleting a charge when using other skill than the basic either, since it the case where we want to use a skill when the basic_skill is on cooldown is possible

Maybe something like this would do:

```suggestion
        if skill_key == BASIC_SKILL_KEY {
```

agustinesco (changes requested):

native/game_backend/src/player.rs
agustinesco: Nice!

native/game_backend/src/player.rs
```diff
@@ -342,6 +363,20 @@ impl Player {
...
+
+    pub fn can_perform_attack(&self, skill_key: &String) -> bool {
+        // Check if player is still performing an action or if skill is still on cooldown.
+        self.can_do_action()
+            && (can_perform_basic_attack(self) || can_perform_skill_attack(self, skill_key))
+    }
+}
+
+fn can_perform_basic_attack(player: &Player) -> bool {
+    player.cooldowns.contains_key(BASIC_SKILL_KEY) && player.available_burst_loads > 0
```
agustinesco, line 375: This is not working as expected, since we need to be able to perform a basic attack to this clause to be true `player.cooldowns.contains_key(BASIC_SKILL_KEY)` Replacing this with some check like the one from the previous comment would do

agustinesco (dismissed):
LGTM!

docs/src/configuration/characters.md
```diff
@@ -9,7 +9,7 @@ Configurable fields:
 - `base_size`: Size of the character for collision math
 - `base_health`: Base health of the character
 - `skills`: This is a map of integer (as string) to skills, the integers represent the id and ordering of skills for calling them
-- `max_inventory_size`: Maximum amount of loots that the character can hold in their inventory
```
AminArria, line 12: Don't delete this

native/game_backend/src/player.rs
```diff
@@ -15,6 +16,8 @@ use crate::loot::Loot;
 use crate::map;
 use crate::map::Position;
 
+const BASIC_SKILL_KEY: &str = "1";
```
AminArria, line 19: This is essentially hardcoding things to a name in the config, the main thing we need to avoid doing in the backend

docs/src/configuration/characters.md
```diff
@@ -9,7 +9,7 @@ Configurable fields:
...
+- `burst_loads`: Number of times the character's basic skill can be used in a row before going on cooldown
```
AminArria, line 12: This should not be a property of the character as they have no concepts of "basic skill"

This property should go in the skill

AminArria (changes requested):

docs/src/configuration/characters.md
Nico-Sanchez, line 12: We're still keeping the `:max_inventory_size` field in the config. Why are we deleting it from the docs?

Nico-Sanchez (commented):
This new feature is really nice! Congrats @ncontinanza 🚀 

Left a comment about the docs, then we merge it.

docs/src/configuration/characters.md
ncontinanza: I restored it 👍

native/game_backend/src/player.rs
ncontinanza: Right, I removed it as part of the responsibilities refactor

docs/src/configuration/characters.md
ncontinanza: I added the property to the skill and removed it from the character. Thanks for the suggestion, I think it's much more adequate and extensible to other skills now!

ncontinanza: It was a mistake, now it should be restored!

agustinesco (approved):
Great job! LGTM

Nico-Sanchez (approved):
Nice corrections, looks much better now 🥳

PR #113, opened 2023-12-18
Implement action tracker instead of action list

agustinesco:
This pr aims to fix an existing issue in main that makes everyone to looks like they're walking in the same place. If you turn off the client prediction you are the one walking in the same place

### DIFF
--- a/native/game_backend/src/player.rs
+++ b/native/game_backend/src/player.rs
@@ -97,7 +97,11 @@ impl Player {
     }
 
     pub fn add_action(&mut self, action: Action, duration_ms: u64) {
-        if !self.action.iter().any(|action_tracker| action_tracker.action == action) {
+        if !self
+            .action
+            .iter()
+            .any(|action_tracker| action_tracker.action == action)
+        {
             self.next_actions.push(ActionTracker {
                 action,
                 duration: duration_ms,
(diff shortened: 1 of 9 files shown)


native/game_backend/src/player.rs
```diff
@@ -93,22 +97,21 @@ impl Player {
...
+        if let Some(action_tracker) = self.next_actions.first_mut() {
+            action_tracker.duration = action_tracker.duration.saturating_sub(elapsed_time_ms);
 
-    fn is_executing_action(&mut self) -> bool {
-        self.action_duration_ms != 0
+            if action_tracker.duration == 0 {
+                self.next_actions.remove(0);
+            }
+        }
+        self.action = self.next_actions.clone();
```
manucamejo, line 114: Maybe having 2 vectors is unnecessary, because we have the actions updated all the time. But is not a blocker. We can fix it in another PR.

native/game_backend/src/player.rs
```diff
@@ -327,6 +330,14 @@ impl Player {
...
         self.inventory[inventory_at].take()
     }
+
+    pub fn can_move(&self) -> bool {
+        self.action.first().is_none()
+    }
+
+    pub fn can_activate(&self) -> bool {
+        self.action.first().is_none()
+    }
```
ncontinanza, line 340: These are just the same behaviour with different name, right?

ncontinanza (changes requested):

native/game_backend/src/player.rs
agustinesco: They're exactly the same yeah, we should delete one of them

manucamejo (approved):

ncontinanza (approved):

PR #114, opened 2023-12-19
Modify position representations from integers to floats

AminArria:
This is a reboot of #60 because the merge conflicts got too complicated to fix and to reduce the scope a bit

This PR modify the different integer types use to represent position related things to a float (`f32` in Rust)

The non-exhaustive list of modified thing:
- size
- speed
- position X and Y coordinates
- radius
- max_distance
- minimum and maximum radius of map

You can test with https://github.com/lambdaclass/curse_of_mirra/pull/1292

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,3 +3,12 @@
 members = [
     "native/game_backend",
 ]
+
+## to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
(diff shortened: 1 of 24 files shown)


ncontinanza (dismissed):
LGTM! I tested it together with the frontend and everything worked fine.

ncontinanza (approved):

manucamejo (changes requested):
I'm getting an error when the game starts: `Could not decode field :game on %{}`. I can't test it :(

PR #116, opened 2023-12-19
check blocker actions in next_action list instead of current actions

manucamejo:
This pr aims to fix an error in the game where for example:
- You are walking left and try to hit to the right muflus ends up throwing a hit to the left


agustinesco (approved):
LGTM! i tried to replicate the error of the muflus direction while hitting and with this pr i wasn't able to replicate that error in a couple of matches, nice!

PR #118, opened 2023-12-20
Add Units

lotuuu:
This PR branches from gh-1000-select-character-functionality because it also affects character selection.

The goal of this PR is to add Units, which are instances of characters tied to a user, stored on DB and crucial to the autobattler. A user may have many units of one or more characters (for Curse of Mirra we will have at max 1, for Champions of Mirra we'll have many).

This PR also integrates Mirra to the Units system by assigning units to a user when they select a character (or, if they had selected it previously, just swapping the selected statuses of the unit he had before with the selected one). 

No endpoint for unit creation has been added since it is not necessary for Curse and we don't really know how we will want to implement them for Champions just yet.

Sample endpoint calls:
`curl -X PUT localhost:4000/users-characters/{self_client_id}/select_unit/{unit_id} -d slot=2`

`curl -X PUT localhost:4000/users-characters/{self_client_id}/unselect_unit/{unit_id} -d slot=`

A little hack has been implemented so that units can be preserved after reimporting the config (when character ids change).


manucamejo (approved):

agustinesco (approved):
nice

PR #120, opened 2023-12-20
Run config db setup on start & update readme

lotuuu:
After #104 you will get errors if you try to run a game without setting up the DB beforehand. This PR creates a .iex.exs file that gets triggered on iex start, which executes the config import. This will take care of config issues for devs.

Note that this does not automate changes to the config though, you will need to restart the app for the command to execute or run it manually.


agustinesco (approved):
LGTM! great adittion

PR #121, opened 2023-12-20
Fix cone angle calculation

agustinesco:
Nowdays we have an incomplete angle calculation when we check if a hit actually hitted an enemy, this is because we're only testing if the up half of the angle hitted and we should check the other side. I'll explain myself in the following diagrams

### disclaimer: the angles represented in these diagram is relative, it'll be always the same since it's the angle from the player to the target, so for example is the target is on top of the player the "270" would be a 0 instead

Before this pr the only side considered as hit were the players inside of the "up hit half" and after this pr we will consider both sides hit
player hit range

The error that this calculation caused was something like this:

no hit case

The target is clearly inside of the cone but since we're not checking the angles from the down half we are missing the hit

This pr will also reduce the muflus's hit angle a little bit so it feels more realistic

### DIFF
--- a/native/game_backend/src/map.rs
+++ b/native/game_backend/src/map.rs
@@ -43,7 +43,7 @@ pub fn in_cone_angle_range(
     let relative_angle = angle - center_player.direction;
     let normalized_angle = (relative_angle + 360.0) % 360.0;
 
-    normalized_angle < (cone_angle / 2.0)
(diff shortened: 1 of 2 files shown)


ncontinanza (approved):
LGTM! Tested it on the game and it's working great

manucamejo (approved):

PR #131, opened 2023-12-24
Fix user_response when there is no selected character

lotuuu:
Was breaking for Champions of Mirra because we didn't handle the case of the user having no selected units

### DIFF
--- a/lib/dark_worlds_server_web/controllers/user_controller.ex
+++ b/lib/dark_worlds_server_web/controllers/user_controller.ex
@@ -144,10 +144,12 @@ defmodule DarkWorldsServerWeb.UserController do
   defp user_response(%User{device_client_id: device_client_id} = user) do
     selected_unit = Units.get_selected_unit(user.id)
 
-    %{
-      device_client_id: device_client_id,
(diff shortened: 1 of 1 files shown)


martinacantaro:
Is this needed for https://github.com/lambdaclass/game_backend/pull/125 to work correctly?

lib/dark_worlds_server_web/controllers/user_controller.ex
```diff
@@ -144,10 +144,12 @@ defmodule DarkWorldsServerWeb.UserController do
   defp user_response(%User{device_client_id: device_client_id} = user) do
     selected_unit = Units.get_selected_unit(user.id)
 
-    %{
-      device_client_id: device_client_id,
-      selected_character: Utils.Characters.character_name_to_game_character_name(selected_unit.character.name)
-    }
+    if is_nil(selected_unit),
```
AminArria, line 147: nit: If you are going to use an `if` like this then you can use a `case` instead and would look/feel the same

AminArria (approved):
LGTM, just the nit, but no need to address it if no other changes are required

PR #141, opened 2024-01-05
Collision detection

manucamejo:
Added collision detection between entities.
Also grouped different objects into one called Entity.


.gitignore
```diff
@@ -28,4 +28,6 @@ erl_crash.dump
 # The directory NPM downloads your dependencies sources to.
 /assets/node_modules/
 
+*.so
```
pefontana, line 31: I think we can restore this file

pefontana (approved):

PR #150, opened 2024-01-09
Add projectiles entities

manucamejo:


lib/game_backend/game_updater.ex
```diff
@@ -10,14 +10,18 @@ defmodule GameBackend.GameUpdater do
...
   def join(game_pid, player_id) do
     GenServer.call(game_pid, {:join, player_id})
   end
 
+  def move(game_pid, player_id, direction) do
+    GenServer.call(game_pid, {:move, player_id, direction})
+  end
+
+  def attack(game_pid, player_id, skill) do
+    GenServer.call(game_pid, {:move, player_id, skill})
```
pefontana, line 22: Why dont we use :attack here?

messages.proto
```diff
@@ -27,3 +53,21 @@ message Entity {
   float speed = 9;
   Direction direction = 10;
 }
+
+/*
+ * Player Actions
+ */
+message Move {
```
pefontana, line 60: Why are we adding this structs?

lib/game_backend/game_updater.ex
manucamejo: You're right! My bad

pefontana (approved):

PR #151, opened 2024-01-09
Collision detection documentation

manucamejo:
- [x] Add documentation
- [ ] Add images of each case

### DIFF
--- a/docs/collision-detection.md
+++ b/docs/collision-detection.md
@@ -1,0 +1,37 @@
+# Collision detection
+
+We are working with entities.
+
+These entities can be of any of these types: Point, Line, Circle, Polygon.
+
+Each entity has attributes, which can be mandatory or not:
+- id: unique identifier
+- shape: Point, Line, Circle, Polygon. To determine which collision algorithm to use
+- radius: in the case of a circle
+- speed: in case of movement
+- direction: in case of movement
+- name: unique name of the entity
+- category: Player, Projectile, Obstacle. We use it to know when to check for collisions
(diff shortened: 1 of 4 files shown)


docs/collision-detection.md
```diff
@@ -0,0 +1,52 @@
...
+Each entity has attributes, which can be mandatory or not:
+- id: unique identifier
+- shape: Point, Line, Circle, Polygon. To determine which collision algorithm to use
+- radius: in the case of a circle
+- speed: in case of movement
+- direction: in case of movement
+- name: unique name of the entity
+- category: Player, Projectile, Obstacle. We use it to know when to check for collisions
+- vertices: in case the entity is defined by points
+- position: position of the entity on the map
```
mgrunwaldt, line 16: I would include the expected / accepted type for each attribute

docs/collision-detection.md
```diff
@@ -0,0 +1,52 @@
...
+
+For collision checking, we iterate through all entities searching for those that collide with the current one and return a list of all collisions.
+
+The possible collision detection algorithms we handle are:
```
mgrunwaldt, line 20: We don't have algorithms for all the possible combinations (polygon polygon for example). It could be good to state why we don't have them, it's not clear if it's because they are not necessary, if they are planned for the future, if by having these algorithms you can represent all the cases, etc.

mgrunwaldt (commented):
Only reviewed the docs

mgrunwaldt (approved):

mgrunwaldt (approved):

PR #152, opened 2024-01-09
Restrict players move when colliding with obstacles

pefontana:
# Restrict players move when colliding with obstacles

* Add walls to the map
* Change Players initial position
* Impl Debug trait to Rust structs
* Add entity.revert_move_entity() method
* Refactor fn collides_with to recibe a reference: entities: &Vec<Entity>
* Revert player move when colliding with an obstacle


priv/static/assets/app.css
manucamejo, line 1: Is this change necessary?

native/physics/src/lib.rs
```diff
@@ -67,11 +69,19 @@ fn move_player(game_state: GameState, player_id: u64, x: f64, y: f64) -> GameSta
 }
 
 #[rustler::nif()]
-fn move_entities(game_state: GameState) -> GameState {
+fn move_entities(game_state: GameState, obstacles: Vec<Entity>) -> GameState {
     let mut game_state: GameState = game_state;
-
     for entity in game_state.entities.values_mut() {
+        // Avoid calculation for entities with no speed
+        if entity.speed == 0.0 {
```
manucamejo, line 76: Can you add a direction condition? direction.x or direction.y should be different to 0.

lib/game_backend/game_updater.ex
```diff
@@ -26,11 +26,106 @@ defmodule GameBackend.GameUpdater do
   def init(%{players: players}) do
     game_id = self() |> :erlang.term_to_binary() |> Base58.encode()
 
+    left_wall_vertices = [
```
manucamejo, line 29: In the future, this should be a config

manucamejo (changes requested):

priv/static/assets/app.css
pefontana: When I compile the project that dependency made an upgrade
Should I restore it?

native/physics/src/lib.rs
pefontana: Sure I can add the check
But is not redundant? If the speed is 0, the player won't move, whether it has direction or no

priv/static/assets/app.css
manucamejo: I think you are doing a downgrade v3.3.2 -> v3.2.7

native/physics/src/lib.rs
manucamejo: Yes, but the speed is something inherent to the entity, it is the speed that it is going to move if it is moving. Right now it is the direction that determines whether it moves or not.

PR #154, opened 2024-01-10
Move phoenix code outside lib

Nico-Sanchez:


Nico-Sanchez:
Will re-do this using umbrella.

PR #156, opened 2024-01-10
Put backend inside umbrella project

Nico-Sanchez:
This PR adds 2 break changes:
- Changed project architecture to an umbrella representation (several applications under apps/ folder).
- Split the backend and client code (move Phoenix + JS code out).

This will benefit us in many ways:
- It will be easier to add more applications.
- The backend application (Arena) will be more simply with all the assets and client dependencies out of it.
- We keep the client (browser game viewer) application to test new features, not losing our current work.


config/config.exs
```diff
@@ -7,30 +16,20 @@
...
-  _: [
-    {"/play/:game_id/:player_id", GameBackend.GameSocketHandler, []},
-    {"/play/:player_id", GameBackend.SocketHandler, []},
-    {:_, Plug.Cowboy.Handler, {GameBackendWeb.Endpoint, []}}
-  ]
-]
-
-config :game_backend,
-  ecto_repos: [GameBackend.Repo],
+config :mirra_backend,
```
lotuuu, line 19: These configurations are obsolote

apps/arena/mix.exs
```diff
@@ -0,0 +1,69 @@
...
+  # Specifies which paths to compile per environment.
+  defp elixirc_paths(:test), do: ["lib", "test/support"]
+  defp elixirc_paths(_), do: ["lib"]
+
+  # Specifies your project dependencies.
+  #
+  # Type `mix help deps` for examples and options.
+  defp deps do
+    [
+      {:phoenix, "~> 1.7.10"},
```
lotuuu, line 35: We should create a helper module that stores our common dependencies among projects

lotuuu (approved):
There are some details about our umbrella implementation that could use some polishing, but overall its working just fine and given that this PR is blocking progress I think it's fine to merge as it is and tackle them in another PR

apps/game_client/Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: deps run start stop generate-ex-protos generate-js-protos generate-protos format lints check
+.PHONY: run deps start stop generate-ex-protos generate-js-protos generate-protos format lints check
```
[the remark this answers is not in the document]
lotuuu: @tvillegas98 and I can give some insight on how to do this.

PR #158, opened 2024-01-10
Change entity creation

manucamejo:
Entity creation refactor.
Now we create entities in elixir instead in rust.

### DIFF
--- a/lib/physics.ex
+++ b/lib/physics.ex
@@ -10,10 +10,7 @@ defmodule Physics do
   # NIF dummies usually just error out when called when the NIF is not loaded, as that should never normally happen.
   def new_game(_game_id), do: :erlang.nif_error(:nif_not_loaded)
   def add(_arg1, _arg2), do: :erlang.nif_error(:nif_not_loaded)
-  def add_player(_game_state, _player_id), do: :erlang.nif_error(:nif_not_loaded)
   def move_player(_game_state, _player_id, _x, _y), do: :erlang.nif_error(:nif_not_loaded)
   def check_collisions(_entity, _entities), do: :erlang.nif_error(:nif_not_loaded)
-  def add_polygon(_game_state), do: :erlang.nif_error(:nif_not_loaded)
   def move_entities(_game_state), do: :erlang.nif_error(:nif_not_loaded)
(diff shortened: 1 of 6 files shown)


lib/physics.ex
```diff
@@ -10,10 +10,7 @@ defmodule Physics do
   # NIF dummies usually just error out when called when the NIF is not loaded, as that should never normally happen.
   def new_game(_game_id), do: :erlang.nif_error(:nif_not_loaded)
   def add(_arg1, _arg2), do: :erlang.nif_error(:nif_not_loaded)
-  def add_player(_game_state, _player_id), do: :erlang.nif_error(:nif_not_loaded)
   def move_player(_game_state, _player_id, _x, _y), do: :erlang.nif_error(:nif_not_loaded)
   def check_collisions(_entity, _entities), do: :erlang.nif_error(:nif_not_loaded)
-  def add_polygon(_game_state), do: :erlang.nif_error(:nif_not_loaded)
   def move_entities(_game_state), do: :erlang.nif_error(:nif_not_loaded)
```
Nico-Sanchez, line 15: If everything is an Entity. Why do we need a `move_player/4` and `move_entities/1` function?
If we need both, I think we should rename `move_player/4` to `move_entity/4`. I think we may want to use it with any entity, not just players.

Nico-Sanchez (commented):
Works like a charm 🥇 

I have a few questions to understand why are we doing this and to give it a second thought:
Why is everything an entity now? Don't we need to know if it's a player or something else? We shouldn't care about that in Rust?

lib/physics.ex
manucamejo: You are right, I miss some changes:
Instead of `move_player` we could call it `set_entity_direction` because that it what it does.
We need the `move_entities` function because is the responsible of moving all entities and check if someone collides with a wall or an obstacle.

manucamejo: The idea is that Rust doesn't care about anything related to the players or the projectiles. This includes the player's health or the projectile's damage Rust simply doesn't care.
Perhaps it wants to know if an entity is "alive" or "active" to determine whether to move it or consider it for collisions. But no additional information.

Nico-Sanchez (approved):

PR #159, opened 2024-01-11
Custom entity info

manucamejo:
Add custom entity info.
Allow us to have information about a player or a projectile depending on they category.


apps/arena/lib/arena/game_updater.ex
```diff
@@ -57,6 +68,7 @@ defmodule Arena.GameUpdater do
           |> Map.put(:category, to_string(entity.category))
           |> Map.put(:shape, to_string(entity.shape))
           |> Map.put(:name, "Entity" <> Integer.to_string(entity_id))
+          |> Map.put(:aditional_info, entity |> Entities.maybe_add_custom_info())
```
AminArria, line 71: I get what this does, but I don't get why we need it... Why can't we have it like?

```elixir
|> Map.put(:aditional_info, entity.aditional_info)
```

pefontana (approved):

apps/arena/lib/arena/game_updater.ex
manucamejo: I tried, but I couldn't get it to work. Maybe it's because of how the struct is defined. I can try again.

AminArria (approved):

PR #160, opened 2024-01-11
separate projectiles and players

manucamejo:
- Separate projectiles and players
- Clean up GameUpdater module


apps/arena/lib/arena/game_updater.ex
```diff
@@ -122,8 +89,69 @@ defmodule Arena.GameUpdater do
...
+
+  # Move entities and add game fields
+  defp update_entities(entities) do
+    new_state = Physics.move_entities(entities)
+
+    Enum.reduce(new_state, %{}, fn {key, value}, acc ->
+      entity =
+        Map.get(entities, key)
+        |> Map.merge(value)
+        |> Map.put(:is_colliding, false)
```
Nico-Sanchez, line 127: This is being set as false and saved in players' state, but when broadcasting to the client we're doing the collision calculation.
We should do the collision check before and saving the same value in both places (state and broadcast message).

Nico-Sanchez (changes requested):
We lost our collision feature in the front-end. Let's fix it.

apps/arena/lib/arena/game_updater.ex
manucamejo: Fixed

Nico-Sanchez (approved):

PR #167, opened 2024-01-15
Distinction between player_id and client_id

AminArria:
- Rename variables to properly reflect if it is `client_id` or `player_id`
- Websockets recieve client_id
- `game_socket_handler` tracks the player_id when it join the games and uses that to communicate to the game 

To test this you can try going to http://localhost:3000/board/notanid, everything should keep working as before

### DIFF
--- a/apps/arena/config/config.exs
+++ b/apps/arena/config/config.exs
@@ -10,8 +10,8 @@ import Config
 # Configures the endpoint
 dispatch = [
   _: [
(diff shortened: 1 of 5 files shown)


Nico-Sanchez (changes requested):
Works as expected 🏅 

Let's rename all wrongly named `players` occurrences before merging it.

Nico-Sanchez (approved):

PR #172, opened 2024-01-16
Normalize movement direction

manucamejo:
There was a bug when using the joystick input.
We were using the received input so if you use the joystick in an intermediate position your player has a lower speed:
image
In order to fix this bug we normalize the joystick input, so the speed doesn't depends on the magnitude of that value.


apps/arena/lib/arena/game_socket_handler.ex
```diff
@@ -34,10 +34,15 @@ defmodule Arena.GameSocketHandler do
...
+        GameUpdater.attack(state.game_pid, String.to_integer(state.client_id), skill)
 
       %{action_type: {:move, %{direction: direction}}, timestamp: timestamp} ->
-        GameUpdater.move(state.game_pid, state.player_id, {direction.x, direction.y}, timestamp)
+        GameUpdater.move(
+          state.game_pid,
+          String.to_integer(state.client_id),
+          {direction.x, direction.y},
+          timestamp
+        )
```
Nico-Sanchez, line 45: Why this change? The players perform the actions not the clients.

apps/game_client/lib/game_client/client_socket_handler.ex
```diff
@@ -48,7 +53,6 @@ defmodule GameClient.ClientSocketHandler do
   end
 
   defp ws_url(player_id, game_id) do
-    # TODO Remove hardcoded host
```
Nico-Sanchez, line 51: Why removing this TODO tag? 
If we want to remove it we should remove it in `apps/game_client/assets/js/hooks/game_queue.js` as well

Nico-Sanchez (changes requested):

Nico-Sanchez:
Also, please add a description of what the PR does.

apps/game_client/lib/game_client/client_socket_handler.ex
manucamejo: Fixed

apps/arena/lib/arena/game_socket_handler.ex
manucamejo: You're right, there was an error in Unity, fixed

apps/arena/lib/arena/game_socket_handler.ex
```diff
@@ -37,7 +37,12 @@ defmodule Arena.GameSocketHandler do
         GameUpdater.attack(state.game_pid, state.player_id, skill)
 
       %{action_type: {:move, %{direction: direction}}, timestamp: timestamp} ->
-        GameUpdater.move(state.game_pid, state.player_id, {direction.x, direction.y}, timestamp)
+        GameUpdater.move(
```
AminArria, line 40: Beware with your formatting, you didn't make any change here to actually make a formatting. Also, the formatting might not be using the same rules

manucamejo: There was a change but I reverted it. However the command `mix format` didn't change it again.

Nico-Sanchez (approved):

PR #180, opened 2024-01-18
Add shell.nix for mirra_backend

klaus993:
With this, every dependency (except Docker) is installed and ready for mirra backend.

Docker is missing because it's not easy to install Docker for Mac programatically, and the docker nix package doesn't work for macOS. To work around this, we need to execute `nix-shell` without the `--pure` flag so Docker is taken from the host, so it needs to be pre-installed. Besides from that, the lack of the `--pure` flags doesn't conflict in any way because elixir and erlang are taken from the nix store and not from the OS.


samoht9277:
Looking good. The idea would be to fully replace docker with nix. We should see what the docker does and replace it with nix entirely.

klaus993:
rust is missing

samoht9277:
Closing this in order to continue the deployment with Nix via devenv.

PR #185, opened 2024-01-23
Add new apps to umbrella

lotuuu:
https://www.notion.so/lambdaclass/Backend-Architecture-df05ae101a0c460cbdb975b21be1634a

Disclaimer: the main objective of this PR is to present and get feedback on the design and architecture of the new backend, implemented as an umbrella project. There are some functionalities missing, such as the unit summon and fusion mechanics, and some implementations that can still improve. We'll take care of those in future PRs.

Adds apps to the umbrella app:
- Gateway
  - Receives messages via websocket and routes them to the corresponding game's application.
- ChampionsOfMirra
  - Application for Champions Of Mirra game. Has modules for:
     - Battle - Simulates battles. For now, only allows fights between a user and a level from the campaign.
    - Campaigns - Creates a campaign with customizable difficulty, in the shape of 5 units per level.
    - Items - Handles logic for Items. Consumes from the Items app for general items logic.
    - Units - Handles logic for Units. Consumes from the Units app for general items logic.
    - Users - Handles logic for Users. Consumes from the Users app for general items logic.
- GameBackend
  - Persistance layer and shared logic. 
  - Users logic that's general among all games. This module is quite incomplete for now, since users are only made of a unique username.
  - Units logic that's general among all games. Defines the schemas for the characters of every game. These act like templates for Units, which are instances of them tied to a user or a campaign level.
  - Items logic that's general among all games. Defines the schemas for the item templates of every game. These act like templates for Items, which are instances of them that belong to a user and can be equipped to a unit. 

What's important to note is that each game's application decides how to use the functionalities these applications have. For example, take a look at how our game implements `Champions.Units.select_unit/3` and `unselect_unit/2`. For the first one, we have some rules on how and when a unit can be selected, so we check they are met before calling the GameBackend app. For the second one, we don't care for the context it is called in, so we just call `GameBackend.Units.unselect_unit/2` instantly. Another game might have different requirements for unit selection/unselection, and it would be handled in its own "NewGame.Units" module. 

All data is stored in the same DB. A sort of multitenancy is implemented by adding a `game_id` field to the tables that need it (Users, Currencies, Characters, ItemTemplates, Levels). Adding it to "template" schemas like Character and ItemTemplate instead of every Unit and Item saves us storage.

A small guide to test the functionalities of the app can be found in the [Gateway app's README](https://github.com/lambdaclass/mirra_backend/blob/823244221a45f36a4a2712b4672df6294480713f/apps/gateway/README.md)


Makefile
```diff
@@ -1,9 +1,7 @@
 .PHONY: start format credo check generate-protos generate-arena-protos generate-game-client-protos
 
 start:
-	cd apps/arena && docker-compose up -d
-	cd apps/game_client && docker-compose up -d
-	cd apps/game_client/assets && npm install
+	docker-compose up -d
```
manucamejo, line 4: Please, use `docker compose` instead of `docker-compose`

manucamejo (changes requested):

Makefile
```diff
@@ -1,7 +1,6 @@
 .PHONY: start format credo check generate-protos generate-arena-protos generate-game-client-protos
 
 start:
-	cd apps/arena && docker compose up -d
 	cd apps/game_client && docker compose up -d
```
manucamejo, line 4: We should delete this line as well
And also the `game_client` ecto config

.tool-versions
```diff
@@ -1,3 +1,3 @@
 erlang 26.0.2
-elixir 1.15.4-otp-26
+elixir 1.16.0-otp-26
```
AminArria, line 2: If you are changing versions here we should also change the CI at least

jrchatruc:
Overall looks good to me, though when running `make start` I get the following error
```
You have configured application :arena in your configuration file,
but the application is not available.

This usually means one of:

  1. You have not added the application as a dependency in a mix.exs file.

  2. You are configuring an application that does not really exist.

Please ensure :arena exists or remove the configuration.
```

and
 
```
18:25:58.990 [error] Postgrex.Protocol (#PID<0.458.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (localhost:5433): connection refused - :econnrefused
```

even though I have setup the db with `docker compose up`.

I think the `README` needs to be updated a bit to explain how to setup everything and how to use it/run a simple example.

igaray (changes requested):

apps/champions_of_mirra/README.md
```diff
@@ -0,0 +1,21 @@
+# Autobattler
+
+**TODO: Add description**
```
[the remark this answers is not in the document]
lotuuu: READMEs for applications have been added! Let me know what you think

apps/champions_of_mirra/lib/champions_of_mirra.ex
```diff
@@ -0,0 +1,71 @@
+defmodule ChampionsOfMirra do
+  @moduledoc """
+  Champions of Mirra request handler. Receives messages from the Gateway app and
```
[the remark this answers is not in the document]
lotuuu: I see the terms used were not the best. I've updated the moduledoc now, let me know if you find it clearer

jrchatruc (approved):

juanbono (dismissed):

manucamejo (approved):

juanbono (approved):

PR #191, opened 2024-01-24
add ping/pong messages

manucamejo:


AminArria (approved):
Pushed minor fixes, but LGTM

PR #203, opened 2024-01-29
stamina first iteration

### LINKED ISSUE #219: Stamina

manucamejo:
Closes #219
Also https://github.com/lambdaclass/curse_of_mirra/issues/985

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -148,6 +148,32 @@ defmodule Arena.GameUpdater do
     {:noreply, state}
   end
 
+  def handle_info({:recharge_stamina, player_id}, state) do
+    player = Map.get(state.game_state.players, player_id)
+
+    player =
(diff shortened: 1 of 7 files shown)


apps/arena/lib/arena/game_updater.ex
```diff
@@ -148,6 +148,32 @@ defmodule Arena.GameUpdater do
     {:noreply, state}
   end
 
+  def handle_info({:recharge_stamina, player_id}, state) do
+    player = Map.get(state.game_state.players, player_id)
+
+    player =
+      case player.aditional_info.available_stamina do
+        3 ->
```
AminArria, line 156: I guess ideally this should be the `base_stamina` from the character

AminArria (approved):
LGTM, one small comment, but no need to fix it now

apps/arena/lib/arena/game_updater.ex
manucamejo: Yes!

PR #215, opened 2024-01-30
add natural healing

manucamejo:


apps/arena/lib/arena/game_updater.ex
```diff
@@ -53,6 +56,8 @@ defmodule Arena.GameUpdater do
   def handle_info(:update_game, %{game_state: game_state} = state) do
     Process.send_after(self(), :update_game, state.game_config.game.tick_rate_ms)
 
+    now = DateTime.utc_now() |> DateTime.to_unix(:millisecond)
```
AminArria, line 59: What do you think of using `System.monotonic_time(:millisecond)` instead? 

Does the client need this timestamp?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -79,7 +84,11 @@ defmodule Arena.GameUpdater do
 
           player ->
             health = max(player.aditional_info.health - projectile.aditional_info.damage, 0)
-            player = put_in(player, [:aditional_info, :health], health)
+
+            player =
+              put_in(player, [:aditional_info, :health], health)
+              |> put_in([:aditional_info, :last_natural_healing_update], now)
```
AminArria, line 90: Don't follow this, why do we need it?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -175,6 +184,34 @@ defmodule Arena.GameUpdater do
     {:noreply, state}
   end
 
+  # Natural healing
+  def handle_info(:natural_healing, state) do
+    Process.send_after(self(), :natural_healing, @natural_healing_interval_ms)
+
+    players =
+      Enum.reduce(state.game_state.players, %{}, fn {player_id, player}, players_acc ->
```
AminArria, line 192: I think this is a good candidate to do `update_in/3` instead of `Enum.reduce/3`

apps/arena/lib/arena/game_updater.ex
manucamejo: Because we will heal players only after some interval, so we need to know when was the last time we healed it.
Do you have another way to do it?

AminArria: Didn't understand why we needed because of the naming jajaja got it now, all good

AminArria (approved):

PR #231, opened 2024-01-31
Implementations for circle and cone hit mechanics

AminArria:


apps/arena/native/physics/src/map.rs
```diff
@@ -89,27 +89,36 @@ impl Entity {
                 continue;
             }
 
-            match entity.shape {
-                Shape::Circle => {
+            let self_shape = self.shape.clone();
+            let entity_shape = entity.shape.clone();
```
AminArria, line 93: Would be good to get a second opinion about doing this

manucamejo (approved):

PR #237, opened 2024-02-02
Increase muflus hitbox and zone start and shrink intervals

BertovDev:


apps/arena/priv/config.json
```diff
@@ -98,7 +98,7 @@
       "name": "muflus",
       "active": true,
       "base_speed": 25.0,
-      "base_size": 50.0,
+      "base_size": 60.0,
```
manucamejo, line 101: Muflus shouldn't have less size than h4ck

manucamejo (changes requested):

manucamejo (approved):

manucamejo (approved):

PR #240, opened 2024-02-06
Add Units functionalities

lotuuu:
Adds many functionalities related to units. Implements them for the Champions app.

Functionalities added:
- Gacha
  - Allows for Gacha mechanics, which involve opening boxes to receive prizes. What we needed in CHoM was to have these prizes be just units, so that's what we implemented. Drops probabilities are decided per-rank, and that is represented in our model through the characters' `ranks_dropped_in` field. It's a list because we want some champions to be able to drop in many ranks. See: [Summon](https://www.notion.so/lambdaclass/Rarities-Summoning-9465b151e8144d30a92c795ce6d3fab4?pvs=4).
- Progression
  - Allows unit progression in the shape of level ups, tier ups and rank ups. These mechanics are a little bit tricky if you're not familiar with autobattlers. The gist of it is that you can level up your unit as much as your tier allows you, and you can tier up your unit as much as your rank allows you. Level and Tier ups share the way in which they are used: they just cost a bunch of currencies from the user. Ranks ups, on the other hand, are only accesible through Fusion. See: [Evolution](https://www.notion.so/lambdaclass/Tiers-Evolution-3b7d9740ad384a5cbf7eaeac377083ee?pvs=4)
- Fusion
  - Allows "combining" many units into one to increase its rank. Effectively it just deletes the other ones. There are pre-established rules for which units are required based on the target unit's rank. Also there are requirements based on character rank (`common` units cannot be fused, for example) See: [Ranks](https://www.notion.so/lambdaclass/Ranks-Fusion-1d1df44f076c4e2d81bf296ee05c4921?pvs=4) 

Other General Changes:
- Implements a CurrencyCost struct and integrates it where we previously used tuples.
- Defines get_rank and get_rarity functions in Units to be able to have a single source for these values. Of course, these are in the Champions app since they are specific to that game.
- Same was done in Champions.Utils for the game_id
- IMPORTANT ONE!  Adds a GameBackend.Transaction module. The idea behind it is that we can keep our game apps repo-agnostic, while still being able to encapsulate game-specific behavior in transactions. This is important for operations that consume currencies and for fusion, for example.
- Added tests using SocketTester. These tests are just some general integration tests to help make development faster.

When testing, remember to run the seeds again as fusion might not work otherwise.


lotuuu:
Commit history got messed up for some reason. Closing this PR and starting a new one from the same `chom-units` branch.

lotuuu:
New PR was showing the same diff, seems like something that can't be solved trivially. We'll just have to review it using only the files changed. For reference, [this](https://github.com/lambdaclass/mirra_backend/pull/240/commits/8e013ee7ad61385dc0f69f5b65bf5ef01aa48a04) was about the point in which the changes specific to this PR started.

PR #242, opened 2024-02-06
[GH-277] Muflus leap

AminArria:
- [x] Movement implementation
- [x] Damage on arrival
- [ ] Slow on arrival (optional for this PR)
- [x] Limit leap to config range
- [x] Skill protobuf messages with parameters 

Test with https://github.com/lambdaclass/curse_of_mirra/pull/1415


apps/arena/lib/arena/game/player.ex
```diff
@@ -167,4 +200,13 @@ defmodule Arena.Game.Player do
...
   defp maybe_trigger_natural_heal(player, _), do: player
+
+  defp add_or_remove_moving_action(current_actions, direction) do
+    if direction == {0.0, 0.0} do
+      current_actions -- [%{action: :MOVING, duration: 0}]
+    else
+      current_actions ++ [%{action: :MOVING, duration: 0}]
+    end
+    |> Enum.uniq()
+  end
```
agustinesco, line 211: just a comment, you could match the direction directly on the declaration:
```
defp add_or_remove_moving_action(current_actions, {0.0, 0.0}), do: REMOVE
defp add_or_remove_moving_action(current_actions, _direction), do: APPEND
```

apps/arena/lib/arena/game_socket_handler.ex
```diff
@@ -81,6 +81,9 @@ defmodule Arena.GameSocketHandler do
           timestamp
         )
 
+      # %{action_type: {:attack, %{skill: skill}}} ->
+      #   GameUpdater.attack(state.game_pid, state.player_id, skill, %{target: %{x: 0.0, y: 0.0}})
+
```
agustinesco, line 86: should we remove this?

apps/arena/lib/arena/game/player.ex
AminArria: If not a problem I'll address this in a follow up because this was copy&paste from `game_updater.ex` and doesn't really have anything to do with this PR

About the uniqueness, sounds reasonable. What we where trying to solve here was that without it the actions grow unbounded because they grow with each move action

agustinesco (changes requested):
The destination of the leap is not working as expected, for example on this video i'm trying to jump to the center of the circle but i end up in an entire different position

Uploading Screen Recording 2024-02-14 at 15.21.17.mov…

apps/arena/lib/arena/game/player.ex
agustinesco: yes! lets address this later, it was just a comment about a possible code change

apps/arena/lib/arena/game_socket_handler.ex
AminArria: removed

apps/arena/lib/arena/game/skill.ex
```diff
@@ -169,6 +170,28 @@ defmodule Arena.Game.Skill do
     |> Map.put(:projectiles, projectiles)
   end
 
+  def do_mechanic(game_state, player, {:leap, leap}, %{target: target}) do
+    Process.send_after(self(), {:stop_leap, player.id, player.speed}, leap.duration_ms)
```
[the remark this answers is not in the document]
AminArria: I've asked in slack about it, this will probably have some debate cause we have to either have a fixed duration or a fixed speed, but not both

apps/arena/lib/arena/game_updater.ex
```diff
@@ -67,6 +66,18 @@ defmodule Arena.GameUpdater do
...
+    #     if Player.forced_moving?(player) do
+    #       moved = Physics.move_entity(ticks_to_move, player, game_state.external_wall)
+    #       Map.put(players_acc, player_id, Map.merge(player, moved))
+    #     else
+    #       Map.put(players_acc, player_id, player)
+    #     end
+    #   end)
+
+    # game_state = Map.put(game_state, :players, players_moved)
+
```
agustinesco, line 80: If we're not using this let's delete it

apps/arena/priv/config.json
```diff
@@ -47,7 +47,7 @@
   },
   "game": {
     "tick_rate_ms": 30,
-    "zone_shrink_start_ms": 20000,
+    "zone_shrink_start_ms": 999999,
```
agustinesco, line 50: Let's restore this

apps/arena/lib/arena/configuration.ex
```diff
@@ -65,4 +66,12 @@ defmodule Arena.Configuration do
       skill
     end
   end
+
+  defp parse_mechanic_fields({:leap, attrs}) do
+    {:leap, %{attrs | on_arrival_mechanic: parse_mechanic_config(attrs.on_arrival_mechanic)}}
```
agustinesco, line 71: Why are we parsing just the leap on arrival mechanics? maybe you can make this method more generic doing:
```
defp parse_mechanic_fields({skill, attrs}) do
{skill, %{attrs | on_arrival_mechanic: parse_mechanic_config(attrs.on_arrival_mechanic)}}
end
```

apps/arena/lib/arena/game/player.ex
```diff
@@ -90,21 +92,52 @@ defmodule Arena.Game.Player do
...
+  def move(%{aditional_info: %{forced_movement: true}} = player, _) do
+    player
+  end
+
+  def move(player, direction) do
+    current_actions =
+      add_or_remove_moving_action(player.aditional_info.current_actions, direction)
+
+    {x, y} = direction
+    is_moving = x != 0.0 || y != 0.0
```
agustinesco, line 104: Remember you can add a `?` sign to variables making booleans more clear

agustinesco (approved):
LGTM!

agustinesco (approved):
🪨

PR #243, opened 2024-02-07
remove projectiles upon collision with external wall

manucamejo:


apps/arena/lib/arena/game_updater.ex
```diff
@@ -58,45 +58,31 @@ defmodule Arena.GameUpdater do
   def handle_info(:update_game, %{game_state: game_state} = state) do
     Process.send_after(self(), :update_game, state.game_config.game.tick_rate_ms)
 
-    entities_to_collide = Map.merge(game_state.players, game_state.projectiles)
+    # only use alive players
```
lotuuu, line 61: We could turn this comment into an issue and reference it with `# #246` instead. This way we can keep the detailed context of this problem.

apps/arena/lib/arena/game_updater.ex
```diff
@@ -469,4 +481,73 @@ defmodule Arena.GameUpdater do
...
+      # check if the projectiles is inside the walls
+      collides_with =
+        case projectile.collides_with do
+          [] -> [external_wall_id]
+          entities -> List.delete(entities, external_wall_id)
+        end
+
+      collided_entity = decide_collided_entity(projectile, collides_with, external_wall_id)
+
+      case {
```
lotuuu, line 497: It would be nice to have this follow the same pattern as `decide_collided_entity` to make the code clearer

lotuuu (changes requested):

apps/arena/lib/arena/game_updater.ex
manucamejo: Thanks Lotu, I created an issue for that (#247) I'll do it in another PR.

lotuuu (approved):

PR #249, opened 2024-02-08
Gh246 exclude alive players

### LINKED ISSUE #246: Only use alive players
- [x] Projectiles should only collides with alive players
- [x] Dead players shouldn't be able to move
- [x] Dead players shouldn't be cured
- [x] Killfeed shouldn't be replicated when a player is dead

manucamejo:
Closes #246 
- [x] Projectiles should only collides with alive players
- [x] Dead players shouldn't be able to move
- [x] Dead players shouldn't be cured
- [x] Killfeed shouldn't be replicated when a player is dead
- [x] Block incomming messages if the player is dead

To test it, use 3 players, and try to execute actions in the browser

### DIFF
--- a/apps/arena/lib/arena/game/player.ex
+++ b/apps/arena/lib/arena/game/player.ex
@@ -34,7 +34,7 @@ defmodule Arena.Game.Player do
 
   def trigger_natural_healings(players) do
     Enum.reduce(players, %{}, fn {player_id, player}, players_acc ->
-      player = maybe_trigger_natural_heal(player)
+      player = maybe_trigger_natural_heal(player, alive?(player))
       Map.put(players_acc, player_id, player)
     end)
   end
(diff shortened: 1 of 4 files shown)


apps/arena/lib/arena/game/player.ex
```diff
@@ -34,7 +34,7 @@ defmodule Arena.Game.Player do
 
   def trigger_natural_healings(players) do
     Enum.reduce(players, %{}, fn {player_id, player}, players_acc ->
-      player = maybe_trigger_natural_heal(player)
+      player = maybe_trigger_natural_heal(player, alive?(player))
```
AminArria, line 37: The only reason we need `alive?(player)` is because we don't have a `player.status`

(Nothing to fix here, sort of me ranting about it)

AminArria (approved):

apps/arena/lib/arena/game/player.ex
manucamejo: Yeah, we have an issue created to fix that.

PR #266, opened 2024-02-15
[GH-247] Refactor projectile collision handler

### LINKED ISSUE #247: Refactor projectile collision resolution
We need to do 2 fixes on this issue
- Rework the function `resolve_collisions`  from the `Arena.GameUpdater` to upgrade his readability , to do so we can simply replace the case with a matching on a function parameters
- Write documentation on how the function decides when it should collide and what to do when that happen

agustinesco:
closes #247

I'll add a simple description about what the method `resolve_collisions` will do when the projectile collide with different elements in the map

Extracted case call to a private function that handles different collision cases
Since the wall collision do the same as the obstacle one we'll handle them in the same case

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -94,8 +94,7 @@ defmodule Arena.GameUpdater do
         game_state.external_wall.id
       )
 
-    players =
-      apply_zone_damage(players, game_state.zone, state.game_config.game)
+    players = apply_zone_damage(players, game_state.zone, state.game_config.game)
 
     game_state =
       game_state
@@ -516,9 +515,16 @@ defmodule Arena.GameUpdater do
   defp maybe_receive_zone_damage(player, _elaptime, _zone_damage_interval, _zone_damage),
     do: player
 
+  # This method will decide what to do when a projectile has collided with something in the map
+  # - If collided with something with the same owner skip that collision
+  # - If collided with external wall or obstacle explode projectile
+  # - If collided with another player do the projectile's damage
+  # - Do nothing on unexpected cases
+  defp resolve_collisions(projectiles, players, obstacles, external_wall_id)
+
   defp resolve_collisions(projectiles, players, obstacles, external_wall_id) do
-    Enum.reduce(projectiles, {projectiles, players}, fn {projectile_id, projectile},
-                                                        {projectiles_acc, players_acc} ->
+    Enum.reduce(projectiles, {projectiles, players}, fn {_projectile_id, projectile},
+                                                        {_projectiles_acc, players_acc} = accs ->
       # check if the projectiles is inside the walls
       collides_with =
         case projectile.collides_with do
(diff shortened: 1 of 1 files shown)


apps/arena/lib/arena/game_updater.ex
```diff
@@ -516,9 +515,16 @@ defmodule Arena.GameUpdater do
   defp maybe_receive_zone_damage(player, _elaptime, _zone_damage_interval, _zone_damage),
     do: player
 
+  # This method will decide what to do when a projectile has collided with something in the map
+  # - If collided with something with the same owner skip that collision
+  # - If collided with external wall or obstacle explode projectile
+  # - If collided with another player do the projectile's damage
+  # - Do nothing on unexpected cases
+  defp resolve_collisions(projectiles, players, obstacles, external_wall_id)
```
AminArria, line 523: Would be nice to rename this to `resolve_projectile_collisions` or something that explicitly says that its about projectiles

apps/arena/lib/arena/game_updater.ex
```diff
@@ -529,52 +535,55 @@ defmodule Arena.GameUpdater do
...
+
+    {
+      Map.put(projectiles_acc, projectile.id, projectile),
+      players_acc
+    }
+  end
+
+  # Projectile collided a player
+  defp apply_collision_updates(projectile, player, _, _, {projectiles_acc, players_acc})
+       when not is_nil(player) do
```
AminArria, line 569: Technically redundant this `when not is_nil(player)` due to the above clause

apps/arena/lib/arena/game_updater.ex
```diff
@@ -529,52 +535,55 @@ defmodule Arena.GameUpdater do
...
+    player = Player.change_health(player, projectile.aditional_info.damage)
+
+    projectile = put_in(projectile, [:aditional_info, :status], :EXPLODED)
+
+    if player.aditional_info.health == 0 do
```
AminArria, line 574: Change to `Player.is_alive?/1`

apps/arena/lib/arena/game_updater.ex
agustinesco: With this we're checking that the projectile actually collided with a player, if we remove it this would be the default clause and we don't want that, the nil match is prioritizing the player collisions with over the obstacles/wall ones

apps/arena/lib/arena/game_updater.ex
agustinesco: Sure thing! nice suggestion

apps/arena/lib/arena/game_updater.ex
agustinesco: Done!

apps/arena/lib/arena/game_updater.ex
AminArria: You are right

manucamejo (approved):

PR #268, opened 2024-02-15
[GH-267] Power up feature

### LINKED ISSUE #267: Drop power ups on death
Whenever a player dies it should drop a power up, later on if a player walks over them they should pick the power up

agustinesco:
Closes #267

Now whenever a player dies it'll drop a power up, the amount dropped will depend on how many power ups the player had at the moment hi dies

When a player walks over a power up it should pick him up

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -332,6 +332,7 @@ defmodule Arena.GameUpdater do
              game_id: state.game_id,
              players: complete_entities(state.players),
              projectiles: complete_entities(state.projectiles),
+             power_ups: complete_entities(state.power_ups),
              server_timestamp: state.server_timestamp,
              player_timestamps: state.player_timestamps,
              zone: state.zone,
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -692,8 +692,8 @@ defmodule Arena.GameUpdater do
   end
 
   defp get_amount_of_power_ups(%{aditional_info: %{power_ups: power_ups}}, power_ups_per_kill) do
-    Enum.sort_by(power_ups_per_kill, fn %{amount: minimun} -> minimun end, :desc)
-    |> Enum.find(fn %{amount: minimun} ->
+    Enum.sort_by(power_ups_per_kill, fn %{minimun_amount: minimun} -> minimun end, :desc)
+    |> Enum.find(fn %{minimun_amount: minimun} ->
       minimun <= power_ups
     end)
     |> case do
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -558,7 +558,12 @@ defmodule Arena.GameUpdater do
 
         # Projectile hit a player
         {_entity_id, player, nil} ->
-          player = Player.change_health(player, projectile.aditional_info.damage)
+          owner_player = Map.get(players, projectile.aditional_info.owner_id)
+
+          real_damage =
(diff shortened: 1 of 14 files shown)


apps/arena/lib/arena/game_updater.ex
```diff
@@ -598,4 +610,72 @@ defmodule Arena.GameUpdater do
...
+      )
+
+    put_in(state, [:game_state, :power_ups, last_id], power_up)
+    |> put_in([:game_state, :last_id], last_id)
+    |> spawn_power_ups(victim, amount - 1)
+  end
+
+  defp spawn_power_ups(state, _victim, _amount), do: state
+
+  defp get_of_power_ups(%{aditional_info: %{power_ups: power_ups}}, power_ups_per_kill) do
```
AminArria, line 642: Please rename this function its really hard to understand what it is supposed to do

apps/arena/lib/arena/game_updater.ex
```diff
@@ -598,4 +610,72 @@ defmodule Arena.GameUpdater do
...
+    end
+  end
+
+  defp handle_power_ups(players, power_ups) do
+    power_ups =
+      Map.reject(power_ups, fn {_power_up_id, power_up} ->
+        power_up.aditional_info.status == :TAKEN
+      end)
+
+    Enum.reduce(players, {players, power_ups}, fn {_player_id, player},
```
AminArria, line 659: If I followed this correctly seems like if multiple players are colliding with the same power up all of them will take

AminArria (changes requested):

apps/arena/lib/arena/game_updater.ex
```diff
@@ -598,4 +610,72 @@ defmodule Arena.GameUpdater do
...
+      Entities.new_power_up(
+        last_id,
+        random_position,
+        victim.direction,
+        victim.id
+      )
+
+    put_in(state, [:game_state, :power_ups, last_id], power_up)
+    |> put_in([:game_state, :last_id], last_id)
+    |> spawn_power_ups(victim, amount - 1)
```
AminArria, line 637: Why not `Enum.reduce`?

agustinesco: This is the best way i found to not spawn power ups when the calculation method returns 0, i though about using a reduce with a range 0..amount but you'll always have at least one cicle with 0. Do you know any way better than this?

apps/arena/lib/arena/game_updater.ex
agustinesco: You are totally right! if multiple player collide with a power up at the same time they'll both pick it up, i'll handle this adding a status check for the power up since this is being updated on each iteration

apps/arena/lib/arena/game_updater.ex
agustinesco: Sure thing! i'm not entirely sure what happened here 😅

apps/arena/priv/config.json
```diff
@@ -17,7 +17,7 @@
   },
   "game": {
     "tick_rate_ms": 30,
-    "zone_shrink_start_ms": 20000,
+    "zone_shrink_start_ms": 999999999,
```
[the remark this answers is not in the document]
agustinesco: Yup, this was something i changed to be able to test the power up mechanic. Restored!

apps/arena/lib/arena/game_updater.ex
agustinesco: That worked! i't more readable now thanks

AminArria (approved):
LGTM

manucamejo (approved):

PR #271, opened 2024-02-16
Add user experience

lotuuu:
Adds experience to users. To test you only really need to check that the Champions.Users.add_experience/2 works well and performs level-ups (and multiple level-ups) when appropiate.

### DIFF
--- a/apps/champions/lib/champions/users.ex
+++ b/apps/champions/lib/champions/users.ex
@@ -88,4 +88,34 @@ defmodule Champions.Users do
     Currencies.add_currency(user.id, Currencies.get_currency_by_name!("Gold").id, 100)
     Currencies.add_currency(user.id, Currencies.get_currency_by_name!("Gems").id, 500)
   end
+
+  @doc """
+  Adds the given experience to a user. If the user were to have enough resulting experience to level up,
+  it is performed automatically.
+  """
+  def add_experience(user_id, experience) do
+    user = get_user(user_id)
+    new_experience = user.experience + experience
+
+    # Level up
+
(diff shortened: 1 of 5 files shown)


ncontinanza (approved):

apps/champions/lib/champions/users.ex
```diff
@@ -88,4 +88,34 @@ defmodule Champions.Users do
     Currencies.add_currency(user.id, Currencies.get_currency_by_name!("Gold").id, 100)
     Currencies.add_currency(user.id, Currencies.get_currency_by_name!("Gems").id, 500)
   end
+
+  @doc """
+  Adds the given experience to a user. If the user were to have enough resulting experience to level up,
+  it is performed automatically.
+  """
```
agustinesco, line 95: Just a personal comment but i think examples of the function usage are really helpful for example this method receives an user_id and a amount of experience and returns either `{:ok, %User{}}` of `{:error, %Changeset{}}`

agustinesco (approved):
## LGTM 

on a side note, please add a way to tests prs through commands so it's easier to tests if everything works for the reviewer

to tests this i first added a user user 
`Champions.Users.register("agustinesco")`

then i started adding experience with the method you mentioned on the pr description

apps/champions/lib/champions/users.ex
lotuuu: Yeah I'm not too much of a fan of examples since they take up a lot of space in the code, I like specs more since they accomplish the same in much less space, but we don't have them in this project :(. I try to just specify the return types in the doc but I missed this one heh

PR #273, opened 2024-02-16
Unit progression

lotuuu:
Adds unit progression (level/tier/rank)

For the GameBackend, all that's going on are updates and deletions in the db. Restrictions and fusion logic are all responsibilities of the Champions app.

Specifically for how they were implemented for Champions, these mechanics can be a little bit tricky if you're not familiar with autobattlers. The gist of it is that you can level up your unit as much as your tier allows you, and you can tier up your unit as much as your rank allows you. Level and Tier ups share the way in which they are performed: they just cost a bunch of currencies from the user. See: [Evolution](https://www.notion.so/lambdaclass/Tiers-Evolution-3b7d9740ad384a5cbf7eaeac377083ee?pvs=4).  Ranks ups, on the other hand, are only accesible through Fusion

Fusion allows "combining" many units into one to increase its rank. Effectively it just deletes the other ones. There are pre-established rules for which units are required based on the target unit's rank. Also there are requirements based on character rank (common units cannot be fused, for example) See: [Ranks](https://www.notion.so/lambdaclass/Ranks-Fusion-1d1df44f076c4e2d81bf296ee05c4921?pvs=4).

Some integration tests have been added for the connection to the Gateway app, as well as the returned values being the expected ones. This will allow faster progress when developing new stuff, ensuring no protobuf messages are broken when adding stuff and being able to test manually faster as well (coding tests and running them has been faster for me than using the SocketTester in the console). **SocketTester will not work on the console anymore** and that's fine, we don't want a test utility to be up when running the app. If you really need it you can move the socket_tester file to the lib folder.

We add a GameBackend.Transaction module. The idea behind it is that we can keep our game apps repo-agnostic, while still being able to encapsulate game-specific behavior in transactions. This is important for operations that consume currencies and for fusion, for example.

We define `get_rank` and `get_rarity` functions in Units to be able to have a single source for these values. Of course, these are in the Champions app since they are specific to that game. Same was done in Champions.Utils for the game_id.

# To test

There are tests implemented that run through a happy path of unit progression. You can just try emulating that in the console while playing with the values and edge cases, or go through said tests line by line checking that calls make sense.


apps/game_client/test/lambda_game_backend_web/controllers/page_controller_test.exs
```diff
@@ -1,8 +1,8 @@
 defmodule GameClientWeb.PageControllerTest do
```
ncontinanza, line 1: Could this file be directly removed? Or do we need to keep it for some reason?

lotuuu: Yeah definitely

ncontinanza (approved):

apps/champions/lib/champions/units.ex
```diff
@@ -1,28 +1,356 @@
...
+    else
+      {:error, "failed"}
+    end
   end
+
+  @doc """
+  Returns whether a unit can level up. Level is blocked by tier.
+  """
+  def can_level_up(unit), do: can_level_up(unit.tier, unit.unit_level)
+  defp can_level_up(1, level) when level < 20, do: true
```
manucamejo, line 133: This smells weird, why don't we use a config file? Like a json so you don't need to redeploy when adding a new value.

apps/champions/lib/champions/units.ex
```diff
@@ -1,28 +1,356 @@
...
+    end
+  end
+
+  @doc """
+  Returns whether a unit can tier up. tier is blocked by rank.
+  """
+  def can_tier_up(unit), do: can_tier_up(unit.rank, unit.tier)
+
+  # What if a unit with level 1 and tier 3 tries to tier up? Should we block that?
+  defp can_tier_up(@star1, tier) when tier < 1, do: true
```
manucamejo, line 201: The same here

manucamejo (changes requested):

apps/champions/lib/champions/units.ex
lotuuu: Sure. Would we fetch the json on every can_level_up call? Unit level up is something that happens a lot among all users, that could result in some overheard (I don't know how costly it is to fetch it everytime). Are there alternatives to this?

lotuuu: I have created the issue for this here! https://github.com/lambdaclass/champions_of_mirra/issues/189

lotuuu:
This PR will be split into smaller, more concrete PRs

PR #274, opened 2024-02-16
Add gacha (Unit summon)

lotuuu:
Allows Gacha mechanics, which involve opening boxes to receive prizes. What we needed in CHoM was to have these prizes be just units, so that's what we implemented. Drops probabilities are decided per-rank, and that is represented in our model through the characters' ranks_dropped_in field. It's a list because we want some champions to be able to drop in many ranks. See: [Summon](https://www.notion.so/lambdaclass/Rarities-Summoning-9465b151e8144d30a92c795ce6d3fab4?pvs=4).

Implements a CurrencyCost struct and integrates it where we previously used tuples.


ncontinanza (approved):

lotuuu:
This PR conflicts with the split of the Units Progression PR, so it will be closed. Its changes will be brought back in a new PR from a different branch.

PR #279, opened 2024-02-19
Add duration to projectiles

AminArria:
This change adds `duration_ms` field to projectiles config so then when spawning them we can setup a removal of them after that timeout

Also, slight refactoring on duplicated code for `repeated_shoot`


agustinesco (dismissed):
LGTM!

agustinesco (approved):
🪨

PR #282, opened 2024-02-19
Skill delay configuration

AminArria:
Currently we had a hardcoded delay in the `cone_hit` mechanic in which we essentially delayed its execution by 300 ms. This change introduces a configurable parameter for all skills which we use to delay the execution of ALL mechanics for that skill

Now when a skill is activated we do all the side effects (add action to player, reduce stamina, etc), but the mechanic execution we delay it by X ms based on the configuration for the skill


apps/arena/priv/config.json
```diff
@@ -102,6 +107,7 @@
       "name": "dash",
       "cooldown_ms": 2000,
       "execution_duration_ms": 500,
+      "activation_delay_ms": 300,
```
manucamejo, line 110: Dashes shouldn't have the same delay as the others.
They need less delay.

manucamejo (changes requested):

apps/arena/priv/config.json
```diff
@@ -28,6 +28,7 @@
       "name": "circle_bash",
       "cooldown_ms": 2000,
       "execution_duration_ms": 500,
+      "activation_delay_ms": 300,
```
manucamejo, line 31: change to 0

apps/arena/priv/config.json
```diff
@@ -57,6 +59,7 @@
       "name": "shot",
       "cooldown_ms": 2000,
       "execution_duration_ms": 500,
+      "activation_delay_ms": 300,
```
manucamejo, line 62: change to 0

apps/arena/priv/config.json
manucamejo, line 110: change to 0

apps/arena/priv/config.json
```diff
@@ -116,6 +122,7 @@
       "name": "avenge",
       "cooldown_ms": 2000,
       "execution_duration_ms": 1000,
+      "activation_delay_ms": 300,
```
manucamejo, line 125: change to 0

apps/arena/priv/config.json
```diff
@@ -133,6 +140,7 @@
       "name": "veil_radiance",
       "cooldown_ms": 2000,
       "execution_duration_ms": 500,
+      "activation_delay_ms": 300,
```
manucamejo, line 143: change to 0

apps/arena/priv/config.json
```diff
@@ -147,6 +155,7 @@
       "name": "sneak",
       "cooldown_ms": 2000,
       "execution_duration_ms": 500,
+      "activation_delay_ms": 300,
```
manucamejo, line 158: change to 0

manucamejo (changes requested):

manucamejo (approved):

PR #284, opened 2024-02-19
Add intermediate step to expired projectiles

agustinesco:
After the merge of #279 We noticed in the client pr number [#1438](https://github.com/lambdaclass/curse_of_mirra/pull/1438) that the expired projectiles remained with the same bug, this is caused by the `game_update` deleting the expired projectile from the list without sending it with a `:EXPLODED` status not even once, so solve this we'll add the `:EXPIRED` intermediate status instead of `:EXPLODED`


BertovDev (approved):
LGTM

manucamejo (approved):

PR #285, opened 2024-02-19
Implementation for items

AminArria:
- [x] Parse items config
- [x] Spawn items
- [x] Player inventory
- [x] Pickup items 
- [x] Use items (check the 3 items planned)
  - golden clock -> recharge stamina
  - magic boots -> speed boost
  - mirra blessing -> damage immunity
- [x] Spawn items in random positions
- [ ] (optional) inventory size is configurable


martinacantaro (approved):

apps/game_client/assets/js/hooks/board_game.js
```diff
@@ -102,6 +103,11 @@ export const BoardGame = function () {
         _this.pushEvent("attack", "3");
         _this.updateDebug("key: " + key);
       }
+
+      if (event.key === "l") {
+        _this.pushEvent("use_item", "1");
+        _this.updateDebug("key: " + key);
+      }
```
AminArria, line 110: This is not working and even makes the client crash (or something like it), so any help is welcomed

agustinesco: I think this has something to do to the fact that we're not sending a direction when we try to send a skill and the server expects us to when we do `skill_params.target` in the `player:140` module right?

agustinesco (changes requested):

apps/arena/lib/arena/game/player.ex
```diff
@@ -213,6 +221,52 @@ defmodule Arena.Game.Player do
...
+        player =
+          Enum.reduce(item.effects, player, &apply_effect/2)
+
+        put_in(player, [:aditional_info, :inventory], nil)
+
+        put_in(game_state, [:players, player.id], player)
+    end
+  end
+
+  def apply_effect({:recover_stamina, recover_stamina}, player) do
```
agustinesco, line 247: I think this should be  `apply_effect(player, effect_to_apply)` so we can pipe the player to this function

apps/game_client/assets/js/hooks/board_game.js
AminArria: No, because this is its own separate thing. It doesn't go through the regular `attack` path

Also, the crash is in the web client, like either browser crashing or something in liveview

apps/arena/lib/arena/game_updater.ex
```diff
@@ -318,6 +325,56 @@ defmodule Arena.GameUpdater do
...
+
+    last_id = state.game_state.last_id + 1
+    ## TODO: make random position
+    position =
+      random_position_in_map(
+        state.game_state.external_wall.radius,
+        state.game_state.external_wall
+      )
+
+    # item_config = Enum.random(state.game_config.items)
```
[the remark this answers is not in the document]
AminArria: Yes, uncommented in 05432db

apps/arena/lib/arena/game/player.ex
AminArria: That would make sense, except this functions are used by the `Enum.reduce/3` in `user_item/2`, so the ordering is determined by that

Since this functions are exclusively used by that, I've marked as private instead of public and if the use case arises that we need them public we can refactor things to have them usable by both parts

apps/game_client/assets/js/hooks/board_game.js
AminArria: FYI since the client is crashing already, don't see any issue with merging with this issue and fixing later

agustinesco: yeah that was due to the problem i mentioned 😅 but it's a different one. That problem was solved on #312

apps/arena/lib/arena/game/player.ex
```diff
@@ -275,4 +309,23 @@ defmodule Arena.Game.Player do
...
+      self(),
+      {:remove_speed_boost, player.id, speed_boost.amount},
+      speed_boost.duration_ms
+    )
+
+    %{player | speed: player.speed + speed_boost.amount}
+  end
+
+  defp apply_effect({:damage_immunity, damage_immunity}, player) do
+    Process.send_after(self(), {:remove_damage_immunity, player.id}, damage_immunity.duration_ms)
```
agustinesco, line 328: Process rocks 🚀

apps/game_client/assets/js/hooks/board_game.js
AminArria: Finally fixed in 3d8dd09

apps/arena/lib/arena/game/player.ex
```diff
@@ -282,4 +320,35 @@ defmodule Arena.Game.Player do
     end
     |> Enum.uniq()
   end
+
+  defp apply_effect({:stamina_faster, stamina_faster}, player) do
+    change_stamina(player, 3)
```
agustinesco, line 325: I think this is not doing anything right?

apps/arena/lib/arena/game/player.ex
```diff
@@ -282,4 +320,35 @@ defmodule Arena.Game.Player do
     end
     |> Enum.uniq()
   end
+
+  defp apply_effect({:stamina_faster, stamina_faster}, player) do
```
agustinesco, line 324: This mechanic is not working correctly, it has a bug when you apply this mechanic multiple times, first the stamina is being reduced and restores correctly but if you use the same item twice you end up with less stamina cooldown permanently, i'll explain myself in a video and a diagram:

The stamina bug is due to the usage of message to restore the stamina, the second message to restore the stamina to a lowest value as the original will be received before the one to restore the stamina to a lowest value

bug stamina

Here i'm using a lot of clocks and the bug is real:
the left terminal is printing the current player stamina value


https://github.com/lambdaclass/mirra_backend/assets/106101218/b8ceb2ca-b5b5-499a-becd-3fce4f81ae35

agustinesco (changes requested):
@saugon pointed out that the clock is working weird and i've just found and error in the reduce stamina mechanic implementation, also i'm wondering should we spawn items outside of the playable area?

agustinesco (approved):
LGTM!

manucamejo (approved):

PR #286, opened 2024-02-19
Autobattle fights

lotuuu:
Implements battle simulation. Battle rules can be found [here](https://docs.google.com/spreadsheets/d/177mvJS75LecaAEpyYotQEcrmhGJWI424UnkE2JHLmyY/edit#gid=64342728) and [here](https://www.notion.so/lambdaclass/Combat-1d675e46fa28498f84d184b31f15690f?pvs=4), but I'll make an effort to summarize them in this description.

Battle happens between up to 6 units on two teams (we don't validate this in the battle simulation though, it should happen before that). Each has an assigned position which they do not move out of. Units have stats that grow with level and items: ATK used for skill damage, HP that if reaches 0 you die, SHIELD that will block damage (not designed yet). They also have energy, which they gain by casting their **basic skill** (among others). They do so whenever its cooldown is ready. When units accumuluate 500 energy, they cast their **ultimate skill**.

Skills are made of effects and a targeting strategy. Effects also have targeting strategies themselves in case we want a skill to target differently for its effects (effect 1: attack 1 *enemy*, effect 2: heal *myself*, both part of the same skill).

Battles are *simultaneous*, meaning that many characters can attack in the same step (as opposed to a turn-based combat where only one character attacks at a time). This is achieved by taking decisions in every step based on the **initial** state of the battle for that step. For example, if we have 2 units that perform the same attack in a step:
- Unit 1 recieves the initial state of the step, decides based on that (hit lowest hp enemy). Let's say target is hit and dies.
- Unit 2 recieves the *same* initial state that unit 2 did. It also decides based on it (hit lowest hp enemy). The target will be the same as Unit 1, since it was the lowest hp. It does not matter that Unit 1 will kill them with their attack, Unit 2 will hit them as well since those actions are taking place at the same time.

Effects have different targetting strategies, stats they can affect, ways they affect the target stat (additive/multiplicative) and ways their amount is interpreted (flat / % based on stat like ATK).

There are many features that are missing in this PR, but it's ready for review & merge because the general outline is finished and future implementations could expand on it. The most important ones are:
- Status effects (stun, slow...)
- Implementation of other targeting strategies
- On-kill effects
- On-getting-hit effects (receive energy)

Sample code to run a battle between two muflus:
```
  user_1 = Champions.Users.register("User1")
  user_2 = Champions.Users.register("User2")
  team1 = Enum.map([1], fn slot -> GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, unit_level: 1, tier: 1, rank: 1, selected: true, user_id: user_1.id, slot: slot}) |> elem(1) |> GameBackend.Repo.preload(character: [:basic_skill, :ultimate_skill]) end)
  team2 = Enum.map([1], fn slot -> GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, unit_level: 1, tier: 1, rank: 1, selected: true, user_id: user_2.id, slot: slot}) |> elem(1) |> GameBackend.Repo.preload(character: [:basic_skill, :ultimate_skill]) end)
  Champions.Battle.Simulator.run_battle team1, team2
```

### DIFF
--- a/apps/champions/lib/champions/battle/simulator.ex
+++ b/apps/champions/lib/champions/battle/simulator.ex
@@ -45,6 +45,7 @@ defmodule Champions.Battle.Simulator do
   require Logger
 
   @maximum_steps 500
+  @ultimate_energy_cost 500
 
   @doc """
   Runs a battle between two teams.
@@ -143,7 +144,7 @@ defmodule Champions.Battle.Simulator do
   defp can_attack(_unit), do: true
 
   # Has enough energy?
-  defp can_cast_ultimate_skill(unit), do: unit.energy >= 500
+  defp can_cast_ultimate_skill(unit), do: unit.energy >= @ultimate_energy_cost
 
   # Is cooldown ready?
   defp can_cast_basic_skill(unit), do: unit.basic_skill.remaining_cooldown <= 0
--- a/apps/champions/lib/champions/battle/simulator.ex
+++ b/apps/champions/lib/champions/battle/simulator.ex
@@ -6,7 +6,7 @@ defmodule Champions.Battle.Simulator do
   has no cooldown and it's cast whenever a unit reaches 500 energy. Energy is gained whenever the target attacks.
   The primary skill has a cooldown and it's cast when it's available if the ultimate is not.
 
-  Skills possess effects that affect other units' or oneself's stats (the on defined in the effect's `stat_affected` field).
+  Skills possess effects that affect other units' or oneself's stats (the one defined in the effect's `stat_affected` field).
 
   They have different application types (checked are implemented):
   [x] Instant - Applied once, irreversible.
@@ -26,6 +26,12 @@ defmodule Champions.Battle.Simulator do
   [ ] Backline - Heroes in slots 2 to 4
   [ ] Factions
   [ ] Classes
+
+  And different ways in which their amount is interpreted:
+  [x] Additive
+  [x] Multiplicative
+  [x] Additive & based on stat - The amount is a % of one of the caster's stats
+  [ ] Multiplicative & based on stat?
 
   Two units can attack the same unit at the same time and over-kill them. This is expected behavior that results
   from having the battle be simultaneous.
--- a/apps/champions/lib/champions/battle/simulator.ex
+++ b/apps/champions/lib/champions/battle/simulator.ex
@@ -2,43 +2,50 @@ defmodule Champions.Battle.Simulator do
   @moduledoc """
   Runs battles.
   """
+  alias GameBackend.Units.Skills.Effect
+  alias GameBackend.Units.Skill
   alias Champions.Units
   alias GameBackend.Units.Unit
 
   require Logger
 
-  @maximum_ticks 500
+  @maximum_steps 500
 
   @doc """
-  team2 = Enum.map(1..6, fn _ -> {:ok, unit} = GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, unit_level: 1, tier: 1, rank: 1, selected: true, user_id: mili_id})
+  user_1 = Champions.Users.register("User1")
+  user_2 = Champions.Users.register("User2")
+  team1 = Enum.map(1..6, fn slot -> GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, unit_level: 1, tier: 1, rank: 1, selected: true, user_id: user_1.id, slot: slot}) |> elem(1) |> GameBackend.Repo.preload(character: [:basic_skill, :ultimate_skill]) end)
+  team2 = Enum.map(1..6, fn slot -> GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, unit_level: 1, tier: 1, rank: 1, selected: true, user_id: user_2.id, slot: slot}) |> elem(1) |> GameBackend.Repo.preload(character: [:basic_skill, :ultimate_skill]) end)
+  Champions.Battle.Simulator.run_battle team1, team2
   """
   def run_battle(team_1, team_2) do
     unit_ids = team_1 ++ team_2
 
-    team_1 = Enum.into(team_1, %{}, fn unit -> create_unit(unit, 1) end)
-    team_2 = Enum.into(team_2, %{}, fn unit -> create_unit(unit, 2) end)
+    team_1 = Enum.into(team_1, %{}, fn unit -> create_unit_map(unit, 1) end)
+    team_2 = Enum.into(team_2, %{}, fn unit -> create_unit_map(unit, 2) end)
     units = Map.merge(team_1, team_2) |> IO.inspect(label: :units)
 
-    Enum.reduce_while(1..@maximum_ticks, units, fn tick, units ->
+    Enum.reduce_while(1..@maximum_steps, units, fn step, units ->
       new_state =
         Enum.reduce(unit_ids, units, fn unit, units ->
-          Logger.log(:info, "Process tick #{tick} for unit #{unit.id}")
-          process_tick(unit, units)
+          Logger.log(:info, "Process step #{step} for unit #{unit.id}")
+          IO.inspect("Process step #{step} for unit #{unit.id}")
+          process_step(unit, units)
         end)
 
-      check_winner(new_state, tick)
+      check_winner(new_state, step)
     end)
   end
 
-  def process_tick(_unit, units) do
+  def process_step(_unit, units) do
     units
   end
 
-  def check_winner(units, tick) do
+  def check_winner(units, step) do
     winner =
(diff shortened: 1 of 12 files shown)


apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -0,0 +1,248 @@
...
+    end)
+  end
+
+  defp process_step(unit, initial_step_state, current_state) do
+    new_state =
+      cond do
+        not can_attack(unit) ->
+          current_state
+
+        can_cast_ultimate(unit) ->
```
ncontinanza, line 91: What do you think of naming it `can_cast_ultimate_skill`?
Same for `can_cast_basic`.

ncontinanza (commented):
Looking good! Just a little comment about naming

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -0,0 +1,254 @@
...
+      result ->
+        {:halt, result}
+    end
+  end
+
+  # TODO: Is immobilized?
+  defp can_attack(_unit), do: true
+
+  # Has enough energy?
+  defp can_cast_ultimate_skill(unit), do: unit.energy >= 500
```
agustinesco, line 146: Maybe this could be stored in a module attribute instead being here

lotuuu: Done!

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -0,0 +1,254 @@
...
+  def run_battle(team_1, team_2, seed \\ 1) do
+    :rand.seed(:default, seed)
+    team_1 = Enum.into(team_1, %{}, fn unit -> create_unit_map(unit, 1) end)
+    team_2 = Enum.into(team_2, %{}, fn unit -> create_unit_map(unit, 2) end)
+    units = Map.merge(team_1, team_2)
+
+    # The initial_step_state is what allows the battle to be simultaneous. If we refreshed the accum on every action,
+    # we would be left with a turn-based battle. Instead we take decisions based on the state of the battle at the beggining
+    # of the step regardless of the changes that happened "before" (execution-wise) in this step.
+    Enum.reduce_while(1..@maximum_steps, units, fn step, initial_step_state ->
```
agustinesco, line 72: Maybe not in this pr but we could replace this `reduce_while` to another function with 3 parameters

`simulate_step(units, winner, steps)`

the base case would be 
`defp simulate_steps(units, winner, step) when not is_nil(winner) or step >= @maximum_steps `

lotuuu: Great! We'll create a ticket to refactor this in the short term.

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -0,0 +1,255 @@
...
+      new_state =
+        initial_step_state
+        |> Enum.map(fn {id, _unit} -> id end)
+        |> Enum.take_random(Enum.count(units))
```
agustinesco, line 76: is this doing anything?

apps/champions/lib/champions/battle/simulator.ex
lotuuu: https://github.com/lambdaclass/champions_of_mirra/issues/188

apps/champions/lib/champions/battle/simulator.ex
lotuuu: Nice catch. This was supposed to make turns for **not-simultaneous** more fair, instead of always making team_1's units go first. With synchronous battles, this doesn't make sense since all actions are decided based on the initial state of the game for that step. Which means this random has to go!

manucamejo:
To test it:
```user_1 = Champions.Users.register("User1")
  user_2 = Champions.Users.register("User2")
  team1 = Enum.map([1], fn slot -> GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, unit_level: 1, tier: 1, rank: 1, selected: true, user_id: user_1.id, slot: slot}) |> elem(1) |> GameBackend.Repo.preload(character: [:basic_skill, :ultimate_skill]) end)
  team2 = Enum.map([1], fn slot -> GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, unit_level: 1, tier: 1, rank: 1, selected: true, user_id: user_2.id, slot: slot}) |> elem(1) |> GameBackend.Repo.preload(character: [:basic_skill, :ultimate_skill]) end)
  Champions.Battle.Simulator.run_battle team1, team2```

**Review by manucamejo — APPROVED** [review]
LGTM

File: apps/champions/lib/champions/battle/simulator.ex (hunk 4 above)
Comment (agustinesco) [reply apps/champions/lib/champions/battle/simulator.ex]: oh okey, nice! if the goal of the take random was to have the units sorted randomly you can also use [`Enum.shuffle/1`](https://hexdocs.pm/elixir/1.12/Enum.html#shuffle/1)

File: apps/champions/lib/champions/battle/simulator.ex [hunk 5]
```diff
@@ -0,0 +1,254 @@
...
+    # The initial_step_state is what allows the battle to be simultaneous. If we refreshed the accum on every action,
+    # we would be left with a turn-based battle. Instead we take decisions based on the state of the battle at the beggining
+    # of the step regardless of the changes that happened "before" (execution-wise) in this step.
+    Enum.reduce_while(1..@maximum_steps, units, fn step, initial_step_state ->
+      new_state =
+        initial_step_state
+        |> Enum.map(fn {id, _unit} -> id end)
+        |> Enum.reduce(initial_step_state, fn unit_id, current_state ->
+          process_step(initial_step_state[unit_id], initial_step_state, current_state)
+        end)
```
agustinesco, line 78: I think this mapping into a reduce isn't necessary, maybe we could replace all this with

```
initial_step_state
|> Enum.reduce(initial_step_state, fn {_unit_id, unit}, current_state ->
     process_step(unit, initial_step_state, current_state)
end)
|> remove_dead_units()
|> check_winner(step)
```

What do you think? also i would specified that the state means the unit states, maybe a better naming would be `units_current_state` and `units_step_initial_state`

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -0,0 +1,254 @@
...
+  defp process_step(unit, initial_step_state, current_state) do
+    new_state =
+      cond do
+        not can_attack(unit) ->
+          current_state
+
+        can_cast_ultimate_skill(unit) ->
+          unit
+          |> cast_skill(unit.ultimate_skill, initial_step_state, current_state)
+          |> put_in([unit.id, :energy], 0)
```
agustinesco, line 94: This may be troublesome if we have a more than 500 energy stored and we cast the ultimate, for example if a unit have 600 energy and cast the ultimate it will lost 100 energy

A fix to this would be using 

`|> update_in([unit.id, :energy], fn current_energy -> current_energy - ultimate_energy_cost end)`

But i'm not sure if this was something driven by desing

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -0,0 +1,254 @@
...
+
+        can_cast_basic_skill(unit) ->
+          new_state = cast_skill(unit, unit.basic_skill, initial_step_state, current_state)
+
+          new_state
+          |> put_in(
+            [unit.id, :basic_skill, :remaining_cooldown],
+            unit.basic_skill.base_cooldown + 1
+          )
+          |> put_in([unit.id, :energy], new_state[unit.id][:energy] + 75)
```
agustinesco, line 104: this should be a module attribute too until we implement some kind of configuration file

agustinesco (approved):

apps/champions/lib/champions/battle/simulator.ex
lotuuu: You're right! I don't know about the naming though... the state of the game _is_ the state of the units. We have nothing else going on in our battle. I think those names end up being too verbose 🤔

Regardless, I've created the issue for this [here!](https://github.com/lambdaclass/mirra_backend/issues/319)

apps/champions/lib/champions/battle/simulator.ex
lotuuu: Yes! This will change eventually when we implement variable energy reception on different skills, so I won't be making a ticket specifically for this though.

apps/champions/lib/champions/battle/simulator.ex
lotuuu: Yeah as you say this is game design :P units cannot go above 500 energy, so in my implementation I just did what was simplest and let it go over 500 while dropping it back to 0. If this proved to be an issue later (which it very well may) we could just do a `min(@maximum_energy, current_energy + @energy_on_hit)` but it did not seem necessary at the time.

PR #293, opened 2024-02-20
Add Mirra champions with CSV import.

lotuuu:
Added a CSV Import function for Champions. This handles name, class, faction, quality (used to be rarity) and base stats for battles. Skills are not handled yet because they are not implemented. Eventually, we might have a JSON import for skills (instead of CSV since skills can have many effects, json is more "object-oriented" than the structured CSV format) and assign them by name in the units csv.


ncontinanza (dismissed):
LGTM!

Nico-Sanchez (approved):
Let's gooooooo 🚀

PR #295, opened 2024-02-20
AFK and level reward values

ncontinanza:
Closes https://github.com/lambdaclass/champions_of_mirra/issues/169 and https://github.com/lambdaclass/champions_of_mirra/issues/125

Implement the currency rewards of each level in the seeds, in order to satisfy the specification: https://docs.google.com/spreadsheets/d/177mvJS75LecaAEpyYotQEcrmhGJWI424UnkE2JHLmyY

Also, the missing currencies stated in the [Economy specification](https://www.notion.so/lambdaclass/Economy-7374b6b0cf924c3bb56e502102094fe8?pvs=4) were added. Notice that they differ from the ones listed in the issue #125, we're taking the Notion as source of truth.


priv/repo/seeds.exs
```diff
@@ -52,39 +52,6 @@ muflus = Characters.get_character_by_name("Muflus")
...
-          amount: -205,
-          stat_based_on: "attack",
-          amount_format: "additive",
-          # TODO: Change back to nearest
-          targeting_strategy: "random",
-          amount_of_targets: 2,
-          targets_allies: false
-        }
-        # TODO: Add stun effect
-      ]
```
Nico-Sanchez, line 87: What? 😂

priv/repo/seeds.exs
```diff
@@ -237,18 +219,32 @@ units =
...
       updated_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
     }
   end)
 
+currency_rewards =
+  currency_rewards ++
+    Enum.map(levels_without_units, fn level ->
+      %{
+        level_id: level.id,
+        amount: (10 * (15 + level.level_number - 1) * 1.025) |> round(),
```
Nico-Sanchez, line 240: We could create a private function for the amounts calculation for each currency and mention there the spreadsheet. Otherwise this calculation looks random.

Nico-Sanchez (changes requested):
Works as expected! Just a minor comment for documentation purpose

priv/repo/seeds.exs
ncontinanza: I found out that these lines were duplicated and that caused a warning, probably originated in some conflict solving. Here I removed the duplicated lines, keeping only the original one.

priv/repo/seeds.exs
ncontinanza: As the seeds file is not a module, I added a comment explaining the origin of this calculation, referencing the spreadsheet. 
If the need of delegating more behavior to private functions increases in the future we should have a utils module.

Nico-Sanchez (approved):

lotuuu (changes requested):
While we're at it let's remove the sample unit and item rewards we added to our seeds!

priv/repo/seeds.exs
Nico-Sanchez: Yeah I thought so, I was just surprised of the triplicated code lol

Nico-Sanchez (approved):

PR #298, opened 2024-02-21
[GH-297] Valtimer ultimate

### LINKED ISSUE #297: Add valthimer ultimate
When you use the valthimer ultimate skill we should spawn a circle that deals damage and pull players to the middle

agustinesco:
Closes #297 

Test this pr with [This client pr](https://github.com/lambdaclass/curse_of_mirra/pull/1473) or you can see the pools going to this [link](http://localhost:3000/board/1/valtimer) while you have the backend running and pressing the "o" key

This pr will add a couple of mechanics to be able to implement the valtimer's ultimate

When you pick valtimer and enter a match you will be able to cast the ultimate skills, this will create a pool i front of you and apply the pull and damage mechanics to players inside of it

## pool mechanic
A "pool" is a temporal entity that is placed by a skill (for now) and apply an effect to the players inside of the area, the player that casted the pool is not affected by those effects , when the pool end or the player is no longer collinding with the area the effects are removed

## Effects handling
On each game update we'll apply the effects that the player has on the `effects` field, on this pr we'll only implement two of this effects:
- Pull: This effect will pull players to the center of the pool position on each tick, the pull force is determined by the `force` field in the config
- Damage: When a player have this effect we'll deal a certain amount of damage, the damage is determines by the `damage` field and the damage will be applied ever some ms determined by the `effect_delay_ms` field


apps/arena/native/physics/src/map.rs
```diff
@@ -51,6 +51,7 @@ pub enum Category {
     Projectile,
     Obstacle,
     PowerUp,
+    Pool,
```
AminArria, line 54: What would a `Pool` be?

agustinesco: A pull is a kind of entity that does not block the player movement but apply some kind of effect, e.g. the valtimer ultimate pull the player to the center and cause damage to it

apps/arena/lib/arena/game/player.ex
```diff
@@ -3,6 +3,7 @@ defmodule Arena.Game.Player do
   Module for interacting with Player entity
   """
 
+  alias Arena.Game.Skill
```
AminArria, line 6: Duplicate line of code (see 8)

apps/arena/lib/arena/game/skill.ex
```diff
@@ -217,15 +242,94 @@ defmodule Arena.Game.Skill do
...
+        Map.put(effect, :id, last_id)
+      )
       |> put_in([:last_id], last_id)
     end)
   end
 
+  def apply_effect_mechanic(players, game_state) do
+    updated_players =
+      Map.filter(players, fn {_, player} -> Player.alive?(player) end)
+      |> Enum.map(fn {_player_id, player} ->
```
AminArria, line 257: You can replace `Enum.map/2` with `Map.new/2`, it behaves exactly the same except it remains as map so you don't need the `Map.new/1` you have in line 265

apps/arena/lib/arena/game/skill.ex
```diff
@@ -217,15 +242,94 @@ defmodule Arena.Game.Skill do
...
+
+  def apply_effect_mechanic(player, effect, game_state) do
+    now = DateTime.utc_now() |> DateTime.to_unix(:millisecond)
+
+    mechanics_to_apply =
+      effect.effect_mechanics
+      |> Map.filter(fn {_name, mechanic} ->
+        now - Map.get(mechanic, :last_application_time, 0) >= mechanic.effect_delay_ms
+      end)
+      |> Enum.map(fn {name, mechanic} ->
```
AminArria, line 278: Same as above, use `Map.new/2`

apps/arena/lib/arena/game/skill.ex
```diff
@@ -217,15 +242,94 @@ defmodule Arena.Game.Skill do
...
+      end)
+      |> Map.new()
+
+    Enum.reduce(mechanics_to_apply, player, fn mechanic, player ->
+      do_effect_mechanics(game_state, player, effect, mechanic)
+    end)
+    |> update_in([:aditional_info, :effects, effect.id, :effect_mechanics], fn mechanics ->
+      Map.merge(mechanics, mechanics_to_apply)
+    end)
+  end
```
AminArria, line 289: So in this function I'm counting 3 iterations over the same list, although performance wise certainly not a problem I think we can do better cause the code becomes confusing

In essence you have 2 behaviours, either enough time has passed and you apply the effect mechanic or you don't, so how about this

```elixir 
  def apply_effect_mechanic(player, effect, game_state) do
    now = System.monotonic_time(:millisecond)
    
    Enum.reduce(effect.effect_mechanics, player, fn {_, mechanic}, player_acc ->
      last_application_at = Map.get(mechanic, :last_application_time, 0)
      
      if now - last_application_at >= mechanic.effect_delay_ms do
        do_effect_mechanics(game_state, player_acc, effect, mechanic)
        |> Map.put_in([:aditional_info, :effects, effect.id, :effect_mechanics, name, :last_application_time], now)
      else
        player_acc
      end
    end)
  end
```

apps/arena/lib/arena/game/skill.ex
```diff
@@ -217,15 +242,94 @@ defmodule Arena.Game.Skill do
...
+        player
+
+      pool ->
+        if pool.position != player.position do
+          direction = Physics.get_direction_from_positions(player.position, pool.position)
+
+          rust_player = Physics.move_entity_to_direction(player, direction, pull_params.force)
+
+          Map.put(rust_player, :aditional_info, player.aditional_info)
+          |> Map.put(:collides_with, player.collides_with)
```
AminArria, line 304: How confident are you that this is all missing when we get the entity back?

apps/arena/priv/config.json
```diff
@@ -345,7 +367,22 @@
...
+        },
+        "damage": {
+          "damage": 4,
+          "effect_delay_ms": 400
+        }
+      }
     }
   ]
-}
+}
```
AminArria, line 388: Check your editor, you are removing the new line at the end of the file

AminArria (changes requested):

apps/arena/lib/arena/game/skill.ex
agustinesco: Wow that was very handy! thanks

apps/arena/lib/arena/game/skill.ex
agustinesco: This was something i had to do since we're calling rust to calculate the next position. When the player entity es returned it has all this missing

apps/arena/lib/arena/game/skill.ex
agustinesco: Since i reworked this functions we don't have this map new here

apps/arena/lib/arena/game/skill.ex
agustinesco: I think we should be a little concerned about in performance in this type of methods since this is executed on every tick, great comment!

apps/arena/lib/arena/game/player.ex
agustinesco: deleted!

apps/arena/lib/arena/game/skill.ex
```diff
@@ -232,15 +257,101 @@ defmodule Arena.Game.Skill do
...
+          end)
+
+        {player.id, player}
+      end)
+
+    Map.merge(players, updated_players)
+  end
+
+  def apply_effect_mechanic(player, effect, game_state) do
+    now = DateTime.utc_now() |> DateTime.to_unix(:millisecond)
```
AminArria, line 285: Can we use `System.monotonic_time/2`?

apps/arena/lib/arena/game/skill.ex
```diff
@@ -232,15 +257,101 @@ defmodule Arena.Game.Skill do
...
+        )
+      else
+        player
+      end
+    end)
+  end
+
+  defp do_effect_mechanics(game_state, player, effect, {:pull, pull_params}) do
+    Map.get(game_state.pools, effect.owner_id)
+    |> case do
```
AminArria, line 314: nit: Put the `Map.get/2` inside the `case`

apps/arena/lib/arena/game/skill.ex
```diff
@@ -232,15 +257,101 @@ defmodule Arena.Game.Skill do
...
+      nil ->
+        player
+
+      pool ->
+        if pool.position != player.position do
```
AminArria, line 319: This `if` should be another branch of the `case`

For example

```elixir
player_position = player.position
case ... do
  nil -> player
  %{pool: position} when position == player_position -> player
  pool -> 
    ...
end
```

apps/arena/lib/arena/game/skill.ex
```diff
@@ -232,15 +257,101 @@ defmodule Arena.Game.Skill do
...
+  end
+
+  defp do_effect_mechanics(game_state, player, effect, {:damage, damage_params}) do
+    Map.get(game_state.pools, effect.owner_id)
+    |> case do
+      nil ->
+        player
+
+      pool ->
+        # TODO not all effects may come from pools, maybe we should upgrade this
```
AminArria, line 339: What do you mean by this?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -372,6 +376,106 @@ defmodule Arena.GameUpdater do
...
   end
 
+  def handle_info({:remove_pool, pool_id}, state) do
+    pools =
+      state.game_state.pools
+      |> Map.delete(pool_id)
+
+    players =
+      state.game_state.players
+      |> Enum.map(fn {player_id, player} ->
```
AminArria, line 386: Replace with `Map.new/2`

apps/arena/lib/arena/game_updater.ex
```diff
@@ -824,4 +934,23 @@ defmodule Arena.GameUpdater do
...
+        |> Map.filter(fn {pool_id, _pool} -> pool_id in player.collides_with end)
+
+      Enum.each(game_state.pools, fn {pool_id, pool} ->
+        if Map.has_key?(colliding_pools, pool_id) do
+          send(
+            self(),
+            {:add_pool_effects, player_id, pool, pool.aditional_info.effects_to_apply}
+          )
+        else
+          send(self(), {:remove_pool_effects, player_id, pool_id})
```
AminArria, line 951: You do a `send/2` because you want to delay the addition/removal of the effects by one tick (on next one)?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -824,4 +934,23 @@ defmodule Arena.GameUpdater do
...
   end
+
+  defp handle_pools(game_state) do
+    Enum.each(game_state.players, fn {player_id, player} ->
+      colliding_pools =
+        game_state.pools
+        |> Map.filter(fn {pool_id, _pool} -> pool_id in player.collides_with end)
+
+      Enum.each(game_state.pools, fn {pool_id, pool} ->
+        if Map.has_key?(colliding_pools, pool_id) do
```
AminArria, line 945: You don't need the `colliding_pools` variable, you can replace this `if` like

```elixir
if pool_id in player.collides_with do
  ...
else
  ...
end
```

AminArria (changes requested):

apps/arena/lib/arena/game/skill.ex
agustinesco: I meant that maybe in the future we'll have another effects that applies the :damage mechanic and we may need to handle them differently

apps/arena/lib/arena/game_updater.ex
agustinesco: Done!

apps/arena/lib/arena/game_updater.ex
agustinesco: i made the change! the colliding pools was only for this check

apps/arena/lib/arena/game_updater.ex
agustinesco: i found the message easier to do to remove the pool effects from a player, i found another walk around to avoid creating a lot of messages but im'n not entirely happy with the outcome, can you give me a helping hand when you've spare time pls?

apps/arena/lib/arena/game/skill.ex
agustinesco: done! it's much better

apps/arena/lib/arena/game/skill.ex
agustinesco: done!

apps/arena/lib/arena/game/skill.ex
agustinesco: hmmm is there any benefit of using System.monotonic_time over the utc now?

apps/arena/lib/arena/game/skill.ex
AminArria: Yeah, that would be an open discussion if we want to specialize (`damage_pool` and `damage`) or we want to keep it generic, but regardless something for when the time comes

apps/arena/lib/arena/game_updater.ex
AminArria: It's fine either way, was only wondering if it was a "product" decision or an implementation detail

apps/arena/lib/arena/game/skill.ex
AminArria: Its way faster, a quick benchmark in my laptop

```elixir
Benchee.run(
  %{
    "DateTime.to_unix/2" => fn -> DateTime.utc_now() |> DateTime.to_unix(:millisecond) end,
    "System.monotonic_time/1" => fn -> System.monotonic_time(:millisecond) end
  }
)
```

Results
```
Name                              ips        average  deviation         median         99th %
System.monotonic_time/1       27.56 M       36.28 ns  ±7595.42%       33.40 ns       45.90 ns
DateTime.to_unix/2             2.81 M      356.43 ns  ±8331.29%         292 ns         459 ns

Comparison: 
System.monotonic_time/1       27.56 M
DateTime.to_unix/2             2.81 M - 9.82x slower +320.14 ns
```

agustinesco: wow! speaking about upgrades. Should be fixed now, also i've noticed that i was adding the last application time to the effect instead of  to each specific mechanic when i was adding the effects to the players.

AminArria (approved):
LGTM

AminArria (dismissed):

apps/arena/lib/arena/game_updater.ex
```diff
@@ -911,6 +937,49 @@ defmodule Arena.GameUpdater do
     end
   end
 
+  defp handle_pools(%{players: players} = game_state, game_config) do
+    Enum.reduce(players, game_state, fn {_player_id, player}, game_state ->
+      Enum.reduce(game_state.pools, game_state, fn {pool_id, pool}, acc ->
```
Nico-Sanchez, line 942: I think we can avoid this second reduce by just taking the pools we collided with (player.collides_with) and then adding those effects and removing the others.

Nico-Sanchez (changes requested):
Nice work @agustinesco 

Let's refactor the code a bit and we're ready to go 🚀

apps/arena/lib/arena/game/skill.ex
```diff
@@ -232,15 +257,101 @@ defmodule Arena.Game.Skill do
...
+      nil ->
+        player
+
+      %{position: pool_position} when pool_position == player.position ->
+        player
+
+      pool ->
+        direction = Physics.get_direction_from_positions(player.position, pool.position)
+
+        rust_player = Physics.move_entity_to_direction(player, direction, pull_params.force)
```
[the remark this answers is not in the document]
agustinesco: Yeah this was something left from an iteration and i forgot to delete it 😅 should be fixed now

apps/arena/lib/arena/game_updater.ex
agustinesco: The thing is that we need to know certainly which effects from the player we should remove, so we need to know if a player collides with a pool or not

apps/arena/lib/arena/game_updater.ex
```diff
@@ -355,7 +372,11 @@ defmodule Arena.GameUpdater do
     {:reply, :ok, %{state | game_state: game_state}}
   end
 
-  def handle_call({:attack, player_id, skill_key, skill_params, timestamp}, _from, state) do
+  def handle_call(
+        {:attack, player_id, skill_key, skill_params, timestamp},
+        _from,
+        state
+      ) do
```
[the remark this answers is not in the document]
agustinesco: hmmm i'm not sure, how do i check that?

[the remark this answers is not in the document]
Nico-Sanchez: Nevermind. We already have a PR to fix that 😛 
It's handled by the formatters.exs file. See. -> #350

Nico-Sanchez (dismissed):

Nico-Sanchez (approved):

Nico-Sanchez (approved):

ncontinanza (approved):
LGTM!

PR #302, opened 2024-02-22
Add countdown to start battle after that

manucamejo:


agustinesco (commented):
LGTM but the client pr is in draft, will approve and merge when the client pr is finshed

agustinesco (approved):

PR #303, opened 2024-02-22
[gh-308] valtimer antimater basic

### LINKED ISSUE #308: Add valtimer basic skill
The valtimer basic skill will be a simple projectile that on collision or expiration creates a circle hit of damage, the projectile will not deal damage alone

agustinesco:
Closes #308

Test this pr with [This client pr](https://github.com/lambdaclass/curse_of_mirra/pull/1465) if not merged

### On explode mechanics
Now projectiles will have a new attribute called `on_explode_mechanics` that's a map of mechanics to be triggered when the projectile explodes, that is:
- The projectile expired
- The projectile collided with the external wall
- The projectile collided with a player

Also this pr will change the function `do_mechanic`a bit, so you can receive and entity plus the owner player, so we can trigger the mechanics from the entity location


apps/arena/lib/arena/game/skill.ex
```diff
@@ -5,17 +5,19 @@ defmodule Arena.Game.Skill do
   alias Arena.Entities
   alias Arena.Game.Player
 
-  def do_mechanic(game_state, player, mechanics, skill_params) when is_list(mechanics) do
+  def do_mechanic(game_state, entity, player, mechanics, skill_params) when is_list(mechanics) do
```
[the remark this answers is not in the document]
agustinesco: Yeah! i think the one parameter that's not necessary is `player` but since we're using it for some things i had to keep it. With your comment i think i can get to something where player is not needed as parameter, thanks! also really good comments

apps/arena/lib/arena/game_updater.ex
```diff
@@ -764,4 +765,20 @@ defmodule Arena.GameUpdater do
       end
     end)
   end
+
+  defp execute_on_explode_mechanics(game_state, projectiles) do
```
AminArria, line 769: Not for this PR, but we should probably start thinking on a `projectile.ex` module

AminArria (approved):
LGTM, but haven't tested it in a game

apps/arena/lib/arena/game_updater.ex
agustinesco: 100% agreed, an effect module could be handy too

PR #311, opened 2024-02-22
Autoaim fixes

AminArria:
- Autoaim will only work if the skill is configured for it, otherwise it will use the player's current direction
- Autoaim will ignore invisible players


agustinesco (approved):
LGTM!

PR #314, opened 2024-02-22
Send skill configuration info

AminArria:
Sends skill information regarding range and angle

Client PR https://github.com/lambdaclass/curse_of_mirra/pull/1483


agustinesco (approved):
LGTM! tested and the messages are reaching the client with the information

manucamejo (approved):

PR #317, opened 2024-02-23
[gh-316] Actually update the kill count on kill

### LINKED ISSUE #316: Fix kill count
We need to display kill count on the metric corner

agustinesco:
Closes #316

Tests with this [Client branch](https://github.com/lambdaclass/curse_of_mirra/pull/1490)

We'll add a kill count was not being updated every time a player killed someone, this pr will make the counter actually increase


saugon (approved):
Working nicely!

PR #318, opened 2024-02-23
add can pick position field to skills

manucamejo:
Add can pick position field to skills and use that to decide if normalize or not the direction

### DIFF
--- a/apps/arena/lib/arena/game/skill.ex
+++ b/apps/arena/lib/arena/game/skill.ex
@@ -263,11 +263,19 @@ defmodule Arena.Game.Skill do
   def maybe_auto_aim(%{x: x, y: y}, skill, player, entities) when x == 0.0 and y == 0.0 do
     case skill.autoaim do
       true -> Physics.nearest_entity_direction(player, entities)
-      false -> player.direction |> Utils.normalize()
+      false -> player.direction |> maybe_normalize(not skill.can_pick_destination)
     end
   end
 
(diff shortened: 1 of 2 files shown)


apps/arena/lib/arena/game/skill.ex
```diff
@@ -263,11 +263,19 @@ defmodule Arena.Game.Skill do
   def maybe_auto_aim(%{x: x, y: y}, skill, player, entities) when x == 0.0 and y == 0.0 do
     case skill.autoaim do
       true -> Physics.nearest_entity_direction(player, entities)
-      false -> player.direction |> Utils.normalize()
+      false -> player.direction |> maybe_normalize(not skill.can_pick_destination)
```
agustinesco, line 266: I don't think this is the correct way of using this boolean, imo negating a boolean in a match leads to confusion 

```
maybe_normalize(skill.can_pick_destination)

defp maybe_normalize(direction, false) do
    Utils.normalize(direction)
  end

  defp maybe_normalize(direction, true) do
    direction
  end
```

agustinesco (changes requested):

apps/arena/lib/arena/game/skill.ex
agustinesco: also i'm not a big fan of using case for booleans like the autoaim case

agustinesco (approved):
## LGTM!

PR #322, opened 2024-02-23
Add stamina_cost to skill and use it when triggering skills

AminArria:


agustinesco (changes requested):
There's an error where if you try to use a skill and doesn't have enough stamina while you try to keep walking you'll keep walking indefinetly

Disclaimer: this is not happening on main


https://github.com/lambdaclass/mirra_backend/assets/106101218/41096617-79a2-47b1-9afc-40308f4fe1e7

manucamejo (approved):

PR #323, opened 2024-02-23
Unit Level up

lotuuu:
Implements Unit level up. Level up costs the user some currency (now pseudo hardcoded to grow based on the current level but should come in a config file later on). Unit level is blocked by tier. The exact rules for that will later on come in a config file, for now we just pattern match the unit's params.

Additional changes:
- JSON encoder @derive attributes have been removed from schemas since they are no longer in use.
- We add a GameBackend.Transaction module. The idea behind it is that we can keep our game apps repo-agnostic, while still being able to encapsulate game-specific behavior in transactions. This is used for operations that consume currencies in our game design, for example.
- Minor refactors and fixes to some units functions

This PR's changes were originally developed [here](https://github.com/lambdaclass/mirra_backend/pull/273). The Gateway Protobuf file was not split up like the rest of the changes for faster development.

# To test

### Setting up a user (comes with some units and 100 gold)
`{:ok, user} = Champions.Users.register "User"`
`unit = hd(user.units)`

### Leveling up a unit
`{:ok, unit_and_currency} = Champions.Units.level_up user.id, unit.id`

`unit_and_currency` should be a `%{unit: _, :user_currency: _}` map. The `unit` should have the level increased by 1, the `user_currency` should have its amount diminished by the unit's previous level squared (`calculate_level_up_cost`'s formula).
Subsequent calls should behave the same way until an `{:error, :cant_afford}` tuple is returned instead (user has run out of gold!)

### Giving a user more gold
`{:ok, _} = GameBackend.Users.Currencies.add_currency(user.id, GameBackend.Users.Currencies.get_currency_by_name!("Gold").id, 9999999)`

### Checking tier block
Now the user has more than enough gold to level up the unit, however it should not be able to get past level 20 because of its tier:

```
Enum.map(1..20, fn _ -> {success, return} = Champions.Units.level_up(user.id, unit.id)
return = case return do
%{} -> :unit_and_currency
error -> error
end
{success, return}
end)
```

Depending on the unit's level, you should eventually start seeing `error: :cant_level_up` at some point in the list.
If you force-change the unit's tier:
`{:ok, unit} = GameBackend.Units.get_unit(unit.id)`
`{:ok, unit} GameBackend.Units.update_unit(unit, %{tier: 2})`
And run the map function again, you should now see the full list as `:ok` tuples (since we can perform exactly 20 more levelups).

### DIFF
--- a/apps/champions/lib/champions/items.ex
+++ b/apps/champions/lib/champions/items.ex
@@ -26,7 +26,7 @@ defmodule Champions.Items do
   def level_up(user_id, item_id) do
     with {:item, {:ok, item}} <- {:item, Items.get_item(item_id)},
          {:item_owned, true} <- {:item_owned, item.user_id == user_id},
-         {level_up_currency_id, level_up_cost} = calculate_level_up_cost(item),
+         [{level_up_currency_id, level_up_cost}] = calculate_level_up_cost(item),
          {:can_afford, true} <-
            {:can_afford, Currencies.can_afford(user_id, level_up_currency_id, level_up_cost)} do
       Items.level_up(item, level_up_currency_id, level_up_cost)
@@ -48,5 +48,5 @@ defmodule Champions.Items do
   end
 
   defp calculate_level_up_cost(item),
-    do: {Currencies.get_currency_by_name!("Gold").id, item.level |> Math.pow(2) |> round()}
+    do: [{Currencies.get_currency_by_name!("Gold").id, item.level |> Math.pow(2) |> round()}]
 end
--- a/apps/champions/lib/champions/units.ex
+++ b/apps/champions/lib/champions/units.ex
@@ -90,7 +90,7 @@ defmodule Champions.Units do
     do: [
       %CurrencyCost{
         currency_id: Currencies.get_currency_by_name!("Gold").id,
-        amount: unit.level |> Math.pow(2) |> round()
+        amount: unit.unit_level |> Math.pow(2) |> round()
       }
     ]
 
@@ -99,7 +99,7 @@ defmodule Champions.Units do
 
   This will eventually be provided by configuration files.
   """
-  def can_level_up(unit), do: can_level_up(unit.tier, unit.level)
+  def can_level_up(unit), do: can_level_up(unit.tier, unit.unit_level)
   defp can_level_up(1, level) when level < 20, do: true
   defp can_level_up(2, level) when level < 40, do: true
   defp can_level_up(3, level) when level < 60, do: true
--- a/apps/champions/lib/champions/units.ex
+++ b/apps/champions/lib/champions/units.ex
@@ -31,12 +31,15 @@ defmodule Champions.Units do
 
   @doc """
   Sets a unit as unselected for a user. Clears the `slot` field.
+
   Returns `:not_found` if unit doesn't exist or if it's now owned by the user.
   Returns the unit's new state if succesful.
   """
   def unselect_unit(user_id, unit_id), do: Units.unselect_unit(user_id, unit_id)
 
+  ############
   # Level Up #
+  ############
 
   @doc """
   Levels up a user's unit and substracts the currency cost from the user.
--- a/apps/champions/lib/champions/units.ex
+++ b/apps/champions/lib/champions/units.ex
@@ -20,6 +20,11 @@ defmodule Champions.Units do
       else: Units.select_unit(user_id, unit_id, slot)
   end
 
+  @doc """
+  Sets a unit as unselected for a user. Clears the `slot` field.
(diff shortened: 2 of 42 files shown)


apps/champions/lib/champions/units.ex
```diff
@@ -1,30 +1,118 @@
...
+  # Level Up #
+  ############
+
+  @doc """
+  Levels up a user's unit and substracts the currency cost from the user.
+
+  Returns `{:error, :not_found}` if unit doesn't exist or if it's not owned by user.
+  Returns `{:error, :cant_afford}` if user cannot afford the cost.
+  Returns `{:error, :cant_level_up}`if unit cant level up due to tier restrictions.
+  Returns `{:ok, unit: %Unit{}, user_currency: %UserCurrency{}}` if succesful.
```
agustinesco, line 50:
```suggestion
  Returns `{:ok, %{unit: unit, user_currency: user_currency}}` if succesful.
```

apps/game_backend/lib/game_backend/users/currencies.ex
```diff
@@ -86,10 +100,41 @@ defmodule GameBackend.Users.Currencies do
...
+  @doc """
+  Substracts all CurrencyCosts from the user.
+  Returns {:ok, results} or {:error, "failed"} tuples so it can be used on transactions.
+  """
+  def substract_currencies(_user_id, []), do: {:ok, []}
+
+  def substract_currencies(user_id, costs) do
+    result =
+      Enum.map(costs, fn %CurrencyCost{currency_id: currency_id, amount: cost} ->
+        add_currency(user_id, currency_id, -cost)
```
agustinesco, line 128: If this fails in the middle of the map execution you may substract a currency and do not give the user the thing they wanted, not mandatory but you could add al the substractions to a multi with reduce instead of mapping the results

lotuuu: Right, the idea is for this to be called inside a multi like you say (check `Units.level_up`). Do you think we should wrap this in a multi too? We'd have a multi inside a multi, I think that's fine right?

lotuuu: But there you're missing the Repo.transaction() call, making the `substract_currencies` function do nothing on its own... What do you think about just clarifying that this can be used inside a multi to avoid this issue in the function doc instead?

agustinesco (approved):
LGTM!

apps/champions/lib/champions/items.ex
```diff
@@ -48,5 +48,5 @@ defmodule Champions.Items do
   end
 
   defp calculate_level_up_cost(item),
-    do: {Currencies.get_currency_by_name!("Gold").id, item.level |> Math.pow(2) |> round()}
+    do: [{Currencies.get_currency_by_name!("Gold").id, item.level |> Math.pow(2) |> round()}]
```
Nico-Sanchez, line 51: Why this change? Does it matter if the tuple is inside a list or not?

apps/game_backend/lib/game_backend/transaction.ex
```diff
@@ -0,0 +1,14 @@
...
+  """
+
+  alias GameBackend.Repo
+
+  @doc """
+  Run a multi transaction.
+  """
+  def run(multi) do
+    Repo.transaction(multi)
+  end
```
Nico-Sanchez, line 13: What is the motivation of doing this instead of just use the repo transaction in the multi?

Nico-Sanchez (changes requested):
Nice work @lotuuu 🪨 

Left some questions to help me understand some decisions in the design and some comments about code readability too.

apps/game_backend/lib/game_backend/users/currencies/currency_cost.ex
```diff
@@ -0,0 +1,21 @@
...
+  """
+  use GameBackend.Schema
+  import Ecto.Changeset
+  alias GameBackend.Users.Currencies.Currency
+
+  @primary_key false
+  embedded_schema do
+    belongs_to(:currency, Currency)
+    field(:amount, :integer)
+  end
```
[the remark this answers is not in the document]
lotuuu: Yes, but it's also used in upcoming PRs for things like summon too! Basically, anything that implies a cost.

apps/champions/lib/champions/items.ex
lotuuu: It does because we want to use substract_currencies, which takes in a list! This function is just a placeholder until we get actual level up cost logic anyways, what matters is that it follows the data type we'll want to have when we do implement it

Nico-Sanchez (approved):

PR #328, opened 2024-02-26
Rename Unit params

lotuuu:
Renames `unit_level` to `level` and `level` assoc to `campaign_level` for clarity.

# To test
Check there weren't any missed changes


agustinesco (dismissed):
LGTM

Nico-Sanchez (commented):
The `:unit_level` change to `:level` looks nice and more alike to other entities (i.e. `items.level`).

I think this PR is ready to merge. However, I would like to point out that the `Level` schema confuses me with the entities' :level fields.

Can you explain to me what's a `Campaign.Level`? Maybe we could rename it in the future to avoid this confusion.

lotuuu:
mmh well campaign levels are just that, different levels (think level 1 or level 2 in a Mario game) that players can fight and attempt to pass.
In Champions. this is Campaign 1, in which you can see Level 1 is unlocked and levels 2 through 5 are locked.
image


They have nothing to do with unit levels of course, so the naming can be hard to understand, but at the same time I cannot think of a better naming for either of the "level"s :(

I thought about "Stage" for campaign levels but I think Stage sounds more like a Campaign (a combination of levels) :/

Nico-Sanchez (approved):

agustinesco (approved):
LGTM!

PR #329, opened 2024-02-26
Unit tier up

lotuuu:
Adds unit tier up. Tier ups are needed for units to increase their level further after a certain point. An issue has been created already to [block units that aren't max level from tiering up](https://github.com/lambdaclass/champions_of_mirra/issues/208). Tier ups are blocked by unit rank (which is increased through to-be-implemented fusion).

# To test

### Setup

Follow the steps in [this](https://github.com/lambdaclass/mirra_backend/pull/323) PR's *To test* section until the point where we force-update its tier. Instead we'll do that through our new functions.

### Try to tier up

You will try to tier up the unit by running
`Champions.Units.tier_up user.id, unit.id`
This will instead return an `{:error, :cant_tier_up}` tuple. The reason for this is that our unit's rank is not high enough (which means `can_tier_up` is working as expected).

### Actually tier up

To brute-force a new rank to the unit just run:
`{:ok, unit} = GameBackend.Units.get_unit(unit.id)`
`{:ok, unit} = GameBackend.Units.update_unit(unit, %{rank: 2})`
Now tier up should work, returning an `{:ok, unit_and_currency}` tuple where `unit_and_currency.unit.tier` is 2 instead of 1 .
`{:ok, unit_and_currency} = Champions.Units.tier_up user.id, unit_id`


apps/champions/lib/champions/units.ex
```diff
@@ -14,6 +13,26 @@ defmodule Champions.Units do
...
+
+  def get_rank(:star1), do: @star1
+  def get_rank(:star2), do: @star2
+  def get_rank(:star3), do: @star3
+  def get_rank(:star4), do: @star4
+  def get_rank(:star5), do: @star5
+  def get_rank(:illumination1), do: @illumination1
+  def get_rank(:illumination2), do: @illumination2
+  def get_rank(:illumination3), do: @illumination3
+  def get_rank(:awakened), do: @awakened
```
agustinesco, line 34: These functions are unused, are they for the future?

agustinesco (approved):
LGTM! let's not forget to replace the logic from `can_tier_up` when we implement the configuration file

apps/champions/lib/champions/units.ex
agustinesco: they are from the fuse pr

agustinesco:
> That's not true! It only happens for the Star ranks, at Illumination it starts skipping by two. I think the verbose approach is clearer

right! i totally missed that for higher ranks the comparison changes

apps/champions/lib/champions/units.ex
```diff
@@ -114,6 +133,80 @@ defmodule Champions.Units do
...
+      {:can_afford, false} -> {:error, :cant_afford}
+    end
+  end
+
+  @doc """
+  Returns whether a unit can tier up. tier is blocked by rank.
+  """
+  def can_tier_up(unit), do: can_tier_up(unit.rank, unit.tier)
+
+  # What if a unit with level 1 and tier 3 tries to tier up? Should we block that?
```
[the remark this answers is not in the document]
lotuuu: Done! [Here's](https://github.com/lambdaclass/champions_of_mirra/issues/227) the issue. I've updated the comment as well.

Nico-Sanchez (approved):

agustinesco (approved):

PR #332, opened 2024-02-26
Unit fusion

lotuuu:
Add unit fusion. Unit fusion is performed to rank up a unit. Fusion involves consuming units and increasing the rank of the target by one. Requirements of units to be consumed are set through the `same_character_requirements` and `same_faction_requirements` functions. These requirements will later be served through configuration files, but until then this will do.

# To test 

### Setup

Follow the steps on https://github.com/lambdaclass/mirra_backend/pull/323 to create a unit. You only need to create it, don't worry about leveling it up. The only thing it needs it to have a character with quality = 3 (:epic)

###

Create some more units! Depending on which requirement you want to test, you will need to create a different amount of same-faction and same-character units (with different ranks). Check the `same_character_requirements` and `same_faction_requirements` functions or consult this [beautiful doc](https://www.notion.so/lambdaclass/Ranks-Fusion-1d1df44f076c4e2d81bf296ee05c4921?pvs=4#c26ec89d1f64417d8f66a3d5cce33919). You can create your units with:
```
{:ok, consumed_unit_n} = GameBackend.Units.insert_unit(%{character_id: character_id, user_id: user_id, level: 1, tier: 1, selected: false, slot: nil, rank: 1})
```
Get your character_id by either getting one that already exists (change Muflus to whatever you want from the seeds)
```
character = GameBackend.Units.Characters.get_character_by_name("Muflus")
```
...or creating one from target unit's character's faction:
```
{:ok, character} = GameBackend.Units.Characters.insert_character(%{faction: unit.character.faction, rarity: 3, game_id: 2, active: true})
```
Remember that you _can_ meet same-faction requirements with a unit with the same character (after all if character is the same, faction is the same).
An example fusion:
```
target: Rank 5 Muflus
consumed units:
- Rank 5 Muflus (same char 1)
- Rank 5 AnotherArabanChampion (same faction 1)
- Rank 5 AnotherArabanChampion (same faction 2)
- Rank 5 Muflus (same faction 3)
- Rank 5 AnotherArabanChampion (same faction 4)
```

### Fuse!
Fuse the units by running
```
{:ok, unit} = Champions.Units.fuse(user.id, unit.id, [consumed_unit_1, consumed_unit_2, ...........])
```

You should be returned the unit with a new rank.

If you get your user's units (`GameBackend.Units.get_units(user.id)`), you should not see any of the consumed units you passed to `fuse`.


apps/champions/lib/champions/units.ex
```diff
@@ -207,6 +217,151 @@ defmodule Champions.Units do
...
+
+      {:consumed_units_valid, false} ->
+        {:error, :consumed_units_invalid}
+    end
+  end
+
+  defp delete_consumed_units(unit_ids) do
+    {amount_deleted, _return} = Units.delete_units(unit_ids)
+
+    if Enum.count(unit_ids) == amount_deleted, do: {:ok, amount_deleted}, else: {:error, "failed"}
```
agustinesco, line 276: does this operation rollback the change in the db if it fails?

lotuuu: Yup! You can check yourself by modifying the delete_units function to always return a wrong number:

```
def delete_units(unit_ids) do
  Repo.delete_all(from(u in Unit, where: u.id in ^unit_ids))
  {-1, nil}
end
```

Calling a "good" fusion (a simple one is 5-star Muflus as a target and 5 5-star Muflus as consumed units)

And checking the units count matches:
`user.id |> GameBackend.Users.get_user() |> Map.get(:units) |> Enum.count()` (run before and after the `Units.fuse()`)

apps/champions/lib/champions/units.ex
```diff
@@ -207,6 +217,151 @@ defmodule Champions.Units do
...
+
+  @doc """
+  Returns whether a unit can rank up, based on its current rank and its character's quality.
+  """
+  def can_rank_up(%Unit{rank: rank, character: %Character{quality: @epic}}), do: rank < @awakened
+
+  def can_rank_up(%Unit{rank: rank, character: %Character{quality: @rare}}),
+    do: rank < @illumination2
+
+  def can_rank_up(_unit), do: false
```
agustinesco, line 359: Dont this need to be private?

lotuuu: No! Later on we will need CanLevelUp, CanRankUp and CanTierUp messages for the frontend to display the appropriate buttons. If you think that we should make them private until then though, that's perfect

apps/champions/lib/champions/units.ex
```diff
@@ -207,6 +217,151 @@ defmodule Champions.Units do
...
+          Enum.find(
+            list,
+            &(&1.character.faction == unit.character.faction and &1.rank == same_faction_rank)
+          )
+
+        if is_nil(same_faction), do: raise("not enough of same faction")
+        if same_faction.user_id != unit.user_id, do: raise("consumed unit not owned")
+
+        List.delete(list, same_faction)
+      end)
```
agustinesco, line 345: I'm not a huge fan of this approach of raising to indicates that the operation was not succefully, why not using the same approach as fuse to check this? if the objective is just to return a boolean when the user have enough units it should be straightfoward

```
  defp validate_consumed_units(unit, unit_list) do
    with {:removed_same_unit, {:ok, removed_same_unit}} <-
           {:removed_same_unit,
            remove_same_unit(unit, unit_list, same_character_requirements(unit))},
         {:removed_same_faction, {:ok, _removed_same_faction}} <-
           {:removed_same_faction,
            remove_same_faction(unit, removed_same_unit, same_faction_requirements(unit))} do
      true
    else
      _ -> false
    end
  end
```

And the two removals would be:

```

  defp remove_same_unit(unit, unit_list, {same_character_amount, same_character_rank}) do
    same_character_list =
      Enum.count(
        unit_list,
        &(&1.character_id == unit.character_id and &1.rank == same_character_rank)
      )

    if Enum.count(same_character_list) >= same_character_amount do
      deleted_same_character =
        Enum.reduce(same_character_list, unit_list, fn same_character, list ->
          List.delete(list, same_character)
        end)

      {:ok, deleted_same_character}
    else
      {:error, :not_enough_same_character}
    end
  end

  defp remove_same_faction(unit, unit_list, {same_faction_amount, same_faction_rank}) do
    same_faction_list =
      Enum.count(
        unit_list,
        &(&1.character.faction == unit.character.faction and &1.rank == same_faction_rank)
      )

    if Enum.count(same_faction_list) >= same_faction_amount do
      deleted_same_faction =
        Enum.reduce(same_faction_list, unit_list, fn same_faction, list ->
          List.delete(list, same_faction)
        end)

      {:ok, deleted_same_faction}
    else
      {:error, :not_enough_same_faction}
    end
  end
```

What do you think?

apps/game_backend/lib/game_backend/units.ex
```diff
@@ -75,6 +75,14 @@ defmodule GameBackend.Units do
   """
   def get_units(), do: Repo.all(Unit) |> Repo.preload([:character, :user, :items])
 
+  @doc """
+  Gets all units from ids in a list.
+  """
+  def get_units_by_ids(unit_ids) when is_list(unit_ids),
+    do:
+      Repo.all(from(u in Unit, where: u.id in ^unit_ids))
+      |> Repo.preload([:character, :user, :items])
```
agustinesco, line 84: The preload should be inside the first query:
```suggestion
      Repo.all(from(u in Unit, where: u.id in ^unit_ids, preload: [:character, :user, :items]))
```

apps/game_backend/lib/game_backend/users/currencies.ex
```diff
@@ -155,4 +155,20 @@ defmodule GameBackend.Users.Currencies do
...
     end
   end
+
+  @doc """
+  Gets how much a user has of a given currency by its name.
+  """
+  def get_amount_of_currency_by_name!(user_id, currency_name),
+    do:
+      user_id
+      |> get_amount_of_currency(get_currency_by_name!(currency_name).id)
```
agustinesco, line 165: We're doing two queries here instead of one why not doing it all at once?
```
get_amount_of_currency_by_name!(currency_name)

def get_amount_of_currency_by_name(currency_name) do
Repo.one(
        from(uc in UserCurrency,
          join: currency in assoc(uc, :currency),
          where: uc.user_id == ^user_id and currency.name == ^currency_name,
          select: uc.amount
        )
      ) || 0
end
```

Do you think this is better?

agustinesco (commented):
We should update the `add_sample_units` since it's using a rank field now

apps/champions/lib/champions/units.ex
```diff
@@ -207,6 +217,151 @@ defmodule Champions.Units do
...
+  # Fusion #
+  ##########
+
+  @doc """
+  Consume a list of units that meet specific rank and character requirements based on the target
+  unit's rank in order to increase it.
+
+  Returns `{:ok, unit}` or `{:error, reason}`.
+  """
+  def fuse(user_id, unit_id, consumed_units_ids) do
```
agustinesco, line 230: We should ensure that the unit_id is not inside the consumed_units_ids list since it should not count as a consume valid unit

apps/game_backend/lib/game_backend/users/currencies.ex
lotuuu: Yes it's definitely better!

apps/champions/lib/champions/units.ex
lotuuu: The idea behind the raises was to stop execution early whenever we found something that was wrong. Is there something inherently bad about raises?

Nevertheless I've taken your approach and fixed it up a little bit, let me know if you think [this](https://github.com/lambdaclass/mirra_backend/pull/332/commits/68028f1770aee879ffb59cf34b23097220bb7836) works and is clearer than what was previously implemented!

agustinesco: hmm i don't like using raises to stop execution because we won't know if the stop was to the player not having enough champions or we have something wrong in the code, for example we change the name of the character in units from `unit.character` to `unit.chara`, the code will rise when we try to do `unit.character.faction` but we'll never know if the rising was due to the code failing or by our check stoping execution

I you wan't to achieve the execution stopping earlier i think there're better approaches like using recursion or `Enum.reduce_while/3`

lotuuu: Right, or we could have caught the exact custom errors we were raising in our rescue too. I've implemented it with reduce_while like you said!

agustinesco: right! yes, handling specific errors could be another way to interrump execution, but i'd rater let the program crash when it should and stop the execution when we wan't explicitly. Maybe you can made my change my mind

apps/champions/lib/champions/units.ex
```diff
@@ -207,6 +217,179 @@ defmodule Champions.Units do
...
+
+    validate_units(
+      unit,
+      unit_list,
+      same_character_amount,
+      same_character_rank,
+      same_faction_amount,
+      same_faction_rank
+    )
+  end
```
agustinesco, line 298: I think this function is not required, it's just a intermediate function between `fuse` and `validate_units`

apps/game_backend/lib/game_backend/units/unit.ex
```diff
@@ -41,12 +41,12 @@ defmodule GameBackend.Units.Unit do
...
-    |> validate_required([:level, :tier, :selected, :character_id])
+    |> validate_required([:level, :selected, :character_id])
   end
 
   @doc """
   Changeset for when updating a unit.
   """
   def update_changeset(unit, attrs),
-    do: cast(unit, attrs, [:selected, :slot, :rank, :level, :tier])
+    do: cast(unit, attrs, [:selected, :slot, :level, :tier, :rank])
```
agustinesco, line 51: This is not changing anything right? 🤔

lotuuu: yeah it's just a prettifying change :P

apps/champions/lib/champions/units.ex
lotuuu: True

apps/game_backend/lib/game_backend/units/unit.ex
agustinesco: ✨

agustinesco (approved):
LGTM! the pr worked as a charm, i tried fusing multiple times and got the expected results, the units are being consuming correctly and the main one is ranking up just fine! great job 🪨

Nico-Sanchez (changes requested):

apps/champions/lib/champions/units.ex
```diff
@@ -207,6 +217,160 @@ defmodule Champions.Units do
...
+    with {:unit, {:ok, unit}} <- {:unit, Units.get_unit(unit_id)},
+         {:unit_owned, true} <- {:unit_owned, unit.user_id == user_id},
+         {:can_rank_up, true} <- {:can_rank_up, can_rank_up(unit)},
+         {:unit_not_in_consumed_units, true} <-
+           {:unit_not_in_consumed_units, unit_id not in consumed_units_ids},
+         consumed_units <- Units.get_units_by_ids(consumed_units_ids),
+         {:consumed_units_owned, true} <-
+           {:consumed_units_owned, Enum.all?(consumed_units, &(&1.user_id == user_id))},
+         {:consumed_units_count, true} <-
+           {:consumed_units_count, Enum.count(consumed_units) == Enum.count(consumed_units_ids)},
```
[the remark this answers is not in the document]
lotuuu: Yeah but that's not the only thing it does (in fact, it's not what I put it in for, I'm just realizing it does that), it also handles the case where we receive an id that doesn't exist, since the query we are doing is just "give me all of the units that have one of these ids". If we receive an ID that doesn't exist, we notify that through `{:error, :consumed_units_not_found}`

apps/champions/lib/champions/units.ex
```diff
@@ -207,6 +217,160 @@ defmodule Champions.Units do
...
+    end
+  end
+
+  defp delete_consumed_units(unit_ids) do
+    {amount_deleted, _return} = Units.delete_units(unit_ids)
+
+    if Enum.count(unit_ids) == amount_deleted, do: {:ok, amount_deleted}, else: {:error, "failed"}
+  end
+
+  defp validate_consumed_units(unit, unit_list) do
```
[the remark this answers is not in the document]
lotuuu: Done!

Nico-Sanchez (approved):

agustinesco (approved):
🚀

PR #333, opened 2024-02-26
Gateway tests

lotuuu:
Implement some basic integration tests. Tests should be expanded later but this serves as a sort of template for them in the future.

# To test

Run `mix test`. No further setup should be needed. Let me know if that does not work on your machine!


apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+
+  describe "users" do
+    test "users", %{socket_tester: socket_tester} do
+      # Create our user
+      :ok = SocketTester.create_user(socket_tester, "Username")
+      fetch_last_message(socket_tester)
+      assert_receive %WebSocketResponse{response_type: {:user, %User{}}}
+
+      fetch_last_message(socket_tester)
+      %WebSocketResponse{response_type: {:user, user}} = get_last_message()
```
agustinesco, line 54: Isn't this the same as above?

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+
+      # Creating another user with the same name fails
+      :ok = SocketTester.create_user(socket_tester, "Username")
+      fetch_last_message(socket_tester)
+      assert_receive %WebSocketResponse{response_type: {:error, %Error{reason: "username_taken"}}}
+
+      # Get user by name
+      :ok = SocketTester.get_user_by_username(socket_tester, "Username")
+      fetch_last_message(socket_tester)
+      assert_receive %WebSocketResponse{response_type: {:user, ^user}}
```
agustinesco, line 64: This would fail if the `Users.get_user_by_username\1` method change their preloads right?

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+    UnitAndCurrencies,
+    User,
+    UserCurrency,
+    WebSocketResponse
+  }
+
+  alias Gateway.SocketTester
+
+  # import Plug.Conn
+  # import Phoenix.ConnTest
```
agustinesco, line 24: Let's delete this

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+      # Unit is selected by default (this will change when we remove sample data in user creation)
+      assert unit_to_unselect.selected
+
+      # Unselect the unit
+      :ok = SocketTester.unselect_unit(socket_tester, user.id, unit_to_unselect.id)
+      fetch_last_message(socket_tester)
+      assert_receive %WebSocketResponse{response_type: {:unit, %Unit{}}}
+
+      fetch_last_message(socket_tester)
+      %WebSocketResponse{response_type: {:unit, unselected_unit}} = get_last_message()
```
agustinesco, line 84: Maybe we should also check that the `unselected_unit` id is the same as the `unit_to_unselect` to ensure the unit is the same

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+      [%CurrencyCost{currency_id: _gold_id, amount: level_up_cost}] = Units.calculate_level_up_cost(unit)
+
+      :ok = SocketTester.level_up_unit(socket_tester, user.id, unit.id)
+      fetch_last_message(socket_tester)
+
+      assert_receive %WebSocketResponse{
+        response_type: {:unit_and_currencies, %UnitAndCurrencies{}}
+      }
+
+      fetch_last_message(socket_tester)
```
agustinesco, line 129: I see that the call of `fetch_last_message` twice is very recurrent, is there a reason?

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+      assert_receive %WebSocketResponse{response_type: {:error, %Error{reason: "cant_level_up"}}}
+
+      #### Tier up
+
+      user_gold = Currencies.get_amount_of_currency_by_name(user.id, "Gold")
+      user_gems = Currencies.get_amount_of_currency_by_name(user.id, "Gems")
+
+      [%CurrencyCost{currency_id: _gold_id, amount: gold_cost}, %CurrencyCost{currency_id: _gems_id, amount: gems_cost}] =
+        Units.calculate_tier_up_cost(unit)
```
agustinesco, line 152: When we implement the real calculate_tier_up_cost there may be a random error here since maybe the currency costs are shuffle and the gem cost appears first

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+                                           } ->
+          {name, amount}
+        end)
+
+      assert unit.level == 20
+      assert unit.tier == 2
+      assert user_currencies["Gold"] == user_gold - gold_cost
+      assert user_currencies["Gems"] == user_gems - gems_cost
+
+      # TODO: Check that we can now level up
```
agustinesco, line 180: Is this todo for another pr?

apps/gateway/test/champions_test.exs
lotuuu: Do you mean if we preloaded different things in get_user and get_user_by_username? I think that should not happen in the first place

apps/gateway/test/champions_test.exs
lotuuu: Yes, is that a bad thing? When we implement actual tier up costs, we should then implement tests that fit the current implementation at that point, right?

apps/gateway/test/champions_test.exs
lotuuu: Yes, the idea behind it was to separate between the send-receive behaviors and the validation of the data, but it resulted from an earlier integration of the SocketTester. I think there's no problem in doing both at the same time so I'm going to get rid of the extra ones!

apps/gateway/test/champions_test.exs
lotuuu: Done!

apps/gateway/test/champions_test.exs
lotuuu: done

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,286 @@
...
+             user_id: user.id,
+             level: 100,
+             tier: 5,
+             rank: Units.get_rank(:star5),
+             selected: false,
+             character_id: same_faction.id
+           })
+           |> elem(1)
+         ])
+      |> Enum.map(& &1.id)
```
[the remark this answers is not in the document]
lotuuu: I agree the function is not clear. A map is what we're doing, actually, but we also need some units to be different so out of the 5 that we're creating, only 3 of them are in the map. In your example you're missing 2 units.

Nevertheless, I will refactor this to make it clearer.

apps/gateway/test/champions_test.exs
lotuuu: I think I'd left these two as TO-DO so as to not make the PR too long. What do you think, should I implement it here?

apps/gateway/test/champions_test.exs
agustinesco: Oh sorry, i didn't noticed that the register function uses `get_user` to retrieve the user after you insert it

apps/gateway/test/champions_test.exs
agustinesco: hmmmm since it's not that big of a chance i would add it here, if we would tests a new functionality yeah we should add it in another pr so this one doesn't scale, but this is just repeating the previous code right?

apps/gateway/test/champions_test.exs
```diff
@@ -0,0 +1,251 @@
...
+        })
+
+      # We will need three 5* Muflus and two i2 of the same faction
+      # For the same faction, we will do one of each unit.
+      units_to_consume = create_units_to_consume(user, muflus, same_faction_character)
+
+      # Characters need a certain quality to rank up
+      assert unit.character.quality == Units.get_quality(:epic)
+
+      # TODO: Check that we cant tier up again without ranking up
```
agustinesco, line 199: I think we should add this on this pr since it's just a validation

lotuuu: Sure

lotuuu: Done!

agustinesco (approved):
LGTM! 🪨  great testing

ncontinanza (approved):
LGTM!

PR #335, opened 2024-02-27
Unit summon

lotuuu:
Implements unit summoning functionality. Different boxes can be created. A box has a specific cost it takes to cast its summon, as well as the different drop rates for each `rank` the unit can be dropped at. Different characters can be dropped at different ranks commons can either drop at 1, 2 or 3 stars (only one of these for each character!), rares can be dropped at 4 star and 5 star, and epics can be dropped at 5 stars. The current implementation could be improved for this game design, let me know if you have any ideas or if you think this does the trick for now!

# To Test

1. Create a user with `{:ok, user} = Champions.Users.register("User")`
2. Choose a box from the list returned by `Champions.Gacha.get_boxes()`
3. Run `Champions.Gacha.summon(user.id, box_id)`
4. Run `Champions.Users.get_user(user.id)` and make sure the right amount of scrolls were discounted and a new unit is added to its units list

If you need. to add more scrolls for summoning you can run: 
`GameBackend.Users.Currencies.add_currency_by_name!(user.id, "Summon Scrolls", 100)`


ncontinanza (approved):
LGTM!

agustinesco (approved):
LGTM! tested and the summon works as expected and the scrolls are being discounted correctly

PR #336, opened 2024-02-27
Config UI

AminArria:
To test this, locally spin up and go to http://localhost:4000/configuration

- [x] Secure endpoint to show and update config
- [x] (simple) validation of config (it parses as json as can be loaded back)
- [x] Update `config.json`
- [ ] Create and update "instance" config
  - This config is generated from base config (`config.json`) if not present
  - Is the one used, `config.json` is not used any more except for first time or for resetting
- [ ] Maintain "instance" config between deployments

Some nice to have, but not deal breaker if in follow ups
- [ ] Basic auth username+password in `runtime.exs` through env variables
- [ ] Full validation of config (what does this even mean??)
- [ ] Nice UI


AminArria:
Replaced by #347

PR #337, opened 2024-02-27
Add level progression

lotuuu:
Allows tracking user progress in campaigns. Users progress through campaigns when they beat levels.

Campaigns have been wrapped in enclosing containers we called "SuperCampaigns". Reason is we will want to have many modes (think, one story where you are the good guys and one where you are the bad guys), and each one of them will have its **sequential** campaigns with their levels. Each of those stories would be a SuperCampaign. This allows for cross-campaign progression `(1-19 -> 1-20 -> 2-1)`.

# Known issues
If we send a call to play the first level of the campaign while we are at a further level, and beat it, we'll advance the level by beating an easy opponent. [Issue has been created.](https://github.com/lambdaclass/champions_of_mirra/issues/214)

CampaignProgress should really be called SuperCampaignProgresss since we have 1 entry for each. [Issue has been created.](https://github.com/lambdaclass/champions_of_mirra/issues/193)

# To test

1. Create User
`{:ok, user} = Champions.Users.register("User")`
2. Add units
`team1 = Enum.map(1..6, fn slot -> GameBackend.Units.insert_unit(%{character_id: GameBackend.Units.Characters.get_character_by_name("Muflus").id, level: 99999, tier: 1, rank: 1, selected: true, user_id: user.id, slot: slot}) |> elem(1) |> GameBackend.Repo.preload(character: [:basic_skill, :ultimate_skill]) end)`
3. Get campaigns
`[campaign_1, campaign_2 | _] = Champions.Campaigns.get_campaigns`
4. Get levels
`[level1, level2, level3, level4, level5, level6, level7, level8, level9, level10] = campaign_1.levels`
5. For each level, fight it, win it and check that the CampaignProgress for user has changed to the next level.
`Champions.Battle.fight_level(user.id, level1.id)`
`GameBackend.Campaigns.get_campaign_progress user_1.id, campaign.id`
When you beat level 10, the last of the campaign, CampaignProgress should change level _and_ campaign

### DIFF
--- a/apps/champions/lib/champions/battle.ex
+++ b/apps/champions/lib/champions/battle.ex
@@ -21,7 +21,6 @@ defmodule Champions.Battle do
          {:level_valid, true} <- {:level_valid, current_level_id == level_id} do
       if battle(user.units, level.units) == :team_1 do
         case Users.advance_level(user_id, level.campaign_id) do
-          # TODO: Validate level fought is the correct one before advancing - [CHoM-#214]
           {:ok, _changes} -> :win
           _error -> {:error, :failed_to_advance}
         end
--- a/apps/champions/lib/champions/battle.ex
+++ b/apps/champions/lib/champions/battle.ex
@@ -17,6 +17,7 @@ defmodule Champions.Battle do
          {:level, {:ok, level}} <- {:level, Campaigns.get_level(level_id)} do
       if battle(user.units, level.units) == :team_1 do
         case Users.advance_level(user_id, level.campaign_id) do
+          # TODO: Validate level fought is the correct one before advancing - [CHoM-#214]
           {:ok, _changes} -> :win
           _error -> {:error, :failed_to_advance}
         end
--- a/apps/champions/lib/champions/battle.ex
+++ b/apps/champions/lib/champions/battle.ex
@@ -6,6 +6,7 @@ defmodule Champions.Battle do
   """
 
   alias GameBackend.Campaigns
+  alias GameBackend.Campaigns.CampaignProgress
   alias GameBackend.Users
 
   @doc """
@@ -14,7 +15,10 @@ defmodule Champions.Battle do
   """
   def fight_level(user_id, level_id) do
     with {:user, {:ok, user}} <- {:user, Users.get_user(user_id)},
-         {:level, {:ok, level}} <- {:level, Campaigns.get_level(level_id)} do
+         {:level, {:ok, level}} <- {:level, Campaigns.get_level(level_id)},
+         {:campaign_progress, {:ok, %CampaignProgress{level_id: current_level_id}}} <-
+           {:campaign_progress, Campaigns.get_campaign_progress(user_id, level.campaign_id)},
+         {:level_valid, true} <- {:level_valid, current_level_id == level_id} do
       if battle(user.units, level.units) == :team_1 do
         case Users.advance_level(user_id, level.campaign_id) do
           # TODO: Validate level fought is the correct one before advancing - [CHoM-#214]
@@ -27,6 +31,8 @@ defmodule Champions.Battle do
     else
       {:user, {:error, :not_found}} -> {:error, :user_not_found}
       {:level, {:error, :not_found}} -> {:error, :level_not_found}
+      {:campaign_progress, {:error, :not_found}} -> {:error, :campaign_progress_not_found}
+      {:level_valid, false} -> {:error, :level_invalid}
     end
   end
 
--- a/apps/champions/lib/champions/battle.ex
+++ b/apps/champions/lib/champions/battle.ex
@@ -10,27 +10,22 @@ defmodule Champions.Battle do
 
   @doc """
   Plays a level for a user, which means fighting its units with their selected ones.
-  Returns `:win` or `:loss` accordingly.
-
-  No tracking for level progress is done yet.
+  Returns `:win` or `:loss` accordingly, and updates the user's progress if they win..
   """
   def fight_level(user_id, level_id) do
-    user = Users.get_user(user_id)
-    level = Campaigns.get_level(level_id)
-
-    cond do
-      is_nil(user) ->
-        {:error, :user_not_found}
-
-      is_nil(level) ->
-        {:error, :level_not_found}
-
-      true ->
-        if battle(user.units, level.units) == :team_1 do
-          :win
-        else
-          :loss
+    with {:user, {:ok, user}} <- {:user, Users.get_user(user_id)},
+         {:level, {:ok, level}} <- {:level, Campaigns.get_level(level_id)} do
+      if battle(user.units, level.units) == :team_1 do
+        case Users.advance_level(user_id, level.campaign_id) do
+          {:ok, _changes} -> :win
+          _error -> {:error, :failed_to_advance}
         end
+      else
+        :loss
+      end
+    else
+      {:user, {:error, :not_found}} -> {:error, :user_not_found}
+      {:level, {:error, :not_found}} -> {:error, :level_not_found}
     end
   end
 
--- a/apps/champions/lib/champions/battle.ex
+++ b/apps/champions/lib/champions/battle.ex
@@ -37,12 +37,12 @@ defmodule Champions.Battle do
   def battle(team_1, team_2) do
     team_1_agg_level =
       Enum.reduce(team_1, 0, fn unit, acc ->
-        unit.unit_level + item_level_agg(unit.items) + acc
+        unit.level + item_level_agg(unit.items) + acc
       end)
 
     team_2_agg_level =
       Enum.reduce(team_2, 0, fn unit, acc ->
-        unit.unit_level + item_level_agg(unit.items) + acc
+        unit.level + item_level_agg(unit.items) + acc
       end)
 
     total_level = team_1_agg_level + team_2_agg_level
--- a/apps/champions/lib/champions/users.ex
+++ b/apps/champions/lib/champions/users.ex
@@ -93,7 +93,7 @@ defmodule Champions.Users do
             game_id: Utils.game_id(),
             user_id: user.id,
             campaign_id: campaign.id,
-            level_id: campaign.levels |> hd() |> Map.get(:id)
+            level_id: campaign.levels |> Enum.sort_by(& &1.level_number) |> hd() |> Map.get(:id)
           })
     end)
   end
--- a/apps/champions/lib/champions/users.ex
+++ b/apps/champions/lib/champions/users.ex
@@ -23,7 +23,7 @@ defmodule Champions.Users do
         add_sample_units(user)
         add_sample_items(user)
         add_sample_currencies(user)
-        add_campaigns_progression(user)
+        add_campaigns_progress(user)
 
         Users.get_user(user.id)
 
@@ -84,11 +84,11 @@ defmodule Champions.Users do
     Currencies.add_currency(user.id, Currencies.get_currency_by_name!("Gems").id, 500)
   end
 
-  defp add_campaigns_progression(user) do
+  defp add_campaigns_progress(user) do
     campaigns = GameBackend.Campaigns.get_campaigns()
 
     Enum.each(campaigns, fn campaign ->
-      GameBackend.Campaigns.insert_campaign_progression(%{
+      GameBackend.Campaigns.insert_campaign_progress(%{
         game_id: @game_id,
         user_id: user.id,
         campaign_id: campaign.id,
--- a/apps/champions/lib/champions/users.ex
+++ b/apps/champions/lib/champions/users.ex
@@ -87,6 +87,7 @@ defmodule Champions.Users do
     campaigns = GameBackend.Campaigns.get_campaigns()
 
     Enum.each(campaigns, fn campaign ->
+      # Only add campaign progress to the first ones of each SuperCampaign
       if campaign.campaign_number == 1,
         do:
           GameBackend.Campaigns.insert_campaign_progress(%{
--- a/apps/champions/lib/champions/users.ex
+++ b/apps/champions/lib/champions/users.ex
@@ -24,7 +24,7 @@ defmodule Champions.Users do
         add_sample_currencies(user)
         add_campaigns_progress(user)
 
-        {:ok, Users.get_user(user.id)}
+        Users.get_user(user.id)
 
       {:error, changeset} ->
         [[first_error | _other_errors] | _other_fields_errors] =
--- a/apps/champions/lib/champions/users.ex
+++ b/apps/champions/lib/champions/users.ex
@@ -3,13 +3,12 @@ defmodule Champions.Users do
   Users logic for Champions Of Mirra.
   """
 
+  alias Champions.Utils
   alias Ecto.Changeset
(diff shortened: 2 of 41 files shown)


apps/champions/lib/champions/users.ex
```diff
@@ -88,19 +83,37 @@ defmodule Champions.Users do
...
+    campaigns = GameBackend.Campaigns.get_campaigns()
+
+    Enum.each(campaigns, fn campaign ->
+      if campaign.campaign_number == 1,
+        do:
+          GameBackend.Campaigns.insert_campaign_progress(%{
+            game_id: @game_id,
+            user_id: user.id,
+            campaign_id: campaign.id,
+            level_id: campaign.levels |> hd() |> Map.get(:id)
```
agustinesco, line 96: This will be incorrect if the `campaign.levels` are not sorted right?  the sort was removed from `get_campaigns`

apps/game_backend/lib/game_backend/campaigns.ex
```diff
@@ -41,10 +37,86 @@ defmodule GameBackend.Campaigns do
...
+
   @doc """
   Get a level by id.
+
+  Returns `{:error, :not_found}` if no level is found.
   """
   def get_level(level_id) do
-    Repo.get(Level, level_id) |> Repo.preload(units: :items, units: :character)
+    level = Repo.get(Level, level_id) |> Repo.preload(units: :items, units: :character)
+    if level, do: {:ok, level}, else: {:error, :not_found}
```
agustinesco, line 74: why are we always wrapping the results of repo getters inside tuples? isn't it easier to just return the value and handle nil or not nil results?

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -42,29 +45,92 @@ defmodule GameBackend.Users do
...
+          else: update_campaign_progress(campaign_progress, next_campaign_id, next_level_id)
+      end)
+      |> Repo.transaction()
+    else
+      {:campaign_data, _transaction_error} -> {:error, :campaign_data_error}
+      {:next_level, _transaction_error} -> {:campaign_progress_error}
+    end
+  end
+
+  def reset_afk_rewards_claim(user_id) do
```
agustinesco, line 101: This functions is not being used

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -42,29 +45,92 @@ defmodule GameBackend.Users do
...
+      Repo.preload(user,
+        items: :template,
+        units: [:character, :items],
+        currencies: :currency
+      )
+
+  defp retrieve_campaign_progress_data(user_id, campaign_id) do
+    progress =
+      Repo.get_by(CampaignProgress, user_id: user_id, campaign_id: campaign_id)
+      |> Repo.preload(level: :campaign)
```
agustinesco, line 120: you already have a getter for this on the `campaigns.ex` module `get_campaign_progress`, maybe adding these preloads there would do it

agustinesco (changes requested):
Correct me if i'm wrong , the logic would be like this:
- You have a container for campaigns calles `SuperCampaigns`
- Each campaigns have a list of levels for the users to beat
- You're implementing a new table called `campaign_progression` to track on witch level of each campaign the user is currently fight
- Each time a user finish a lever you're updating the `campaign_progression` register from that campaign to the next level unless it's the last
That's correct?
The tables would be something like: 

super campaigns

agustinesco:
I'm getting a compilation warning:
```
    warning: undefined module attribute @game_id, please remove access to @game_id or explicitly set it before access
    \x{2502}
 93 \x{2502}             game_id: @game_id,
    \x{2502}             ~~~~~~~~~~~~~~~~~~
    \x{2502}
    \x{2514}\x{2500} lib/champions/users.ex:93: Champions.Users (module)
```
I already deleted the _build folder and recompiled everything

apps/game_backend/lib/game_backend/campaigns.ex
lotuuu: This allows are functions to use `with` patterns that make code much more readable! See Unit level_up for example

agustinesco:
I'm trying to use the command you give me to fight a level but i'm having troubles:
1. `Champions.Users.register("User")`is returning {:ok, user}, resulting in {:ok, {:ok, user}}
2. For some reason , if you try to preload the units you get an error that `Ecto.Changeset.__Schema__ is undefined` also is this necessary? 
3. no issues
4. no issues
5. I can't win even tho my whole team is level 9999...  `{:error, :failed_to_advance}`

lotuuu:
> Correct me if i'm wrong , the logic would be like this:
> 
> * You have a container for campaigns calles `SuperCampaigns`
> * Each campaigns have a list of levels for the users to beat
> * You're implementing a new table called `campaign_progression` to track on witch level of each campaign the user is currently fight
> * Each time a user finish a lever you're updating the `campaign_progression` register from that campaign to the next level unless it's the last
>   That's correct?
>   The tables would be something like:
> 
> super campaigns

Not quite, campaign progress is linear between SuperCampaigns, meaning a user can only be sitting on 1 level of a supercampaign (you cannot be at level 5 for campaign 1 and also level 8 for campaign 2). This is why we've created a [ticket](https://github.com/lambdaclass/champions_of_mirra/issues/193) to refactor it to SuperCampaignProgress later. I've [added some of this](https://github.com/lambdaclass/mirra_backend/pull/337/commits/1aa2f342d7224d33dd21d14184cb1f077bf7538d) to the CampaignProgress doc to make it clearer, let me know if it's good enough.

lotuuu:
> I'm getting a compilation warning:
> 
> ```
>     warning: undefined module attribute @game_id, please remove access to @game_id or explicitly set it before access
>     \x{2502}
>  93 \x{2502}             game_id: @game_id,
>     \x{2502}             ~~~~~~~~~~~~~~~~~~
>     \x{2502}
>     \x{2514}\x{2500} lib/champions/users.ex:93: Champions.Users (module)
> ```
> 
> I already deleted the _build folder and recompiled everything

Yes I think this is a merge issue from when we introduced Champions.Utils.game_id(). I've removed the unexisting module attribute.

lotuuu:
> I'm trying to use the command you give me to fight a level but i'm having troubles:
> 
> 1. `Champions.Users.register("User")`is returning {:ok, user}, resulting in {:ok, {:ok, user}}
> 2. For some reason , if you try to preload the units you get an error that `Ecto.Changeset.__Schema__ is undefined` also is this necessary?
> 3. no issues
> 4. no issues
> 5. I can't win even tho my whole team is level 9999...  `{:error, :failed_to_advance}`

1. Solved!
2. Documentation issue that arose from merge. It's not really about the preloads but rather the fact that unit_level was rename to level in our code but not in the PR documentation. I've fixed it now! Also yeah the preload is probably unneccesary because fight_level is not using our new battle system yet.
5. Probably due to the fact that you were not creating your units well (you weren't passing in the required "level" field, as stated in point 2)

apps/game_backend/lib/game_backend/users.ex
lotuuu: Woops, this one belongs to another PR hehe

agustinesco:
> > Correct me if i'm wrong , the logic would be like this:
> > 
> > * You have a container for campaigns calles `SuperCampaigns`
> > * Each campaigns have a list of levels for the users to beat
> > * You're implementing a new table called `campaign_progression` to track on witch level of each campaign the user is currently fight
> > * Each time a user finish a lever you're updating the `campaign_progression` register from that campaign to the next level unless it's the last
> >   That's correct?
> >   The tables would be something like:
> > 
> > super campaigns
> 
> Not quite, campaign progress is linear between SuperCampaigns, meaning a user can only be sitting on 1 level of a supercampaign (you cannot be at level 5 for campaign 1 and also level 8 for campaign 2). This is why we've created a [ticket](https://github.com/lambdaclass/champions_of_mirra/issues/193) to refactor it to SuperCampaignProgress later. I've [added some of this](https://github.com/lambdaclass/mirra_backend/pull/337/commits/1aa2f342d7224d33dd21d14184cb1f077bf7538d) to the CampaignProgress doc to make it clearer, let me know if it's good enough.

nice! thanks for explaining

apps/champions/lib/champions/users.ex
lotuuu: Yes indeed! Fixed by sorting the list.

agustinesco (dismissed):
LGTM!

apps/champions/lib/champions/battle.ex
```diff
@@ -10,27 +10,23 @@ defmodule Champions.Battle do
...
-          :win
-        else
-          :loss
+    with {:user, {:ok, user}} <- {:user, Users.get_user(user_id)},
+         {:level, {:ok, level}} <- {:level, Campaigns.get_level(level_id)} do
+      if battle(user.units, level.units) == :team_1 do
+        case Users.advance_level(user_id, level.campaign_id) do
+          # TODO: Validate level fought is the correct one before advancing - [CHoM-#214]
+          {:ok, _changes} -> :win
+          _error -> {:error, :failed_to_advance}
```
Nico-Sanchez, line 22: This code is a bit messy. It's hard to notice the returning value.
We have a with, an if and a case in it. I have a few suggestions to improve its readability:
- Avoid with statement for getters, we can just pattern-match with :ok. If we want to change the msg error, we could do that in the getter function itself instead of doing it here.
- Replace the if statement with an auxiliary function. You can pattern-match the :team_1 atom to handle each return.
- For the case statement, we can do what I mentioned above for the others (pattern-match, change function returning value).

apps/champions/lib/champions/users.ex
```diff
@@ -88,19 +83,37 @@ defmodule Champions.Users do
     Currencies.add_currency(user.id, Currencies.get_currency_by_name!("Gems").id, 500)
   end
 
+  defp add_campaigns_progress(user) do
+    campaigns = GameBackend.Campaigns.get_campaigns()
+
+    Enum.each(campaigns, fn campaign ->
+      if campaign.campaign_number == 1,
```
Nico-Sanchez, line 90: What is this `campaign_number == 1` for?

apps/champions/lib/champions/users.ex
```diff
@@ -88,19 +83,37 @@ defmodule Champions.Users do
...
   @doc """
   Adds the given experience to a user. If the user were to have enough resulting experience to level up,
   it is performed automatically.
   """
   def add_experience(user_id, experience) do
-    user = get_user(user_id)
-    new_experience = user.experience + experience
+    case get_user(user_id) do
+      {:ok, user} ->
+        new_experience = user.experience + experience
```
Nico-Sanchez, line 108: We can replace this case pattern-matching the get_user/1 returning value:
```suggestion
    {:ok, user} = get_user(user_id)
    new_experience = user.experience + experience
```

apps/game_backend/lib/game_backend/campaigns.ex
Nico-Sanchez: You can still use with statements with `{:ok, entity}` and `nil` returning values.
Also, sometimes the `with` statement is really hard to follow, specially when it comes with an else clause. Use it carefully when it improves the code readability.

Quoting [elixirschool.com](https://elixirschool.com/en/lessons/basics/control_structures):
> The special form with/1 is useful when you might use a nested case/2 statement or situations that cannot cleanly be piped together.

apps/game_backend/lib/game_backend/campaigns.ex
```diff
@@ -41,10 +37,88 @@ defmodule GameBackend.Campaigns do
...
+          campaign_number: campaign.campaign_number + 1
+        )
+
+      if next_campaign do
+        first_level = Repo.get_by(Level, campaign_id: next_campaign.id, level_number: 1)
+        {next_campaign.id, first_level.id}
+      else
+        {campaign.id, level.id}
+      end
+    end
```
Nico-Sanchez, line 122: We could avoid these nested if's using a case with 3 statements:
1. has_next_level? ->
2. has_next_campaign? ->
3. true ->

Nico-Sanchez (changes requested):
Nice work @lotuuu 

A few comments regarding the code, but the feature works as expected.
Also, the PR says that:

>  If we send a call to play the first level of the campaign while we are at a further level, and beat it, we'll advance the level by beating an easy opponent. https://github.com/lambdaclass/champions_of_mirra/issues/214

I see that the Issue is marked as done but it's not fixed. We could merge with main and test if its really fixed or not. I did it and it's not solved but maybe I did something wrong.

apps/champions/lib/champions/battle.ex
lotuuu: I'm not sure I understand your suggestions here. For #1, we need this with to avoid having multiple nested ifs/cases. For #3 we're using pattern matching for the return of `Users.advance_level` like you say.

In #2 we could use a case instead of an if in order to pattern match it, but would that really make the code clearer? I think the "better" way for when you have two outcomes is an if/else rather than a case with two matches.

apps/champions/lib/champions/users.ex
lotuuu: It's the first campaign of the SuperCampaign!

apps/champions/lib/champions/users.ex
lotuuu: That would make our application raise if the user cannot be found. Isn't it better to handle these expected errors gracefully?

apps/champions/lib/champions/battle.ex
lotuuu: This is what I could come up with to make the returned values clearer:
```
def fight_level(user_id, level_id) do
  with {:user, {:ok, user}} <- {:user, Users.get_user(user_id)},
       {:level, {:ok, level}} <- {:level, Campaigns.get_level(level_id)},
       {:campaign_progress, {:ok, %CampaignProgress{level_id: current_level_id}}} <-
         {:campaign_progress, Campaigns.get_campaign_progress(user_id, level.campaign_id)},
       {:level_valid, true} <- {:level_valid, current_level_id == level_id} do
    cond do
      battle(user.units, level.units) != :team_1 -> :loss
      advance_level(user_id, level.campaig_id) -> :win
      true -> {:error, :failed_to_advance}
    end
  else
    {:user, {:error, :not_found}} -> {:error, :user_not_found}
    {:level, {:error, :not_found}} -> {:error, :level_not_found}
    {:campaign_progress, {:error, :not_found}} -> {:error, :campaign_progress_not_found}
    {:level_valid, false} -> {:error, :level_invalid}
  end
end

defp advance_level(user_id, campaign_id) do
  case Users.advance_level(user_id, campaign_id) do
    {:ok, _changes} -> true
    _error -> false
  end
end
```
I'm not a fan of this approach. It "hides" our calls in the conditions of the `cond` and also adds an auxiliary function that doesn't really do anything. In any case, if you think this way is clearer we can go for it.

apps/champions/lib/champions/users.ex
Nico-Sanchez: Yes, you're right.

Nico-Sanchez (approved):

agustinesco (approved):
🪨

Issue #334, opened 2024-02-27
[REFACTOR]  Analyze and refactor code of longest modules


Nico-Sanchez:
Using cloc (and ignoring protobuf messages) these are the longest ones:

- game_updater.ex
- skill.ex
- player.ex

<img width="874" alt="Screenshot 2024-02-27 at 13 24 03" src="https://github.com/lambdaclass/mirra_backend/assets/65268603/4ae79f08-aaae-4847-b8f9-2a256d9e7b47">

PR #338, opened 2024-02-28
[GH-334] Refactor every game update (tick) action 

Nico-Sanchez:
1st iteration of #334 

Game update handler was messy. This PR wants to expose each action to see refactor oportunities with more clarity.

- Moves out every tick action to a separate function.
- Each of the mentioned functions have to receive and return the game state map.

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -623,7 +623,8 @@ defmodule Arena.GameUpdater do
           nil ->
             {players_acc, power_ups_acc}
 
-          power_up ->
+          power_up_id ->
+            power_up = Map.get(power_ups_acc, power_up_id)
             process_power_up(player, power_up, players_acc, power_ups_acc)
         end
       end)
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -566,7 +566,7 @@ defmodule Arena.GameUpdater do
     # We don't want to move recently exploded projectiles
     {recently_exploded_projectiles, alive_projectiles} =
       projectiles
-      |> Enum.split_with(fn {_projectile_id, projectile} ->
+      |> Map.split_with(fn {_projectile_id, projectile} ->
         projectile.aditional_info.status == :EXPLODED
       end)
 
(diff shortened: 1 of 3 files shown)


apps/arena/lib/arena/game_updater.ex
```diff
@@ -557,12 +505,208 @@ defmodule Arena.GameUpdater do
...
+
+  # Remove exploded projectiles
+  # Update expired to explode status
+  defp update_projectiles_status(%{projectiles: projectiles} = game_state) do
+    updated_projectiles =
+      projectiles
+      |> Map.reject(fn {_projectile_id, projectile} ->
+        projectile.aditional_info.status == :EXPLODED
+      end)
+      |> Enum.reduce(%{}, fn {projectile_id, projectile}, acc ->
```
AminArria, line 545: You are iterating twice over `projectiles`, should we do this both things in `Enum.reduce/3`?

Sort of like
```elixir
Enum.reduce(projectiles, projectiles, fn {projectile_id, projectile}, acc ->
  case projectile.aditional_info.status do
    :EXPLODED -> Map.delete(acc, projectile_id)
    :EXPIRED -> put_in(acc, [projectile_id, :aditional_info, :status], :EXPLODED))
    _ -> acc
  end
end)

File: apps/arena/lib/arena/game_updater.ex [hunk 2]
```diff
@@ -557,12 +505,208 @@ defmodule Arena.GameUpdater do
...
+           players: players,
+           obstacles: obstacles,
+           external_wall: external_wall,
+           ticks_to_move: ticks_to_move
+         } = game_state
+       ) do
+    # We don't want to move recently exploded projectiles
+    {recently_exploded_projectiles, alive_projectiles} =
+      projectiles
+      |> Enum.split_with(fn {_projectile_id, projectile} ->
```
AminArria, line 576: You can do `Map.split_with/2` with almost the same result, except the tuple will contain maps so you don't need the `Map.new/1` later on

AminArria (changes requested):

apps/arena/lib/arena/game_updater.ex
Nico-Sanchez: Nice catch! I always forget that Map has some enum functions 😅

apps/arena/lib/arena/game_updater.ex
Nico-Sanchez: Yeah, looks much better. Done!

agustinesco (dismissed):
LGTM! i tested this pr with multiple players, projectiles and mechanics and everythoing worked fine! maybe we could deploy this and test it with multiple players online!

manucamejo (approved):

manucamejo:
@Nico-Sanchez I think that you can take outside the assigns to the new game from lines `500`, `501` and `502` in `game_updater.ex`

Nico-Sanchez:
> @Nico-Sanchez I think that you can take outside the assigns to the new game from lines `500`, `501` and `502` in `game_updater.ex`

Nice catch! Done!

manucamejo (approved):

AminArria (approved):

PR #339, opened 2024-02-28
Campaign rewards

lotuuu:
Waiting on #337

Implements level rewards. Rewards can come in the form of currency, units (always level 1 and tier 1) and items (always level 1). Rewards are given automatically when a level is beaten.

Unit ranks will be customizable per reward but we don't have them yet (see: https://github.com/lambdaclass/mirra_backend/pull/329).

# To test

- Create user, get campaigns, get levels

```
{:ok, user} = Champions.Users.register("User")
{:ok, [campaign | _]} = Champions.Campaigns.get_campaigns
[level1, level2, level3 | _] = campaign.levels |> Enum.sort_by(& &1.level_number)

```

- Check out your gold

```
GameBackend.Users.Currencies.get_amount_of_currency(user.id, GameBackend.Users.Currencies.get_currency_by_name!("Gold").id)
```

- Beat level 1 (gives only currency)

```
Champions.Battle.fight_level(user.id, level1.id)
```

- Check out your gold again (should have grown)

```
GameBackend.Users.Currencies.get_amount_of_currency(user.id, GameBackend.Users.Currencies.get_currency_by_name!("Gold").id)
```

- Check your items

```
GameBackend.Users.get_user(user.id) |> elem(1) |> Map.get(:items) |> Enum.count()
```

- Fight level 2 (gives an item)

```
Champions.Battle.fight_level(user.id, level2.id)
```

- Check your items again. Should have grown.

```
GameBackend.Users.get_user(user.id) |> elem(1) |> Map.get(:items) |> Enum.count()
```

- Same for units and level 3
```
Champions.Battle.fight_level(user.id, level3.id)
```

```
GameBackend.Users.get_user(user.id) |> elem(1) |> Map.get(:units) |> Enum.count()
```


apps/game_backend/lib/game_backend/campaigns.ex
```diff
@@ -13,9 +13,9 @@ defmodule GameBackend.Campaigns do
   def get_campaigns() do
     campaigns =
       Repo.all(from(c in Campaign))
-      |> Repo.preload(levels: [units: :items])
+      |> Repo.preload(levels: [:currency_rewards, units: [:items, :character]])
 
-    if Enum.empty?(campaigns), do: {:error, :no_campaigns}, else: campaigns
+    if Enum.empty?(campaigns), do: {:error, :no_campaigns}, else: Enum.sort_by(campaigns, & &1.campaign_number)
```
agustinesco, line 18: why not creating a single query that does all instead of pipe the results?
```
q = 
   from c in Campaign,
   preload: [levels: [:currency_rewards, units: [:items, :character]]],
   order_by: [asc: c.campaign_number]
```
do you think the pipes are more readable?

apps/game_backend/lib/game_backend/campaigns.ex
```diff
@@ -24,7 +24,7 @@ defmodule GameBackend.Campaigns do
   def get_campaign(campaign_id) do
     case Repo.get(Campaign, campaign_id) |> Repo.preload(levels: [:units]) do
       nil -> {:error, :not_found}
-      campaign -> {:ok, campaign}
+      campaign -> {:ok, Map.put(campaign, :levels, Enum.sort_by(campaign.levels, & &1.level_number))}
```
agustinesco, line 27: same as above, in this case it may be different because you are sorting a list inside the registers, it may be solved with a subquery but they are a bit hard to follow

apps/game_backend/lib/game_backend/campaigns.ex
lotuuu: No, not at all :P your solution is much better.

apps/game_backend/lib/game_backend/campaigns.ex
lotuuu: I don't think it'll be hard to see what the subquery does. The correct thing would be to handle it in the query just like the previous comment.

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -113,4 +130,56 @@ defmodule GameBackend.Users do
...
+          inserted_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second),
+          updated_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
+        }
+      end)
+    end)
+    |> List.flatten()
+  end
+
+  defp build_unit_rewards_params(user_id, unit_rewards) do
+    IO.inspect(unit_rewards, label: :rewards)
```
agustinesco, line 157: lets delete this

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -113,4 +130,56 @@ defmodule GameBackend.Users do
...
+
+  defp build_item_rewards_params(user_id, item_rewards) do
+    Enum.map(item_rewards, fn item_reward ->
+      Enum.map(1..item_reward.amount, fn _ ->
+        %{
+          user_id: user_id,
+          template_id: item_reward.item_template_id,
+          level: 1,
+          inserted_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second),
+          updated_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
```
agustinesco, line 149: I believe that Ecto add these fields when the table has the timestamps() method added to the migration, i think they're missing because you're not running this map through the item chageset?

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -113,4 +130,56 @@ defmodule GameBackend.Users do
...
+    Enum.map(unit_rewards, fn unit_reward ->
+      Enum.map(1..unit_reward.amount, fn _ ->
+        %{
+          user_id: user_id,
+          character_id: unit_reward.character_id,
+          unit_level: 1,
+          tier: 1,
+          selected: false,
+          inserted_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second),
+          updated_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
```
agustinesco, line 168: same as above

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -113,4 +130,56 @@ defmodule GameBackend.Users do
...
+  defp check_result(result, element_name) do
+    if Enum.all?(result, fn
+         {:ok, _} -> true
+         _ -> false
+       end) do
+      {:ok, result}
+    else
+      {:error, "Failed to apply " <> Atom.to_string(element_name)}
+    end
+  end
```
agustinesco, line 184: a few things about this:
1. You're just using this function for `apply_currency_rewards` and imo the can be in the same function
2. the enum all clause is a bit weird, why not just comparing the first element of the tuple? 
```
if Enum.all?(result, fn {result, _} -> result == :ok end)
```
3. this may be a personal opinion but why not adding the currency updating to the main multi? that way you don't need this validation
```
      Multi.new()
      |> apply_currency_rewards(user_id, level.currency_rewards)



defp apply_currency_rewards(multi, user_id, currency_rewards) do
  Enum.reduce(currency_rewards, multi, fn currency_reward, multi -> 
    Multi.run({:add_currency, currency_reward}, Currencies.add_currency(user_id, currency_reward.currency_id, currency_reward.amount))
  end)
end
```

agustinesco:
The function `get_campaign_progress` has missing some level preloads causing the rewards to fail `preload: [level: [:campaign, :currency_rewards, :item_rewards, :unit_rewards]]` this are the preloads necessary

agustinesco:
The unit rewards are not working as expected, i got the next error when fighting the third level:
```
 unknown field `:unit_level` in schema GameBackend.Units.Unit given to insert_all. Note virtual fields and associations are not supported
```

apps/game_backend/lib/game_backend/users.ex
lotuuu: Exactly, but we're not running it to the changeset because we're using insert_all, and it doesn't support automatic timestamps: https://github.com/elixir-ecto/ecto/issues/1932

apps/game_backend/lib/game_backend/users.ex
lotuuu: Whoops, good thing we have credo ci now :P

agustinesco (approved):
LGTM!

Nico-Sanchez (approved):

PR #344, opened 2024-03-01
Fix get_characters functions

lotuuu:
The preload was left based on an outdated implementation of skills.

# To test

Make sure you have some characters in your db (seeds should be creating some) and then run these functions to get them.


Nico-Sanchez (approved):

agustinesco (approved):
LGTM! 🪨

PR #347, opened 2024-03-05
Configurator app

AminArria:
This PR is probably best reviewed by ignoring the initial 2 generator ones, you can [click here for that diff](https://github.com/lambdaclass/mirra_backend/pull/347/files/9fbdcc8bd30df310cc704be99106d78d3ab28477..a2fa2fa9c1d421b3db0c9102e0c8081098733a97)

Functionality wise this PR should make no noticeable change, so for testing purposes if you can play similar to `main` then all good

### What is Configurator?

`Configurator` is the app for configuration (duh...), it will be in charge of storing the configs needed by the games and making it available for editing and getting

**IMPORTANT**: This PR moves `config.json` from `arena` to `configurator`

Current considerations
- The configuration is stored as a text field. This is because using `map` and a `jsonb` column had an issue with the floats, for example `4000.0` was transformed to `4000` when saving and seems that is "expected" behaviour, but I would need to dive deeper into ecto+jason+postgrex to figure out the problem
- No `game_id` column, I'll rather handle that as a next step. This PR is big enough and I want to lookup/understand some things before doing it
- Uses its own `Repo` module. This will be kinda against what we are doing in Champions (not using `game_backend`), but I'm not entirely sure we are doing a correct thing with that and I rather have all in this app and extract it later
- All configurations are umodifiable, instead you can use one (or default) as base an edit specific things you want
- A nice JSON editor is left for a future iteration (as is also a better UI in general)
- You can only set one config as default

Finally the last change was to make Arena fetch the config from Configurator


apps/configurator/mix.exs
```diff
@@ -0,0 +1,89 @@
...
+
+  def project do
+    [
+      app: :configurator,
+      version: "0.1.0",
+      build_path: "../../_build",
+      config_path: "../../config/config.exs",
+      deps_path: "../../deps",
+      lockfile: "../../mix.lock",
+      elixir: "~> 1.14",
```
Nico-Sanchez, line 12: Shouldn't be the same elixir version as the other apps?

apps/configurator/priv/repo/migrations/.formatter.exs
```diff
@@ -0,0 +1,4 @@
+[
+  import_deps: [:ecto_sql],
+  inputs: ["*.exs"]
```
Nico-Sanchez, line 3: Do we need a formatter here? If so, let's add the :line_length parameter

apps/configurator/priv/repo/seeds.exs
```diff
@@ -0,0 +1,5 @@
+## TODO: Change this once PR#336 is merged
```
Nico-Sanchez, line 1: What's this comment for? #336 is closed 🤔

apps/configurator/priv/static/robots.txt
```diff
@@ -0,0 +1,5 @@
+# See https://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+#
+# To ban all spiders from the entire site uncomment the next two lines:
+# User-agent: *
```
Nico-Sanchez, line 4: ??????????????????????

apps/configurator/test/configurator_web/controllers/page_controller_test.exs
```diff
@@ -0,0 +1,8 @@
+defmodule ConfiguratorWeb.PageControllerTest do
+  use ConfiguratorWeb.ConnCase
+
+  test "GET /", %{conn: conn} do
+    conn = get(conn, ~p"/")
+    assert html_response(conn, 200) =~ "Peace of mind from prototype to production"
+  end
+end
```
Nico-Sanchez, line 8: We can delete this.

config/config.exs
```diff
@@ -150,6 +171,33 @@ config :gateway, Gateway.Endpoint,
...
+  live_view: [signing_salt: "6A8twvHJ"]
+
+# Configures the mailer
+#
+# By default it uses the "Local" adapter which stores the emails
+# locally. You can see the emails in your browser, at "/dev/mailbox".
+#
+# For production it's recommended to configure a different adapter
+# at the `config/runtime.exs`.
+config :configurator, Configurator.Mailer, adapter: Swoosh.Adapters.Local
```
Nico-Sanchez, line 199: I think we don't need the mailer.

config/runtime.exs
```diff
@@ -220,3 +220,107 @@ end
...
+  # variable instead.
+  secret_key_base =
+    System.get_env("SECRET_KEY_BASE") ||
+      raise """
+      environment variable SECRET_KEY_BASE is missing.
+      You can generate one by calling: mix phx.gen.secret
+      """
+
+  host = System.get_env("PHX_HOST") || "example.com"
+  port = String.to_integer(System.get_env("PORT") || "4000")
```
Nico-Sanchez, line 261: Isn't this 4100?
```suggestion
  port = String.to_integer(System.get_env("PORT") || "4100")
```

apps/configurator/mix.exs
```diff
@@ -0,0 +1,89 @@
...
+       depth: 1},
+      {:swoosh, "~> 1.5"},
+      {:finch, "~> 0.13"},
+      {:telemetry_metrics, "~> 0.6"},
+      {:telemetry_poller, "~> 1.0"},
+      {:gettext, "~> 0.20"},
+      {:jason, "~> 1.2"},
+      {:dns_cluster, "~> 0.1.1"},
+      {:bandit, "~> 1.2"}
+    ]
```
Nico-Sanchez, line 65: do we need all these deps? Maybe we could delete some.

apps/configurator/lib/configurator_web/router.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  pipeline :api do
+    plug :accepts, ["json"]
+  end
+
+  scope "/", ConfiguratorWeb do
+    pipe_through :browser
+
+    get "/", PageController, :home
+    resources "/configurations", ConfigurationController, only: [:index, :new, :create, :show]
```
Nico-Sanchez, line 21: This could be the home page.

Nico-Sanchez (changes requested):
Left lot of comments about boilerplate code and modules. Let me know if you agree or want to do it in further iterations, but I personally think we should cut all that code out.

Also, when I try to create a new config it crashes:

image

apps/arena/lib/arena/game_launcher.ex
```diff
@@ -4,7 +4,7 @@ defmodule Arena.GameLauncher do
   use GenServer
 
   # Amount of clients needed to start a game
-  @clients_needed 10
+  @clients_needed 1
```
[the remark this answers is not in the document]
AminArria: Fuuu I keep committing this by accident

apps/arena/rel/overlays/bin/server.bat
```diff
@@ -0,0 +1,2 @@
+set PHX_SERVER=true
+call "%~dp0\arena" start
```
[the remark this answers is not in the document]
AminArria: Pretty sure this `rel` folder was commited by accident

apps/configurator/priv/static/robots.txt
AminArria: Jajajaja but what is your question?

apps/configurator/mix.exs
AminArria: Deleted some and left others cause they might not be used right now, but if we add them in the next PR it will be a major pain to unremove

apps/configurator/lib/configurator_web/router.ex
AminArria: Changed to do that

apps/configurator/priv/repo/seeds.exs
AminArria: Removed seeds (content), it will be dynamically done at runtime cause it makes sense and its easier

apps/configurator/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Ready to run in production? Please [check our deployment guides](https://hexdocs.pm/phoenix/deployment.html).
+
+## Learn more
+
+  * Official website: https://www.phoenixframework.org/
+  * Guides: https://hexdocs.pm/phoenix/overview.html
+  * Docs: https://hexdocs.pm/phoenix
+  * Forum: https://elixirforum.com/c/phoenix-forum
+  * Source: https://github.com/phoenixframework/phoenix
```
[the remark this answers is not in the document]
AminArria: Updated READMEs with information https://github.com/lambdaclass/mirra_backend/pull/347/commits/3b5eb323289e6053aef403feb5ffead99622071b

README.md
```diff
@@ -62,6 +62,15 @@ Note you need to have ran `devenv up` earlier.
...
 Each of the applications, as mentioned in the previous paragraph, has a specific responsibility:
 
+## Applications
+
+- [Arena](#arena)
+- [Game Client](#game-client)
+- [Gateway](#gateway)
+- [ChampionsOfMirra](#championsofmirra)
+- [GameBackend](#gamebackend)
+- [Configurator]()
```
Nico-Sanchez, line 72:
```suggestion
- [Configurator](#configurator)
```

Nico-Sanchez (changes requested):
I have a few bugs using the UI:

- Modifying a config creates a new one instead of updating it.
- Delete config endpoint is broken.

Also a question: a default config is the one being used? I think a name like enabled or current is more accurate than default.

AminArria:
> Modifying a config creates a new one instead of updating it.

Yep, configs are inmutable. "Modifying" one actually creates another one so we can have traceability, rollbacks, and A/B testing without major works

> Delete config endpoint is broken.

Removed the link, should not have been there

> Also a question: a default config is the one being used? I think a name like enabled or current is more accurate than default.

We can take that conversation in Slack, seems like a really bikesheddy discussion to have and is a small change in a subsequent PR

Nico-Sanchez:
> > Modifying a config creates a new one instead of updating it.
> 
> Yep, configs are inmutable. "Modifying" one actually creates another one so we can have traceability, rollbacks, and A/B testing without major work

Sounds good. However, we could call it "copy" so the user knows that it's creating a copy instead of modifying it in place.

>
> > Also a question: a default config is the one being used? I think a name like enabled or current is more accurate than default.
> 
> We can take that conversation in Slack, seems like a really bikesheddy discussion to have and is a small change in a subsequent PR

Doing it in a subsequent PR sounds good to me.

AminArria:
> Sounds good. However, we could call it "copy" so the user knows that it's creating a copy instead of modifying it in place.

Good idea, made the change https://github.com/lambdaclass/mirra_backend/pull/347/commits/58fc16d2c4a42f15155d2064b82e5c0dd33300c5

Nico-Sanchez (approved):

manucamejo:
@AminArria I merged `main` and I'm getting an error when the game starts because the `config.json` doesn't exist anymore and we are doing this `Application.app_dir(:arena, "priv/config.json")` (apps/arena/lib/arena/configuration.ex:12), maybe it's a merge error.
Also, the current config is longer, is there a problem with the format? Or can we replace it with the new one and that's it?

AminArria:
Fixed now

apps/configurator/README.md
```diff
@@ -0,0 +1,43 @@
+# Configurator
+
+To start your Phoenix server:
+
+  * Run `mix setup` to install and setup dependencies
+  * Start Phoenix endpoint with `mix phx.server` or inside IEx with `iex -S mix phx.server`
+
+Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
```
agustinesco, line 8: These ports are wrong right?

agustinesco (changes requested):
Don't we also need to chenge the `get_game_config()` method in the `Arena.Configuration` module to use the default config? i think the changes we're macking in the configurator app are not affecting the games, or it's outside of this pr's scope?

AminArria:
> Don't we also need to chenge the `get_game_config()` method in the `Arena.Configuration` module to use the default config? i think the changes we're making in the configurator app are not affecting the games, or it's outside of this pr's scope?

@agustinesco correct, the changes in this PR should not change anything. This is because there is a big risk of breaking things and also people would need to adjust rather quickly to this so I'll introduce it in stages

agustinesco (approved):
LGTM!

PR #349, opened 2024-03-06
Adjust Antimatter config values.

BertovDev:
In order for the animations and vfx to work well with the client we should use this config for now.
      
      "execution_duration_ms": 300,
      "activation_delay_ms": 0
 to ->
      "execution_duration_ms": 800,
      "activation_delay_ms": 400,

In the client -> https://github.com/lambdaclass/curse_of_mirra/pull/1512

### DIFF
--- a/apps/arena/priv/config.json
+++ b/apps/arena/priv/config.json
@@ -277,7 +277,7 @@
       "name": "antimatter",
       "cooldown_ms": 2000,
       "execution_duration_ms": 800,
-      "activation_delay_ms": 400,
+      "activation_delay_ms": 500,
       "is_passive": false,
       "autoaim": true,
       "stamina_cost": 1,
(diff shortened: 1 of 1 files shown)


apps/arena/priv/config.json
```diff
@@ -276,8 +276,8 @@
     {
       "name": "antimatter",
       "cooldown_ms": 2000,
-      "execution_duration_ms": 300,
-      "activation_delay_ms": 0,
+      "execution_duration_ms": 800,
+      "activation_delay_ms": 400,
```
AngieDutra, line 280: I tried here with 500 and the orbe shoot feels more like the result of the animation instead of having a weird offset

agustinesco:
LGTM! maybe we should research a way to adapt the animation duration in the client to the duration in the back and make the model perform that full animation in a determined period of time, if that's a solution

manucamejo (approved):

PR #350, opened 2024-03-06
Some love to the umbrella repo 

AminArria:
- Fixed arena and game_client apps, they where not properly configured as in an umbrella
- All apps use `.formatter.exs` with `line_length: 120` (debatable, please raise your voice)
- Central credo configuration, same checks applied to all apps
- Removed some old tests that where not needed
- CI now runs on the entire repo, not only `arena` app

Future plans
- Add and require `CODEOWNERS` for each app
- Enforce `credo --strict` 
- Enable more Credo lints that we think useful (have a couple ideas already)
- Any other ideas


manucamejo (approved):

Nico-Sanchez (approved):
Nice job @AminArria 

Thank you for this ❤️

PR #351, opened 2024-03-06
Fix Battles using all units instead of selected units

lotuuu:
Closes https://github.com/orgs/lambdaclass/projects/18/views/7?pane=issue&itemId=56587539

Battle was using all units instead of only selected units. This PR fixes that.

To check if this works:
0- Add an inspect in the code to print the units that are being used in fights. This should be an `IO.inspect(team_1) `call at the top of the `battle` function in `Champions.Battle`

1- Create a user
{:ok, user} = Champions.Users.register("Name")

2- Unselect a unit (get the id from user.units)
Champions.Units.unselect_unit(user.id, unit_id)

3- Run a battle against a level (check #337 for tutorial). Check that the unselected unit is not on the printed team.


agustinesco (approved):
LGTM!

apps/champions/lib/champions/battle.ex
```diff
@@ -7,19 +7,22 @@ defmodule Champions.Battle do
...
+  alias GameBackend.Units
   alias GameBackend.Users
 
   @doc """
   Plays a level for a user, which means fighting its units with their selected ones.
   Returns `:win` or `:loss` accordingly, and updates the user's progress if they win..
   """
   def fight_level(user_id, level_id) do
-    with {:user, {:ok, user}} <- {:user, Users.get_user(user_id)},
+    with {:user, {:ok, _user}} <- {:user, Users.get_user(user_id)},
```
ncontinanza, line 18: What's the reason of this `get_user()` here? Just checking that it exists?

ncontinanza (commented):
Looking good! I left a question to understand better one of the changes

apps/champions/lib/champions/battle.ex
lotuuu: yes!

lotuuu: But we have a new User.exists? function now so i'm going to replace this with that one

ncontinanza (approved):
LGTM!

agustinesco (approved):
LGTM!

PR #352, opened 2024-03-07
Add prod configs

lotuuu:
After PR #350, [arena deploy is not working](https://github.com/lambdaclass/mirra_backend/actions/runs/8192618076) because of missing prod.exs config files in some apps. This PR adds them even if only empty, in order to [fix arena deploys](https://github.com/lambdaclass/mirra_backend/actions/runs/8194927724).

### DIFF
--- a/apps/champions/config/prod.exs
+++ b/apps/champions/config/prod.exs
@@ -1,0 +1,1 @@
+import Config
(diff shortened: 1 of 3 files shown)


Nico-Sanchez (approved):

AminArria:
All good with this, but please check https://github.com/lambdaclass/mirra_backend/pull/354 before merging this to avoid unnecessary conflicts cause that one might should fix things so we can reject this one

lotuuu:
Since #354 fixes this already, this PR is not needed.

PR #353, opened 2024-03-07
Add missing User test

lotuuu:
Closes https://github.com/orgs/lambdaclass/projects/18/views/7?pane=issue&itemId=56588631

Adds tests for GetUser.

Originally this PR would have also added a missing test to units: GetUnits, but we only had the message for it, we hadn't implemented it. After discussing with @tkz00 if this implementation was necessary, we decided it's better to drop it.


agustinesco (approved):
LGTM!

Nico-Sanchez (approved):

PR #356, opened 2024-03-08
Campaign tests

ncontinanza:
Part of https://github.com/lambdaclass/champions_of_mirra/issues/253
This PR adds the following tests to the `champions_test.exs`:

- Get campaigns and levels, testing the GetCampaigns, GetCampaign and GetLevel requests.
- Fight level, a battle's happy path, where the user fights the next level of the campaign progression.
- A test to check that a user can not fight a level that is not the next level in the progression.

Also some errors in the request handling were found, that are fixed in this very same branch.


apps/gateway/test/champions_test.exs
```diff
@@ -295,6 +298,121 @@ defmodule Gateway.Test.Champions do
...
+  describe "campaigns" do
+    test "get campaigns and levels", %{socket_tester: socket_tester} do
+      # Register user
+      {:ok, user} = Users.register("campaign_user")
+
+      # GetCampaigns
+      SocketTester.get_campaigns(socket_tester, user.id)
+      fetch_last_message(socket_tester)
+
+      assert_receive %WebSocketResponse{
```
lotuuu, line 310: Do you think this approach is better than using `get_last_message()`?

apps/gateway/test/champions_test.exs
```diff
@@ -295,6 +298,121 @@ defmodule Gateway.Test.Champions do
...
+
+      fetch_last_message(socket_tester)
+
+      %WebSocketResponse{
+        response_type: {:battle_result, _ = battle_result}
+      } = get_last_message()
+
+      # Battle result should be either win or loss
+      assert battle_result.result == "win" or battle_result.result == "loss"
+
```
lotuuu, line 380: We should check that the campaign progression is updated correctly to the next level! when we win

lotuuu: Maybe we could do this in another test, where we set up a strong team to ensure victory (since we don´t have seeds yet)

lotuuu (changes requested):
Tests are passing succesfully. Left a couple of comments.

apps/gateway/test/champions_test.exs
ncontinanza: I saw the way you used it in previous tests and thought that the intention was clearer using `assert_receive`, do you see any downside to it?

apps/gateway/test/champions_test.exs
ncontinanza: Done!

apps/gateway/test/champions_test.exs
lotuuu: Well the one downside I see here is that we have no timeout to the assertion, but that can be fixed two ways:
a. add the timeout parameter to assert_receive which is nil by default
b. use get_last_message() which already has a timeout

I'm not sure why I created get_last_message if just asserting the response works fine. Thing is, we should probably choose one strategy and stick with it, so we should either do all assert_recieve or all get_last_message().

ncontinanza: As we talked recently, I removed the get_last_message() function and just kept the `assert_receive` with their corresponding matchings.

lotuuu (approved):

apps/gateway/test/champions_test.exs
```diff
@@ -295,13 +280,134 @@ defmodule Gateway.Test.Champions do
...
+      # GetCampaigns
+      SocketTester.get_campaigns(socket_tester, user.id)
+      fetch_last_message(socket_tester)
+
+      assert_receive %WebSocketResponse{
+        response_type: {:campaigns, %Campaigns{} = campaigns}
+      }
+
+      assert Enum.count(campaigns.campaigns) > 0
```
agustinesco, line 296: I think comparing that there's any campaign is not enough, if there's not archiving of soft-delete mechanism we could do a `GameBackend.Repo.all(GameBackend.Campaigns)`  and check if the numbers match to ensure that we're retrieving the campaigns correctly

apps/gateway/test/champions_test.exs
```diff
@@ -295,13 +280,134 @@ defmodule Gateway.Test.Champions do
...
+
+      assert battle_result.result == "win"
+
+      {:ok, advanced_user} = Users.get_user(user.id)
+
+      [advanced_campaign_progression | _] = advanced_user.campaign_progresses
+
+      assert user.id == advanced_user.id
+      assert advanced_campaign_progression.level_id != level_id
+      assert advanced_campaign_progression.level.level_number > level_number
```
agustinesco, line 379: if the progression will be always linear we should check that the number is `level_number + 1` to see that the campaign is progressing correctly

agustinesco (changes requested):

apps/gateway/test/champions_test.exs
```diff
@@ -301,6 +302,13 @@ defmodule Gateway.Test.Champions do
 
       assert Enum.count(campaigns.campaigns) > 0
 
+      # Check the campaign numbers match the expected range
+      campaign_numbers = 1..length(campaigns.campaigns)
+
+      Enum.each(campaign_numbers, fn number ->
+        assert Enum.any?(campaigns.campaigns, &(&1.campaign_number == number))
+      end)
```
agustinesco, line 310: What are we testing here? that the list of campaigns has the correct numbers? if this is due to the previous comment it wasn't my intention, the aim of the comment was to check that the get_campaigns returns all the campaigns available in the database

agustinesco (approved):
LGTM!

lotuuu (approved):

PR #362, opened 2024-03-12
[Rewrite Battle] Schemas

lotuuu:

Only merge this PR if all "Rewrite Battle" PRs are approved and ready to merge. We will "cascade" these PRs starting from the latest one (#383 -> #373 -> #362 -> `main`) so that we never break main, while keeping the PRs not-so-impossible to review.

Modifies battle schemas and the places they are used in. Battles won't work here since we haven't actually implemented handling for new effects.

How to test: Check that you can create skills using the new schema. App will compile with warnings since simulator.ex uses the old implementation. This will be fixed when we merge #373 here.

Sample character creation with all of the new effect fields:
```
{:ok, muflus} =
  GameBackend.Units.Characters.insert_character(%{
    game_id: 2,
    active: true,
    name: "NewChar",
    faction: "Araban",
    quality: Champions.Units.get_quality(:epic),
    base_health: 621,
    base_attack: 63,
    base_defense: 78,
    basic_skill: %{
      name: "Muflus Basic",
      effects: [
        %{
          type: "instant",
          initial_delay: 0,
          components: [%{"type" => "ChanceToApply", "chance" => 0.753}],
          modifiers: [
            %{
              modifier_operation: "Add",
              magnitude_calc_type: "Float",
              float_magnitude: 50.2,
              attribute: "attack"
            }
          ],
          executions: [
            %{"type" => "DealDamage", "attack_ratio" => 80, "energy_recharge" => 50, "delay" => 0}
          ],
          target_strategy: "random",
          target_count: 2,
          target_allies: false,
          target_attribute: nil,
        }
      ],
      cooldown: 5
    },
    ultimate_skill: %{
      name: "Muflus Ultimate",
      effects: [
        %{
          type: "instant",
          initial_delay: 0,
          components: [],
          modifier: [],
          executions: [
            %{"type" => "DealDamage", "attack_ratio" => 205, "energy_recharge" => 0, "delay" => 0}
          ],
          target_strategy: "random",
          target_count: 2,
          target_allies: false,
          target_attribute: "Health"
        }
      ]
    }
  })
```


apps/game_backend/lib/game_backend/units/skills/effects/target_strategy.ex
```diff
@@ -0,0 +1,69 @@
...
+
+  def load(string), do: {:ok, target_strategy_from_string(string)}
+
+  def dump(strategy), do: {:ok, target_strategy_to_string(strategy)}
+
+  defp target_strategy_to_string("random"), do: "random"
+  defp target_strategy_to_string("nearest"), do: "nearest"
+  defp target_strategy_to_string("furthest"), do: "furthest"
+  defp target_strategy_to_string("frontline"), do: "frontline"
+  defp target_strategy_to_string("backline"), do: "backline"
```
agustinesco, line 29: We can replace this with a `when is_binary(var)` guard

apps/game_backend/lib/game_backend/units/skills/effect.ex
```diff
@@ -4,50 +4,93 @@ defmodule GameBackend.Units.Skills.Effect do
...
-      :targeting_strategy,
-      :targets_allies
+      :initial_delay,
+      :target_count,
+      :target_strategy,
+      :target_allies,
+      :target_attribute
     ])
-  end
+    |> validate_change(:executions, fn :executions, executions ->
```
agustinesco, line 45: I really like this usage of pattern match to validate the executions! 🚀

apps/game_backend/lib/game_backend/units/skills/effects/target_strategy.ex
agustinesco: We wont be doing this because, even tho we're validating that the strategy has some of these values, we want to check that in both ways, enconding and decoding

agustinesco (approved):
LGTM!

Nico-Sanchez (approved):

agustinesco (approved):
LGTM! 🪨

PR #363, opened 2024-03-12
Items tests

ncontinanza:
Part of https://github.com/lambdaclass/champions_of_mirra/issues/253

This PR adds the following tests to the champions_test.exs:

- Get an item that belongs to a user.
- Equip an item to a unit and unequip it afterwards.
- Level up an item, and increase its level until the user can't afford more upgrades.


apps/gateway/test/champions_test.exs
```diff
@@ -295,6 +297,105 @@ defmodule Gateway.Test.Champions do
...
+
+      :ok = SocketTester.get_item(socket_tester, user.id, item.id)
+      fetch_last_message(socket_tester)
+
+      %WebSocketResponse{
+        response_type: {:item, %Item{} = fetched_item}
+      } = get_last_message()
+
+      assert fetched_item.id == item.id
+      assert fetched_item.user_id == user.id
```
lotuuu, line 321: Maybe also assert other fields like name and level while we're at it?

apps/gateway/test/champions_test.exs
```diff
@@ -295,6 +297,105 @@ defmodule Gateway.Test.Champions do
...
+      assert fetched_item.unit_id == ""
```
lotuuu, line 322: This is perfect but let's add a comment explaining it just in case it's not clear enough for someone not familiar with our app and protobuf.

Something along the lines of:
`# We expect the item to be unequipped after creation. Since protobuf can't handle null messages, we get an empty string.`

apps/gateway/test/champions_test.exs
```diff
@@ -295,6 +297,105 @@ defmodule Gateway.Test.Champions do
...
+    end
+
+    test "equip and unequip item", %{socket_tester: socket_tester} do
+      # Register user
+      {:ok, user} = Users.register("EquipItemUser")
+
+      unit = user.units |> Enum.at(0)
```
lotuuu, line 329: I think the `[unit | _] = user.units` syntax is cleaner here

apps/gateway/test/champions_test.exs
```diff
@@ -295,6 +297,105 @@ defmodule Gateway.Test.Champions do
...
+      fetch_last_message(socket_tester)
+
+      %WebSocketResponse{
+        response_type: {:item, %Item{} = equipped_item}
+      } = get_last_message()
+
+      # The item is now equipped to the unit
+      assert equipped_item.user_id == user.id
+      assert equipped_item.unit_id == unit.id
+
```
lotuuu, line 351: Before unequipping, let's test equipping the same item to another unit!

apps/gateway/test/champions_test.exs
```diff
@@ -295,6 +297,105 @@ defmodule Gateway.Test.Champions do
...
+      %WebSocketResponse{
+        response_type: {:item, %Item{} = leveled_up_item}
+      } = get_last_message()
+
+      assert leveled_up_item.level == item.level + 1
+      assert Currencies.get_amount_of_currency_by_name(user.id, "Gold") < gold_amount_before_level_up
+
+      # Can not level up item if currency is not enough
+      level_up_item_until_unaffordable(socket_tester, user.id, item.id)
```
lotuuu, line 392: I think it would be better if we set up our user to only have enough currency to upgrade once

lotuuu (changes requested):
Tests run succesfully! Left a couple of comments

apps/gateway/test/champions_test.exs
```diff
@@ -333,4 +459,17 @@ defmodule Gateway.Test.Champions do
 
     same_character_ids ++ same_faction_ids
   end
+
+  defp level_up_item_until_unaffordable(socket_tester, user_id, item_id) do
```
lotuuu, line 463: Let' delete this unused func

lotuuu (changes requested):

lotuuu (approved):

agustinesco (approved):
LGTM!

PR #367, opened 2024-03-12
Fix `not an already existing atom` error

### LINKED ISSUE #358: ArgumentError when starting the game with a skill instead of moving
When you start the game and your first action is a skill, the backend crashes:

```elixir
[error] GenServer #PID<0.2857.0> terminating
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not an already existing atom

    :erlang.binary_to_existing_atom("EXECUTING_SKILL_2", :utf8)
    (arena 0.1.0) lib/arena/game/player.ex:285: Arena.Game.Player.skill_key_execution_action/1
    (arena 0.1.0) lib/arena/game/player.ex:191: Arena.Game.Player.use_skill/4
    (arena 0.1.0) lib/arena/game_updater.ex:70: Arena.GameUpdater.handle_call/3
    (stdlib 5.0.2) gen_server.erl:1113: :gen_server.try_handle_call/4
    (stdlib 5.0.2) gen_server.erl:1142: :gen_server.handle_msg/6
    (stdlib 5.0.2) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
Last message (from #PID<0.2901.0>): {:attack, 1, "2", %Arena.Serialization.AttackParameters{target: %Arena.Serialization.Direction{x: 0.0, y: 0.0, __unknown_fields__: []}, __unknown_fields__: []}, 0}
```

AminArria:
Closes #358 

Replace the dynamic atom creation for the protobuf enum to a static return, although the dynamic part was nice since it converts to an enum it makes little sense, better to crash and remember to add the additional line here

Sadly I wasn't able to reproduce the issue so this fixes it by removing the cause, but the underlying reason remains sort of a mystery (although we have theories)

### DIFF
--- a/apps/arena/lib/arena/game/player.ex
+++ b/apps/arena/lib/arena/game/player.ex
@@ -279,9 +279,9 @@ defmodule Arena.Game.Player do
   ####################
   # Internal helpers #
   ####################
-  defp skill_key_execution_action(skill_key) do
-    "EXECUTING_SKILL_#{String.upcase(skill_key)}" |> String.to_existing_atom()
-  end
+  defp skill_key_execution_action("1"), do: :EXECUTING_SKILL_1
+  defp skill_key_execution_action("2"), do: :EXECUTING_SKILL_2
+  defp skill_key_execution_action("3"), do: :EXECUTING_SKILL_3
 
   defp maybe_trigger_natural_heal(player, true) do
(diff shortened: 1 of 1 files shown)


Nico-Sanchez (approved):
Nice!

agustinesco (approved):
Great! 🔨

PR #372, opened 2024-03-13
Format seeds

lotuuu:
Closes https://github.com/orgs/lambdaclass/projects/18/views/7?pane=issue&itemId=56588673

seeds.exs was not formatted correctly. This PR does just that, and also adds it to the make format command.

I could not find a way to add it to the .formatter.exs subdirectories correctly, so we won't be able to check for its formatting in the credo either.


Makefile
```diff
@@ -18,6 +18,7 @@ purge:
 
 format:
 	mix format
+	mix format priv/repo/seeds.exs
```
AminArria, line 21: Did you try changing `.formatter.exs` with 

```
inputs: ["mix.exs", "config/*.exs", "priv/*/seeds.exs"]
```

AminArria (changes requested):

Makefile
lotuuu: oh, that does it! I thought these were configuration files for the formatter, not the actual files to format. I figured those should go in `subdirectories` 🤔

AminArria (approved):

AminArria (approved):

PR #373, opened 2024-03-13
[Rewrite Battle] Implement Effects

lotuuu:

Overhaul for the Effects of skills (skills and the battle in general have also been retouched) 

# Changes

### State remake
- State now consists of more than just units:
  - Units - what we previously had
  - Skills being cast - for skill delays, that come in the shape of animations and effect triggers
  - Pending effects - for effect delays
  - Pending executions - for execution delays (not yet implemented for PR size issues)

### Animation Delays
 
Skills now have an animation duration (during which the unit won't be able to take other actions) and an animation trigger, which is the amount of steps after animation starts in which the effects are launched.

Effects also have their own delay, until they impact the enemy player. An example of this is a fireball taking some time from when it's launched until it travels the distance to hit its target.

[Here](https://www.notion.so/lambdaclass/Example-of-skill-effect-animation-delays-edb26462668f43b7a687159b64d08ff4?pvs=4) is a notion link to a small example of this!

### New effects

New effects consist of 3 main elements: `components`, `modifiers` and `executions`.
- Components are used to modify the `effect`'s effect
- Modifiers are buffs & debuffs that the effect will apply to its targets for the effect's set duration 
- Executions are instant applications the effect does, for example DealDamage
More detail can be found in modules documentation and in the [Notion page](https://www.notion.so/lambdaclass/Skills-cd9814e237b4459eb1c6c426ed9b167e?pvs=4), however anything that's too hard to grasp might be better off leaving commented in the code for better maintenance.

### Adds info logs to battle

Logs have been added for many different actions of the battle, in order to make it easier to test it out while we lack animations. There's an existing problem going on where `Logger` seems to be overloading (just my assumption) and stops logging at all after a given point for a battle. What's interesting is that it's always the same point, for example step 40 of a 2 player battle (number may vary for your machine). While we look for a fix, this change is still valuable as one can simply replace all `Logger.info` calls with `IO.inspect` to accomplish observability of the battle.

# To Test

Follow the steps on `battle.md` to initiate a battle. You can play with the configuration of a skill by modifying it in the `seeds.exs`


apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
-          new_state
+          current_state
+          |> Map.put(:skills_being_cast, [unit.basic_skill | current_state.skills_being_cast])
           |> put_in(
-            [unit.id, :basic_skill, :remaining_cooldown],
+            [:units, unit.id, :basic_skill, :remaining_cooldown],
             unit.basic_skill.base_cooldown + 1
           )
-          |> put_in([unit.id, :energy], new_state[unit.id][:energy] + 75)
+          |> put_in([:units, unit.id, :energy], current_state.units[unit.id].energy + unit.basic_skill.energy_regen)
```
agustinesco, line 169: you can also use update_in
`|> update_in([:units, unit.id, :energy], fn current_energy -> current_energy + unit.basic_skill.energy_regen end)`

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
         can_cast_basic_skill(unit) ->
-          new_state = cast_skill(unit, unit.basic_skill, initial_step_state, current_state)
+          Logger.info("Unit #{format_unit_name(unit)} casting basic skill")
 
-          new_state
+          current_state
+          |> Map.put(:skills_being_cast, [unit.basic_skill | current_state.skills_being_cast])
           |> put_in(
-            [unit.id, :basic_skill, :remaining_cooldown],
+            [:units, unit.id, :basic_skill, :remaining_cooldown],
```
agustinesco, line 166: Won't this operation always remain the same? is `unit.basic_skill.base_cooldown` being updated somewhere? you can also replace this put in an update_in

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
 
+      new_state =
+        Enum.reduce(new_state.skills_being_cast, new_state, fn skill, current_state ->
+          Logger.info("Process step #{step} for skill #{skill.name} cast by #{String.slice(skill.caster_id, 0..2)}")
+
+          # We need the initial_step_state to decide effect targets
+          process_step_for_skill(skill, current_state, initial_step_state)
+        end)
+
+      new_state =
```
agustinesco, line 89: Since we're updating the same map multiple times i think a best approach would be create a bigger function for each step that traverse the players/skills/effects and do what it needs to do (you can check how `apps/arena/lib/arena/game_updater.ex` handles the game update in line 103)

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
-    )
+    # Reduce modifier remaining timers & remove expired ones
+    new_modifiers = %{
+      additives: reduce_modifier_timers(unit.modifiers.additives, unit),
+      multiplicatives: reduce_modifier_timers(unit.modifiers.multiplicatives, unit),
+      overrides: reduce_modifier_timers(unit.modifiers.overrides, unit)
+    }
 
-    # TODO: decrease_effects_remaining_steps(unit, current_state)
+    # Reduce basic skill cooldown
```
agustinesco, line 182: If you handle cooldowns first you could remove the +1 that this put has `remove_dead_units(new_state)
      |> check_winner(step)`

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
         true ->
           current_state
       end
 
-    put_in(
-      new_state,
-      [unit.id, :basic_skill, :remaining_cooldown],
-      max(new_state[unit.id].basic_skill.remaining_cooldown - 1, 0)
-    )
+    # Reduce modifier remaining timers & remove expired ones
```
agustinesco, line 175: same as above, does the modifiers needs to be in the same function as the skill usage?

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
-          do: {:halt, :timeout},
-          else: {:cont, units}
-
-      result ->
-        {:halt, result}
+  # Calculate the new state of the battle after a step passes for a skill being cast.
+  # Reduces the remaining animation and effect trigger, and casts the effects if the latter has ended.
+  defp process_step_for_skill(skill, current_state, initial_step_state) do
+    # Check if the casting unit has died
+    if Enum.any?(current_state.units, fn {unit_id, _unit} -> unit_id == skill.caster_id end) do
```
agustinesco, line 221: you can change this for Map.has_key(units, unit_id)

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+      {new_skill, new_state} = process_skill_effects_trigger_value(skill, current_state, initial_step_state)
```
agustinesco, line 222: i think `process_skill_effects_trigger_value` it's too short to be extracted in a different method and doesn't adds up to the readability of the code we could replace this with a `cond do` or `case skill do` something like that

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+
+    Map.put(current_state, :pending_effects, List.delete(current_state.pending_effects, effect))
+  end
+
+  # Calculate the new state of the battle after a step passes for a pending effect.
+  # If the effect isn't ready to be processed, we reduce its remaining delay.
+  defp process_step_for_effect(effect, current_state) do
+    Map.put(current_state, :pending_effects, [
+      %{effect | delay: effect.delay - 1} | List.delete(current_state.pending_effects, effect)
+    ])
```
agustinesco, line 281: I'm not sure if i'm a huge fan of this recurrent tactic of updating something from a list deleting it  and adding it later on, maybe we could reduce the effects from the game state and updating/deleting/applying them in the same reduce
```
{updated_effects, updated_game_state}
Enum.reduce(new_state.pending_effects, {[], new_state], fn effect, {result_effect, current_state} ->
  cond do 
     apply_effect -> {updated_effect, updated_game_state}
     # Rest of cases
  end
end
```

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+
+  # Check if the unit can cast their basic skill this step.
+  defp can_cast_basic_skill(unit), do: unit.basic_skill.remaining_cooldown <= 0
+
+  # Called when a skill being cast reaches effect_trigger 0.
+  # "Queues" the effect to be processed when its delay reaches 0.
+  defp trigger_skill_effects(skill, current_state, initial_step_state) do
+    caster = current_state.units[skill.caster_id]
 
+    # We store the caster's state in the effect in case the unit dies before the effect's delay ends
```
agustinesco, line 303: We want to apply the effect if the caster dies before the delay ends? if that's not true why adding the caster to the skill if we're checking that the caster is alive in the `process_step_for_skill` method?

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+
+  # Check if the unit can cast their ultimate skill this step.
+  defp can_cast_ultimate_skill(unit), do: unit.energy >= @ultimate_energy_cost
+
+  # Check if the unit can cast their basic skill this step.
+  defp can_cast_basic_skill(unit), do: unit.basic_skill.remaining_cooldown <= 0
+
+  # Called when a skill being cast reaches effect_trigger 0.
+  # "Queues" the effect to be processed when its delay reaches 0.
+  defp trigger_skill_effects(skill, current_state, initial_step_state) do
```
agustinesco, line 300: i dont think the name "trigger_skill_effect" is suitable for what this method is doing, it leads me to think that we'll change the unit states with the skill effects but we're just updating the effect and adding it to the pending effects list, maybe `queue_skill_effects` or something like that would be more suitable

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+         |> Enum.take_random(count)
          |> Enum.map(fn {id, _unit} -> id end)
 
-  defp apply_effect(%{type: "instant"} = effect, caster, target) do
-    new_value = calculate_value(effect, caster, target)
-    Map.put(target, String.to_atom(effect.stat_affected), new_value)
+  # Apply an effect to its target *if it hits*.
+  # Hit chance is affected by the ChanceToApply component and could be expanded later on.
+  # Returns the new state of the target.
+  defp maybe_apply_effect(effect, target, caster, current_step_number) do
```
agustinesco, line 332: I think `maybe_apply_effect` is only working a wrapper of an if statement for `effect_hits` and `apply_effect`, maybe we could delete `apply_effect` and do it in the same `maybe_apply_effect` method

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+
+          "Override" ->
+            put_in(target, [:modifiers, :overrides], [new_modifier | target.modifiers.overrides])
+        end
+      end)
+
+    target_after_executions =
+      Enum.reduce(effect.executions, target_after_modifiers, fn execution, target_acc ->
+        process_execution(execution, target_acc, caster)
+      end)
```
agustinesco, line 384: You can simple return the `Enum.reduce` without assigning it to target_after_executions

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+        case modifier.modifier_operation do
+          "Add" ->
+            put_in(target, [:modifiers, :additives], [new_modifier | target.modifiers.additives])
+
+          "Multiply" ->
+            put_in(target, [:modifiers, :multiplicatives], [new_modifier | target.modifiers.multiplicatives])
+
+          "Override" ->
+            put_in(target, [:modifiers, :overrides], [new_modifier | target.modifiers.overrides])
+        end
```
agustinesco, line 378: You're putting all the modifiers into the target but later on the `calculate_unit_stat` uses the modifiers from the caster instead of the target

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+           "type" => "DealDamage",
+           "attack_ratio" => attack_ratio,
+           "energy_recharge" => energy_recharge,
+           # TODO
+           "delay" => _delay
+         },
+         target,
+         caster
+       ) do
+    damage = max(floor(attack_ratio * calculate_unit_stat(caster, :attack)), 0)
```
agustinesco, line 401: this approach to the modifiers causes to calculate the attack every time we wan't to deal damage , why not modifying the damage when we add or remove the effects and then simply use it? let me know if you think this simplify the logic or just makes it harder

something like this pseudo code
apply_effect -> 
 added amount = calculate amount
 unit.attribute += added amount
  map.put(modifier, :added_amount, added_amount)
add modifier to unit
remove_effect -> 
 unit.attribute -= effect.added_amount
 remove modifier from unit

apps/champions/lib/champions/battle/simulator.ex
lotuuu: nice!

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
       remove_dead_units(new_state)
       |> check_winner(step)
     end)
   end
 
-  defp process_step(unit, initial_step_state, current_state) do
+  # Removes dead units from the battle state.
+  defp remove_dead_units(state) do
+    new_units =
+      Enum.reduce(state.units, %{}, fn {unit_id, unit}, units ->
```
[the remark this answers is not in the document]
lotuuu: This used to be a filter, but we changed it to a reduce to be able to log when a unit dies.

apps/champions/lib/champions/battle/simulator.ex
lotuuu: nice!

apps/champions/lib/champions/battle/simulator.ex
lotuuu: Mmh I think it's clearer when all the process_step_for_x follow the same pattern but I'll change it if you think it's not

lotuuu: Nevermind, un-nesting it results in a warning because "the funciton body is nested too deep", seems like nesting is the way to go

apps/champions/lib/champions/battle/simulator.ex
lotuuu: I think it's clearer to separate between behaviors. When we start to add more mechanics like dodge or accuracy, we'll end up with functions that are way too big. Separating them now is a good way to reduce function overloads both currently and in the long run

apps/champions/lib/champions/battle/simulator.ex
lotuuu: Not a fan of it either! Your solutions is pretty ingenious, I'll work on implementing it!

apps/champions/lib/champions/battle/simulator.ex
lotuuu: I've intentionally avoided calling it "queue" since it's not technically a queue, as processing depends on the delay and not on order of creation. I chose this name because the configuration name the game designe asked for was "animation_trigger", so I left trigger as the verb, but I can see why one could confuse it with application. Can you think of a third alternative?

apps/champions/lib/champions/battle/simulator.ex
lotuuu: I'm quite pleased with the result, let me know what you think! https://github.com/lambdaclass/mirra_backend/pull/373/commits/6ad8ac4096727feee66b13ec09b245ec86ebb41b

apps/champions/lib/champions/battle/simulator.ex
agustinesco: right!

apps/champions/lib/champions/battle/simulator.ex
lotuuu: I'd rather not sacrifice any more performance than necessary since battles could be susceptible to turning too long to process, so I think going through the units list once will be better than handling it separately. On handling the cooldowns before processing, I think that could work! I'll work on it

apps/champions/lib/champions/battle/simulator.ex
lotuuu: The reason is that modifiers don't necessarily expire in the same order they were applied, so this could happen:

1. Unit has 100 base_damage
2. A modifier that expires in a short time reduces his damage by 20 (additive). Now its 80.
3. A modifier that expires in a long time reduces his damage in half (multiplicative). Now its 40.
4. The additive expires, we restore the 20. Now we have 60.
5. The multiplicative expires, we restore the halving of the damage. Now its 120.

The unit now has more damage than what it started with. This is bad! Even if we found a workaround for restoring, the result for the stat during modifiers would depend on the order they were applied, which is also probably bad

apps/champions/lib/champions/battle/simulator.ex
agustinesco: Hmm we could do something like `add_skill_effects_to_pending_effects` or `append_skill_effects_for_application`

apps/champions/lib/champions/battle/simulator.ex
lotuuu: I think there might be a misunderstanding here. calculate_unit_stat is not used to set new values, it's used to get how the stat of a unit looks like right now, when you factor in the modifiers that affect it. It's used, for example, when calculating how much the attack_ratio of a DealDamage execution should be multiplied by. The comment above calculate_unit_stat "Calculate the current amount of the given attribute that the unit has, based on its modifiers." aims to explain this. Let me kwow if you think I should add any more comments to make this clearer for future reviewers/mantainers.

apps/champions/lib/champions/battle/simulator.ex
lotuuu: I added it to explain what was going on in the reduce as a sort of implicit documentation. Do you think it's unnecessary?

apps/champions/lib/champions/battle/simulator.ex
agustinesco: hmmm maybe we can solve this with pattern match checking if the effect should be applied in the call?

```
maybe_apply_effect(effect, current_state.units[id], current_state.step_number, effect_hits?(effect))

defp maybe_apply_effect(_effect, target, _current_step_number, false = _should_apply?), do: target
defp maybe_apply_effect(%{caster: caster} = effect, target, current_step_number, should_apply?), do: destroy target
```

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,459 @@ defmodule Champions.Battle.Simulator do
...
+  # For now this applies the executions on the spot.
+  # Later on, it will "cast" them as we do with skills and effects to account for execution delays.
+  # Returns the new state of the target.
+  defp apply_effect(effect, target, caster, current_step_number) do
+    target_after_modifiers =
+      Enum.reduce(effect.modifiers, target, fn modifier, target ->
+        # If it's permanent, we set its duration to -1
+        new_modifier =
+          modifier
+          |> Map.put(:remaining_steps, Map.get(effect.type, "duration", -1))
```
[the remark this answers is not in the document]
lotuuu: No, since it is only called once for each effect (when it is applied). Afterwards, it is reduced when we call reduce_modifier_timers.

apps/champions/lib/champions/battle/simulator.ex
agustinesco: I really like the change! it's a lot easier to follow now. 
just a small comment, you can iterate over an enumerable and create a new map with `Map.new` so if you iterate the units changing them you can then merge that map with the one in the state, something like:

```
targets_after_effect =
              Map.new(effect.targets, fn id ->
               {id,
                maybe_apply_effect(effect, current_state.units[id], effect.caster, current_state.step_number)}
              end)

new_state =
              update_in(current_state, [:units], fn units -> Map.merge(units, targets_after_effect) end)
```

apps/champions/lib/champions/battle/simulator.ex
lotuuu: 1. We **do** want to apply the effect if the caster dies before the **effect delay** ends, if the effect has already been triggered. When we don't want to do that is when the effect hasn't been triggered and the unit dies, and that's handled on skill cooldown reduce.
2. I'll explain another reason for this implementation with an example:
- You receive a buff that duplicates your damage
- You cast a fireball
- The fireball takes a while to travel the screen
- Meanwhile, you lose your buff
- Your fireball finally hits its target
- The damage it deals isn't doubled because your buff expired, even though the fireball was cast when you indeed had it.
This is not the behaviour we want! I'll add another comment to mention this in the comments since it's missing

apps/champions/lib/champions/battle/simulator.ex
lotuuu: add_skill_effects_to_pending_effects works just fine!

apps/champions/lib/champions/battle/simulator.ex
agustinesco: hmmm yeah, you can already tell that the output will be the target with the effect executed by the function name, maybe if you want to be clearer you can rename the function from `apply_effect` to `apply_effect_to_unit` or `apply_effect_to_target` or `execute_effects_on_unit`

apps/champions/lib/champions/battle/simulator.ex
lotuuu: I tried getting rid of the maybe_apply_effect function and ran into the same "function body nested too deep" credo error.

If you find calling the `effect_hits?` function from within `maybe_apply_effect`'s params clearer than an `if` clause, say the word and we'll go for it. I really do not see the advantage in readibility but it may be because of me getting used to the code I wrote.

apps/champions/lib/champions/battle/simulator.ex
lotuuu: Done in 0af43030! It adds more wrapper functions though 🤔

apps/champions/lib/champions/battle/simulator.ex
lotuuu: Yeah that's cleaner!

apps/champions/lib/champions/battle/simulator.ex
agustinesco: We talked about cooldown and we couldn't find a better solution to the +1 issue, so we'll just document it and go on.

Also i don't think we should optimize before hand, maybe reaching a first iteration of the game is more important than make it performant.

apps/champions/lib/champions/battle/simulator.ex
agustinesco: ooooh okey, this aim to make the effects not dynamic, what i mean with this is that the modifiers will only be applied when the ability is casts and won't affect the ability later on. Sounds good!

apps/champions/lib/champions/battle/simulator.ex
agustinesco: yeah i think it's better because we can get rid of the `apply_effect` method and just do what we need to do without calling anything else. I'm open to debate tho

apps/champions/lib/champions/battle/simulator.ex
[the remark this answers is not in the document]
agustinesco: nice

apps/champions/lib/champions/battle/simulator.ex
lotuuu: We'll just agree to disagree then 😅

I'll implement your function with the boolean as parameter!

apps/champions/lib/champions/battle/simulator.ex
agustinesco: hmmm no, i think i was in the wrong here not understanding. After reading it again i get it now!

apps/champions/lib/champions/battle/simulator.ex
agustinesco: I think this problem of modifying some attibute from a player is very common since i've seen it in curse of mirra before too 😂  ok it has sense now! thanks

apps/champions/lib/champions/battle/simulator.ex
agustinesco: yeah, i think the change is worth tho. 

```
initial_step_state
        |> Map.put(:step_number, step)
        |> process_step_for_units()
        |> process_step_for_skills(initial_step_state)
        |> process_step_for_effects()
```

it's a lot more readable than the previous iteration imo! 🪨

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,465 @@ defmodule Champions.Battle.Simulator do
...
+  # Calculate the new state of the battle after a step passes for a skill being cast, specifically for its `effects_trigger` value.
+  # If the effect hasn't triggered yet, reduce the remaining effects_trigger counter.
+  defp process_skill_effects_trigger_value(skill, current_state, _initial_step_state) do
+    {%{skill | effects_trigger: skill.effects_trigger - 1}, current_state}
+  end
 
-  defp cast_skill(unit, skill, initial_step_state, current_state) do
-    target_ids = choose_targets(unit, skill, initial_step_state)
+  # Calculate the new state of the battle after a step passes for all pending effects
+  def process_step_for_effects(state) do
```
agustinesco, line 262: 🪨

agustinesco (approved):
LGTM! tested and worked as expected! nice work on this rework @lotuuu 🚀

AminArria:
>  There's an existing problem going on where Logger seems to be overloading (just my assumption) and stops logging at all after a given point for a battle

Is this still an issue?

lotuuu:
> >  There's an existing problem going on where Logger seems to be overloading (just my assumption) and stops logging at all after a given point for a battle
> 
> 
> 
> Is this still an issue?

Yes, it is

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -63,187 +64,465 @@ defmodule Champions.Battle.Simulator do
...
 
-          new_state
+          current_state
+          |> Map.put(:skills_being_cast, [unit.basic_skill | current_state.skills_being_cast])
           |> put_in(
-            [unit.id, :basic_skill, :remaining_cooldown],
+            [:units, unit.id, :basic_skill, :remaining_cooldown],
+            # We need this + 1 because we're going to reduce the cooldown at the end of the step
             unit.basic_skill.base_cooldown + 1
           )
```
Nico-Sanchez, line 161: Can't you just avoid the cooldown decrease if the skill being cast is a basic, instead of doing this trick?

lotuuu: mmh this is the same conversation that arose with @agustinesco [here](https://github.com/lambdaclass/mirra_backend/pull/373#discussion_r1541545956) 🤔

It's clearly a faulty point in the implementation but I think it's also the best option we currently have. If we were to undo the trick, we'd have to replicate the cooldown reduction process in every single other action we create, which doesn't scale nicely, so the trick attempts to fix just that.

Nico-Sanchez (approved):

PR #374, opened 2024-03-13
Send skill damage in configSkill

BertovDev:
Clinet -> https://github.com/lambdaclass/curse_of_mirra/pull/1585

### DIFF
--- a/apps/arena/lib/arena/serialization/messages.pb.ex
+++ b/apps/arena/lib/arena/serialization/messages.pb.ex
@@ -1,7 +1,7 @@
 defmodule Arena.Serialization.GameStatus do
   @moduledoc false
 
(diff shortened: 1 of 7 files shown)


apps/arena/lib/arena/serialization/messages.pb.ex
```diff
@@ -483,7 +484,7 @@ end
 defmodule Arena.Serialization.KillEntry do
   @moduledoc false
 
-  use Protobuf, protoc_gen_elixir_version: "0.12.0", syntax: :proto3
+  use Protobuf, syntax: :proto3, protoc_gen_elixir_version: "0.12.0"
```
Nico-Sanchez, line 487: I think you're using the wrong version of elixir, `syntax: :proto3` should be at the end when formatting it.

Nico-Sanchez (changes requested):

PR #378, opened 2024-03-14
Cooldown mechanisms

AminArria:
We introduce a new field for skills called `cooldown_mechanism` which (for now) can be one of two values
- `time` the cooldown is based on time since last execution, this time is read from `cooldown_ms` field. When a skill has this mechanism it will not use stamina or anything related to that
- `stamina` their is no real cooldown rather the skill can be used as long as you have enough stamina. Skills using this won't ever be added to the cooldowns list

You can use https://github.com/lambdaclass/curse_of_mirra/pull/1595 to test or `main`, either thing should work

### DIFF
--- a/apps/arena/lib/arena/configuration.ex
+++ b/apps/arena/lib/arena/configuration.ex
@@ -21,10 +21,23 @@ defmodule Arena.Configuration do
 
   defp parse_skills_config(skills_config) do
     Enum.reduce(skills_config, [], fn skill_config, skills ->
-      mechanics = parse_mechanics_config(skill_config.mechanics)
-      skill = %{skill_config | mechanics: mechanics}
+      skill = parse_skill_config(skill_config)
       [skill | skills]
     end)
+  end
+
+  defp parse_skill_config(%{cooldown_mechanism: "stamina", stamina_cost: cost} = skill_config) when cost >= 0 do
+    mechanics = parse_mechanics_config(skill_config.mechanics)
+    %{skill_config | mechanics: mechanics}
+  end
+
+  defp parse_skill_config(%{cooldown_mechanism: "time", cooldown_ms: cooldown} = skill_config) when cooldown >= 0 do
+    mechanics = parse_mechanics_config(skill_config.mechanics)
+    %{skill_config | mechanics: mechanics}
+  end
+
+  defp parse_skill_config(skill_config) do
+    raise "Invalid Skill config for `#{skill_config[:name]}`: cooldown_mechanism or corresponding attribute is invalid"
   end
 
   defp parse_mechanics_config(mechanics_config) do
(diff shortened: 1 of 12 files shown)


apps/arena/lib/arena/game_launcher.ex
```diff
@@ -4,7 +4,7 @@ defmodule Arena.GameLauncher do
   use GenServer
 
   # Amount of clients needed to start a game
-  @clients_needed 10
+  @clients_needed 1
```
manucamejo, line 7: Please, revert this!

manucamejo (changes requested):

Nico-Sanchez (changes requested):
When a skill is on cooldown and I try to trigger it I lose control of my character.
I don't know if this should be fixed here or disable the button in the client.
Let's either fix it here or wait for a client PR that fixes this.

https://github.com/lambdaclass/mirra_backend/assets/65268603/ae77c46f-58b3-4063-80b2-22eadbe9d0b0

manucamejo:
> When a skill is on cooldown and I try to trigger it I lose control of my character. I don't know if this should be fixed here or disable the button in the client. Let's either fix it here or wait for a client PR that fixes this.
> Screen.Recording.2024-03-14.at.19.02.55.mov

I think that is a front end issue, @BertovDev can you check it?

BertovDev:
> > When a skill is on cooldown and I try to trigger it I lose control of my character. I don't know if this should be fixed here or disable the button in the client. Let's either fix it here or wait for a client PR that fixes this.
> > Screen.Recording.2024-03-14.at.19.02.55.mov
> 
> I think that is a front end issue, @BertovDev can you check it?

Sure! Fixing it in my PR in the client.

apps/arena/lib/arena/game_launcher.ex
AminArria: Done

apps/arena/lib/arena/configuration.ex
```diff
@@ -21,12 +21,25 @@ defmodule Arena.Configuration do
...
+    %{skill_config | mechanics: mechanics}
+  end
+
+  defp parse_skill_config(%{cooldown_mechanism: "time", cooldown_ms: cooldown} = skill_config) when cooldown >= 0 do
+    mechanics = parse_mechanics_config(skill_config.mechanics)
+    %{skill_config | mechanics: mechanics}
+  end
+
+  defp parse_skill_config(skill_config) do
+    raise "Invalid Skill config for `#{skill_config[:name]}`: cooldown_mechanism or corresponding attribute is invalid"
```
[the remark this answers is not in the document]
AminArria: Done

Nico-Sanchez (approved):

manucamejo (approved):

agustinesco (approved):
LGTM! 🪨

PR #379, opened 2024-03-15
Afk rewards tests

ncontinanza:
Closes https://github.com/lambdaclass/champions_of_mirra/issues/253


apps/gateway/test/champions_test.exs
```diff
@@ -300,6 +301,114 @@ defmodule Gateway.Test.Champions do
...
+             end)
+
+      assert Enum.any?(claimed_user.currencies, fn currency ->
+               user_gems_currency = Enum.find(claimed_user.currencies, &(&1.currency.name == "Gems"))
+               initial_gems = user_initial_currencies |> Enum.find(&(&1.currency.name == "Gems"))
+
+               user_gems_currency.amount > initial_gems.amount
+             end)
+
+      # TODO: check that the afk rewards rates have been reset after [CHoM-380] is solved
```
lotuuu, line 377: Let's create an issue for this too! And link it here so the task isn't lost in the code

apps/gateway/test/champions_test.exs
```diff
@@ -300,6 +301,114 @@ defmodule Gateway.Test.Champions do
...
+      {:ok, user} = Users.register("AfkRewardsUser")
+      user_initial_currencies = user.currencies
+
+      # Get initial afk rewards
+      SocketTester.get_afk_rewards(socket_tester, user.id)
+      fetch_last_message(socket_tester)
+
+      assert_receive %WebSocketResponse{response_type: {:afk_rewards, %AfkRewards{afk_rewards: initial_afk_rewards}}}
+
+      # Check that te initial afk reward rates are all 0
```
lotuuu, line 315:
```suggestion
      # Check that the initial afk reward rates are all 0
```

apps/gateway/test/champions_test.exs
```diff
@@ -300,6 +301,114 @@ defmodule Gateway.Test.Champions do
...
+      SocketTester.claim_afk_rewards(socket_tester, advanced_user.id)
+      fetch_last_message(socket_tester)
+      assert_receive %WebSocketResponse{response_type: {:user, %User{} = claimed_user}}
+
+      # Check that the user has received gold and gems
+      assert Enum.any?(claimed_user.currencies, fn currency ->
+               user_gold_currency = Enum.find(claimed_user.currencies, &(&1.currency.name == "Gold"))
+               initial_gold = user_initial_currencies |> Enum.find(&(&1.currency.name == "Gold"))
+
+               user_gold_currency.amount > initial_gold.amount
```
lotuuu, line 367: maybe we could check that the reward is inside a short range around the afk reward rate * the time passed?

apps/gateway/test/champions_test.exs
```diff
@@ -300,6 +301,114 @@ defmodule Gateway.Test.Champions do
...
+      fetch_last_message(socket_tester)
+      assert_receive %WebSocketResponse{response_type: {:user, %User{} = more_advanced_user}}
+
+      # Check that the gold and gems afk rewards rates are now greater than the rewards before the second battle
+      gold_currency_id = Currencies.get_currency_by_name!("Gold").id
+      gems_currency_id = Currencies.get_currency_by_name!("Gems").id
+
+      assert Enum.any?(more_advanced_user.afk_reward_rates, fn rate ->
+               previous_rate = Enum.find(advanced_user.afk_reward_rates, &(&1.currency_id == gold_currency_id)).rate
+               rate.currency_id == gold_currency_id && rate.rate > previous_rate
```
lotuuu, line 402: I wouldn't compare against the previous rate here, I'd rather check that the new rate is the one set on the level rewards

apps/gateway/test/champions_test.exs
```diff
@@ -300,6 +301,114 @@ defmodule Gateway.Test.Champions do
...
+             end)
+
+      assert Enum.any?(more_advanced_user.afk_reward_rates, fn rate ->
+               previous_rate = Enum.find(advanced_user.afk_reward_rates, &(&1.currency_id == gems_currency_id)).rate
+               rate.currency_id == gems_currency_id && rate.rate > previous_rate
```
lotuuu, line 407: Same

lotuuu (changes requested):

apps/gateway/test/champions_test.exs
```diff
@@ -352,26 +352,35 @@ defmodule Gateway.Test.Champions do
...
+      # Check that the user has received gold and gems.
+      # The amount should be greater than the initial amount, and be in the range of the expected amount considering the time waited.
+      # We add 1 to the time waited to account for the time it takes to process the message.
       assert Enum.any?(claimed_user.currencies, fn currency ->
                user_gold_currency = Enum.find(claimed_user.currencies, &(&1.currency.name == "Gold"))
                initial_gold = user_initial_currencies |> Enum.find(&(&1.currency.name == "Gold"))
+               reward_rate = Enum.find(claimed_user.afk_reward_rates, &(&1.currency_id == gold_currency_id)).rate
 
-               user_gold_currency.amount > initial_gold.amount
+               user_gold_currency.amount in initial_gold.amount..trunc(
```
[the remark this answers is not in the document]
lotuuu: With this test, claim could be not working at all and it would still pass since the `initial_gold` will always be in the range `initial_gold..N`

apps/gateway/test/champions_test.exs
```diff
@@ -406,14 +407,21 @@ defmodule Gateway.Test.Champions do
...
+      # Get the current level number and check that the afk rewards rates have increased proportionally
+      current_level_id = hd(more_advanced_user.campaign_progresses).level_id
+      {:ok, level} = Campaigns.get_level(current_level_id)
+      current_level_number = level.level_number
+
       assert Enum.any?(more_advanced_user.afk_reward_rates, fn rate ->
-               reward_rate = Enum.find(more_advanced_user.afk_reward_rates, &(&1.currency_id == gold_currency_id)).rate
-               rate.currency_id == gold_currency_id && rate.rate == reward_rate
+               previous_rate = Enum.find(advanced_user.afk_reward_rates, &(&1.currency_id == gold_currency_id)).rate
+               new_rate = previous_rate + 10 * current_level_number
```
lotuuu, line 417: Let's not "hardcode" the function to build the value, but rather get it by querying the `Level's` `afk_rewards_increments`

lotuuu (changes requested):

lotuuu (changes requested):

apps/gateway/test/champions_test.exs
```diff
@@ -477,14 +477,17 @@ defmodule Gateway.Test.Champions do
...
-
       # Claim afk rewards
       gold_before_claiming = Currencies.get_amount_of_currency_by_name(advanced_user.id, "Gold")
       gems_before_claiming = Currencies.get_amount_of_currency_by_name(advanced_user.id, "Gems")
 
+      # Simulate waiting 2 seconds before claiming the rewards
+      seconds_to_wait = 2
+      {:ok, advanced_user_db} = Users.get_user(advanced_user.id)
+      advanced_user_db
+      |> GameBackend.Users.User.changeset(%{last_afk_reward_claim: DateTime.utc_now() |> DateTime.add((-1) * seconds_to_wait, :second)})
```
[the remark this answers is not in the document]
lotuuu: You won't be able to commit this since after formatting it updated it to `(-1 * seconds_to_wait)`. I still think removing the 1 altogether makes the code clearer though

lotuuu (dismissed):

apps/gateway/test/champions_test.exs
```diff
@@ -424,6 +428,142 @@ defmodule Gateway.Test.Champions do
...
+      SocketTester.get_user(socket_tester, user.id)
+      fetch_last_message(socket_tester)
+      assert_receive %WebSocketResponse{response_type: {:user, %User{} = advanced_user}}
+
+      # Check that the gold and gems afk rewards rates are now greater than the initial rewards
+      gold_currency_id = Currencies.get_currency_by_name!("Gold").id
+      gems_currency_id = Currencies.get_currency_by_name!("Gems").id
+
+      assert Enum.any?(advanced_user.afk_reward_rates, fn rate ->
+               rate.currency_id == gold_currency_id && rate.rate > 0
```
agustinesco, line 473: You can also check the currency by name and avoid using a query here, something like:
`rate.currency.name == "Gold" && rate.rate > 0`

apps/gateway/test/champions_test.exs
```diff
@@ -424,6 +428,142 @@ defmodule Gateway.Test.Champions do
...
+               expected_amount = trunc(gold_before_claiming + reward_rate * seconds_to_wait)
+               user_gold_currency.amount in expected_amount..trunc(expected_amount * 1.1)
+             end)
+
+      assert Enum.any?(claimed_user.currencies, fn currency ->
+               user_gems_currency = Enum.find(claimed_user.currencies, &(&1.currency.name == "Gems"))
+               reward_rate = Enum.find(claimed_user.afk_reward_rates, &(&1.currency_id == gems_currency_id)).rate
+               expected_amount = trunc(gems_before_claiming + reward_rate * seconds_to_wait)
+               user_gems_currency.amount in expected_amount..trunc(expected_amount * 1.1)
+             end)
```
agustinesco, line 514: I think we can replace this two `Enum.any ` with an `Enum.all` and check that every currency in the user has the proper increase instead of hardcoding "Gold" and "Gems"

apps/gateway/test/champions_test.exs
```diff
@@ -424,6 +428,142 @@ defmodule Gateway.Test.Champions do
...
+      gold_currency_id = Currencies.get_currency_by_name!("Gold").id
+      gems_currency_id = Currencies.get_currency_by_name!("Gems").id
+
+      # Get the current level number and check that the afk rewards rates have increased proportionally
+      current_level_id = hd(more_advanced_user.campaign_progresses).level_id
+
+      current_level_afk_rewards_increments =
+        Repo.all(from(r in CurrencyReward, where: r.level_id == ^current_level_id and r.afk_reward))
+
+      assert Enum.any?(more_advanced_user.afk_reward_rates, fn rate ->
```
agustinesco, line 545: same as above

agustinesco (changes requested):

apps/gateway/test/champions_test.exs
ncontinanza: The reason for this is that an `AFKRewardRate` has no currency field, only the corresponding `currency_id`. That's why I'm using the id to compare.

ncontinanza: I found sort of a workaround for this, tell me what you think about it

agustinesco: aaaahh okey, that's true the struct `AfkRewardRate ` doesn't have this field in the `gateway.proto` file

agustinesco (approved):
LGTM!

lotuuu (approved):

PR #388, opened 2024-03-19
Configure releases

lotuuu:
Configures releases for deploys. Unifies scripts into one and adds a RELEASE flag for them. Worklfows that use this new method already exist in main and have been tested:
https://github.com/lambdaclass/mirra_backend/actions/runs/8396000150
https://github.com/lambdaclass/mirra_backend/actions/runs/8396399047
https://github.com/lambdaclass/mirra_backend/actions/runs/8396386117


Nico-Sanchez (approved):

devops/deploy.sh
```diff
@@ -8,31 +8,31 @@ fi
...
-mix release --overwrite
+mix release ${RELEASE} --overwrite
+if [ ${RELEASE} == "game_backend" ]; then
+	mix ecto.migrate
+fi
 
 rm -rf $HOME/mirra_backend
 mv /tmp/mirra_backend $HOME/
 
 mkdir -p $HOME/.config/systemd/user/
```
samoht9277, line 27:
```suggestion
mkdir -p $HOME/.config/systemd/user/

if [ "$(ls ~/.config/systemd/user/*.service 2>/dev/null | wc -l)" -ne 1 ]; then
    exit 1
fi
```

This `if` statement checks if more than one `systemd` service is installed on the systemd service location.

devops/deploy.sh
```diff
@@ -8,31 +8,35 @@ fi
...
```
samoht9277, line 27:
```suggestion
mkdir -p $HOME/.config/systemd/user/

if [ "$(ls ~/.config/systemd/user/*.service 2>/dev/null | wc -l)" -gt 1 ]; then
    exit 1
fi
```

devops/deploy.sh
```diff
@@ -42,16 +47,17 @@ LimitNOFILE=4000
 WantedBy=multi-user.target
 EOF
 
-systemctl --user enable arena
+systemctl --user enable $RELEASE
 
 cat <<EOF >$HOME/.env
 PHX_HOST=${PHX_HOST}
```
samoht9277, line 53:
```suggestion
PHX_HOST=${PHX_HOST}
DATABASE_URL=${DATABASE_URL}
```

samoht9277: I believe we should add the database URL, because this `.env` file is sourced at runtime. 

Meaning that these environment variables are loaded when the compiled binary is executed by systemd. Not passing this variable would result in an error, if I'm not mistaken.

samoht9277 (approved):

Nico-Sanchez (approved):

PR #389, opened 2024-03-19
Initial implementation of effects when killing another player

AminArria:


apps/arena/lib/arena/game/skill.ex
```diff
@@ -351,6 +351,15 @@ defmodule Arena.Game.Skill do
     end
   end
 
+  defp do_effect_mechanics(_game_state, player, effect, {:speed_boost, speed_boost_params}) do
+    Process.send_after(self(), {:remove_speed_boost, player.id, speed_boost_params.amount}, effect.duration_ms)
+    Map.put(player, :speed, player.speed + speed_boost_params.amount)
+  end
+
+  defp do_effect_mechanics(_game_state, player, _effect, {:damage_up, _}) do
```
Nico-Sanchez, line 359: What about the invisible one?

apps/arena/lib/arena/game/skill.ex
```diff
@@ -288,8 +288,8 @@ defmodule Arena.Game.Skill do
     Enum.reduce(effect.effect_mechanics, player, fn {mechanic_name, mechanic_params} = mechanic, player ->
       should_re_apply? =
         is_nil(Map.get(mechanic_params, :last_application_time)) or
-          now - Map.get(mechanic_params, :last_application_time) >=
-            mechanic_params.effect_delay_ms
+          (not is_nil(Map.get(mechanic_params, :effect_delay_ms)) and
```
Nico-Sanchez, line 291: What is this new check for?

apps/arena/lib/arena/game/player.ex
```diff
@@ -314,6 +314,11 @@ defmodule Arena.Game.Player do
     |> Enum.uniq()
   end
 
+  def apply_on_death_effect(player, effect, effect_id) do
```
Nico-Sanchez, line 317: Shouldn't be `on_kill`? Since the given player is the killer (who receives the effect).

apps/arena/lib/arena/game_updater.ex
```diff
@@ -314,20 +314,24 @@ defmodule Arena.GameUpdater do
...
+      |> update_in([:aditional_info, :kill_count], fn count -> count + 1 end)
+
     amount_of_power_ups = get_amount_of_power_ups(victim, game_config.power_ups.power_ups_per_kill)
 
     state =
-      update_in(state, [:game_state, :killfeed], fn killfeed -> [entry | killfeed] end)
-      |> update_in([:game_state, :players, killer_id, :aditional_info, :kill_count], fn count ->
-        count + 1
-      end)
+      put_in(state, [:game_state, :players, killer.id], killer)
```
Nico-Sanchez, line 333: Maybe we could do something here like the `:update_game` callback?
Something like:
```elixir
state =
  state
  |> update_killer_player()
  |> update_killfeed()
``` 
that way we could isolate those behaviors in separate functions and is easier to code/follow.

apps/arena/priv/config.json
```diff
@@ -491,6 +495,26 @@
...
+      "effect_mechanics": {
+        "damage_up": {
+          "amount": 30
+        }
+      }
+    },
+    {
+      "name": "h4ck_death",
+      "duration_ms": 5000,
+      "remove_on_action": false,
```
Nico-Sanchez, line 512: These effects should be called speed and damage. I know they belong to muflus and h4ck, and they trigger on death, but that may change in the future and could be reusable like the invisible one.

Nico-Sanchez (changes requested):
Please add a PR description and attach its issue (if it exists).

I left some comments and I tested it by killing a h4ck and it indeed increases my speed for a while. I don't know if this is enough.

apps/arena/lib/arena/game/skill.ex
AminArria: It doesn't have mechanics... That's a pending refactor, but not something for now

apps/arena/lib/arena/game/skill.ex
AminArria: Not all mechanics need to be reapplied so no `effect_delay_ms`

apps/arena/lib/arena/game/player.ex
AminArria: Honestly it should just be `apply_effect` but I didn't want to deal the clash with the other functions... I'll rename to `put_effect`

apps/arena/priv/config.json
AminArria: That makes sense until we have 2 speed (or whatever) effects with slightly different configs. Also, the client probably uses the name of the effect for the visual indicators

PS: once we move items to be effect we do have 2 speed effects

apps/arena/lib/arena/game/player.ex
AminArria: Done https://github.com/lambdaclass/mirra_backend/pull/389#:~:text=to%20put_effect/3-,85e4be6,-Refactor%20handle_info(%3Ato_killfeed

apps/arena/lib/arena/game_updater.ex
AminArria: Made a refactor for this https://github.com/lambdaclass/mirra_backend/pull/389/commits/84b198e65cd780f63953eafb1a16ce56bd544e66

marco-paulucci:
Is this a different mechanic than power-ups? Because power-ups are meant to be permanent and also farmable from crates around the map. For temporal effects that are less generic like HP+DMG and do something interesting the goal imo would be to use items

PR #390, opened 2024-03-20
Dynamic execution duration for leap

AminArria:
Changes leap from having a fixed duration and dynamic speed to having fixed speed and dynamic duration


apps/arena/lib/arena/game/player.ex
```diff
@@ -362,4 +357,23 @@ defmodule Arena.Game.Player do
         player
     end
   end
+
+  ## Yes, we are pattern matching on exactly one mechanic. As of time writing we only have one mechanic per skill
+  ## so to simplify my life an executive decision was made to take thas as a fact
+  ## When the time comes to have more than one mechanic per skill this function will need to be refactored, good thing
+  ## is that it will crash here so not something we can ignore
+  defp calculate_duration(%{mechanics: [%{leap: leap}]}, position, direction) do
```
manucamejo, line 365: This is not working. It should be:
`defp calculate_duration(%{mechanics: [leap: leap]}, position, direction) do`

manucamejo, line 365: This is not working. It should be:
`defp calculate_duration(%{mechanics: [leap: leap]}, position, direction) do`

manucamejo (changes requested):

apps/arena/lib/arena/game/player.ex
AminArria: Fixed https://github.com/lambdaclass/mirra_backend/pull/390/commits/7864206913ebb60788f65b02aedf1a7e47d19e62

apps/arena/lib/arena/game/player.ex
```diff
@@ -396,6 +391,25 @@ defmodule Arena.Game.Player do
...
+  ## is that it will crash here so not something we can ignore
+  defp calculate_duration(%{mechanics: [{:leap, leap}]}, position, direction) do
+    ## TODO: Cap target_position to leap.range
+    target_position = %{
+      x: position.x + direction.x * leap.range,
+      y: position.y + direction.y * leap.range
+    }
+
+    ## TODO: Magic number needs to be replaced with state.game_config.game.tick_rate_ms
+    Physics.calculate_duration(position, target_position, leap.speed) * 30
```
agustinesco, line 406: not in this pr but maybe we can aslo include this number in the config.exs file

apps/arena/lib/arena/game/player.ex
```diff
@@ -396,6 +391,25 @@ defmodule Arena.Game.Player do
     end
   end
 
+  ## Yes, we are pattern matching on exactly one mechanic. As of time writing we only have one mechanic per skill
+  ## so to simplify my life an executive decision was made to take thas as a fact
+  ## When the time comes to have more than one mechanic per skill this function will need to be refactored, good thing
+  ## is that it will crash here so not something we can ignore
+  defp calculate_duration(%{mechanics: [{:leap, leap}]}, position, direction) do
+    ## TODO: Cap target_position to leap.range
```
agustinesco, line 399: For what i know the distance is already caped at the leap range right?

apps/arena/lib/arena/game/skill.ex
```diff
@@ -268,14 +262,21 @@ defmodule Arena.Game.Skill do
...
 
     ## TODO: Magic number needs to be replaced with state.game_config.game.tick_rate_ms
-    speed = Physics.calculate_speed(entity.position, target_position, leap.duration_ms) * 30
+    duration = Physics.calculate_duration(entity.position, target_position, leap.speed) * 30
+
+    Process.send_after(
+      self(),
+      {:stop_leap, entity.id, entity.aditional_info.base_speed, leap.on_arrival_mechanic},
+      duration
+    )
```
agustinesco, line 271: we're calculating the duration twice, maybe we could include it in the skill params map that `do_mechanic` receives and use that, what do you think? To do so you need to add the duration to the send message of `delayed_skill_mechanics`

apps/arena/priv/config.json
```diff
@@ -948,8 +948,8 @@
       "mechanics": [
         {
           "leap": {
-            "range": 1500.0,
-            "duration_ms": 600,
+            "range": 1000.0,
```
agustinesco, line 951: Why reducing the range?

apps/arena/lib/arena/game/player.ex
AminArria: Nope, there is no backend validation on the actual position you are sending. Meaning that a hacked client could in theory send anything

apps/arena/lib/arena/game/skill.ex
AminArria: Nice catch, moved it as you said

apps/arena/priv/config.json
AminArria: Because merging is hard... Fixed now

agustinesco: 😂

agustinesco (changes requested):
I've found another error i think that, if you leap with the magic boots the final positions ends up being farther that you intended

https://github.com/lambdaclass/mirra_backend/assets/106101218/654d3e6b-e68f-49e4-a71a-27f41ad555b4

agustinesco (approved):
LGTM!

manucamejo (approved):

agustinesco (approved):

PR #391, opened 2024-03-20
Restore item seeds

lotuuu:
Closes https://github.com/lambdaclass/champions_of_mirra/issues/289

Item seeds were deleted in #295 because they were just placeholder values and items weren't fully implemented in the backend either, but the frontend needs them for now so we're going to restore this chance


agustinesco (approved):
LGTM! items are being added correctly to the db 👍

ncontinanza (approved):
LGTM!

PR #392, opened 2024-03-20
Add range to dash config

BertovDev:
Add range to dash config

In order to draw the indicator of the skill we should have this field in the dash skill.


apps/arena/priv/config.json
```diff
@@ -244,7 +244,8 @@
         {
           "dash": {
             "speed": 60.0,
-            "duration": 500
+            "duration": 500,
+            "range": 500
```
manucamejo, line 248: What is the idea here? Are we expecting the character and the indicator to match where they end?
Because with a little change in the speed, this value gets outdated. Maybe we should calculate the range using the duration and the speed, also the front end can do that.

@saugon what do you think?

manucamejo (changes requested):

apps/arena/priv/config.json
saugon: Yes, it should be calculated. How do you think we should decide when to calculate it? If the value doesn't exist?

agustinesco:
This will be fixed sending the mechanics config and handling the logic in the client

PR #395, opened 2024-03-21
Create Europe Arena release workflow

lotuuu:
Do not merge label added until this PR is reviewed by @klaus993. Other reviews in the meantime are encouraged, though! Just don't merge if it gets enough approvals :P

This PR introduces the workflow to be used after #388. The idea is to merge this first to check if the deploy with releases (branch `configure-releases`) effectively works when ran through the workflow (already checked manually).

Tu run the workflow we might need to do some configuration on the GH repo for the new RELEASE en variable


Nico-Sanchez (approved):

.github/workflows/testing-europe-deploy-arena-release.yml
```diff
@@ -0,0 +1,77 @@
...
+          chmod 400 ~/.ssh/id_ed25519
+          retries=5; until ssh-keyscan $TS_HOST >> ~/.ssh/known_hosts || [ $retries -eq 0 ]; do ((retries--)); sleep 5; done
+
+      - name: Copy deploy script
+        env:
+          SSH_USERNAME: ${{ secrets.SSH_USERNAME }}
+          SSH_HOST: ${{ vars.TS_EUROPE_HOST }}
+        run: |
+          set -ex
+          rsync -avz --mkpath /devops/deploy.sh ${SSH_USERNAME}@${SSH_HOST}:/home/${SSH_USERNAME}/deploy-script/
```
samoht9277, line 50:
```suggestion
          rsync -avz --mkpath devops/deploy.sh ${SSH_USERNAME}@${SSH_HOST}:/home/${SSH_USERNAME}/deploy-script/
```
the `deploy.sh` path shall not be absolute!

Nico-Sanchez: Devops sight ❇️

samoht9277 (approved):
Looks great! 

We could potentially improve this in the future by passing the location via `inputs`, so we only have a single deploy workflow, where you have to select the `release` and the `location`.

.github/workflows/testing-europe-deploy-arena-release.yml
```diff
@@ -0,0 +1,77 @@
...
+          chmod 400 ~/.ssh/id_ed25519
+          retries=5; until ssh-keyscan $TS_HOST >> ~/.ssh/known_hosts || [ $retries -eq 0 ]; do ((retries--)); sleep 5; done
+
+      - name: Copy deploy script
+        env:
+          SSH_USERNAME: ${{ secrets.SSH_USERNAME }}
+          SSH_HOST: ${{ vars.TS_EUROPE_HOST }}
+        run: |
+          set -ex
+          rsync -avz --mkpath devops/deploy.sh ${SSH_USERNAME}@${SSH_HOST}:/home/${SSH_USERNAME}/deploy-script/
```
klaus993, line 50: `devops/deploy.sh` path doesn't exist

.github/workflows/testing-europe-deploy-arena-release.yml
```diff
@@ -0,0 +1,77 @@
...
+                BRANCH_NAME=${BRANCH_NAME} \
+                MIX_ENV=${MIX_ENV} \
+                RELEASE=${RELEASE} \
+                PHX_SERVER=${PHX_SERVER} \
+                PHX_HOST=${PHX_HOST} \
+                PORT=${PORT} \
+                SECRET_KEY_BASE=${SECRET_KEY_BASE} \
+                NEWRELIC_APP_NAME=${NEWRELIC_APP_NAME} \
+                NEWRELIC_KEY=${NEWRELIC_KEY} \
+                /home/${SSH_USERNAME}/deploy-script/deploy.sh
```
klaus993, line 77: won't work because deploy script doesn't exist

klaus993 (changes requested):
This will need changes after it's merged, so why not make them now?
Or else, target the other branch, `configure-releasees`, instead of `main`

.github/workflows/testing-europe-deploy-arena-release.yml
lotuuu: It does exist in `configure-releases` branch! The idea is to run this workflow on that branch to check if the new releases work. If we merge it all together we may break main, and then none of our workflows would work

klaus993 (approved):

Nico-Sanchez (approved):

PR #398, opened 2024-03-21
22/03 balance pass

marco-paulucci:
### BALANCE INTENT
- The goal of this balance iteration is to make the basic skills of characters (+ dash) actually useful so players don’t rely on only using their ultimate. If a character feels useless in this regard please report it.
- Ultimates are planned to work more as finishers or strong openers of situations. Also, it's expected their charging mechanism will be changed from cooldown into "mana"-like system where you charge it by dealing damage or receiving damage, depending on the character, to promote engagement from players.
- The above change is also one of the big reasons we need to make the basic+dash kit "good enough" on its own while people charge their ult. 
- In the meantime ults will have a cooldown of 8-9s (or we'll adjust as necessary)

The objectives, in specific order:
1) Make sure things work as expected and the game is responsive (current priority), i.e. what you expect to happen happens
2) Balance the numbers to try make the characters equally competitive
3) Make the game fun and find outlier skills or kits that are not fun to play and improve or replace them for more interesting stuff

### THINGS THAT WILL MAKE THE GAME MORE FUN

There are two incoming features (not included here) that I think will make the game more enjoyable in the near future:

#### 1) Breakable power-up crates.
- They will allow more interaction with the map
- Make your character stronger
- Create confrontation points / points of interest

#### 2) Bots
- Bots are meant to be accessible kills, which should make the game more entertaining. If in a given game there are 3 real players and 7 bots, the likely win rate for a player jumps from 10% to 33%. And winning is FUN

### DISCLAIMER
- This is a blind pass because there are no bots to test against. Will continue iterating once its deployed.

.
.
.

## Balance Changes

Items
- Boots: reduced speed boost to something more reasonable
- Mirra’s Blessing: disabled (duration set to 1ms). Until there is no proper VFX this will only hinder playtesting (you cant tell if you didn’t hit someone if you should, because maybe it was invincible)

Base Chars
- Made base speeds slightly less symmetric across chars. 
- Increased Muflus' HP to compensate for slightly slower speed

Skills: Muflus
- Basic
    - Now uses a circle to calculate the collision, should work more accurately
    - Damage was increased
    - Fastened execution and animation time.
- Leap
    - Increased targeting radius
    - Slightly increased AoE radius and damage accounting for longer cooldown
- Dash
    - Will have a low-ish cooldown so it works well with his basic
    - Slightly tweaked params

Skills: H4ck
- Slingshot:
    - Slightly increased spread and projectile size
    - Increased projectile speed
    - Increased execution speed
- Denial of Service
    - Reduced execution time so that some of the bullets you can control after your main animation ran out (it’s how the skill was meant to work, but due to animation issues this is a middleground test)
    - Increased projectile speed

Skills: Uma
- Avenge
    - Faster execution. VFX is a bit broken because I couldn’t adjust it to faster speed.
    - Reduced range
    - Increased how much she moves forward
    - Now uses circles to calculate collisions, should feel more accurate
- Veil of Radiance
    - Increased damage as to not make the invisibility the only relevant aspect of the skill
    - Slightly faster execution

Misc changes
-  added prefixes to skill names with char
-  separated muflus and h4ck dash (characters sharing skills is not ideal)
-  removed some fields in cases where they would not have an effect (cd skills w/ stami


marco-paulucci:
ready for review

PR #401, opened 2024-03-22
[gh-400] Change muflus hit to circle hit

### LINKED ISSUE #400: Replace muflus cone hit ability with a circle hit
istead of a cone it should be a circle hit with an offset to match his front end representation

agustinesco:
Closes #400 

Replaced muflus "hit" skill with a circle hit instead of a cone hit

Added offset to the circle generated with the circle hit mechanic same as the projectile spawning

Hit that damaged the enemy
Screenshot 2024-03-22 at 11 37 03

Missed hit
Screenshot 2024-03-22 at 11 38 17


agustinesco:
This pr was merged to #398

PR #402, opened 2024-03-22
Fix browser links

agustinesco:
The browser links from [This link](http://localhost:3000/) were broken since the implementation of the name


Nico-Sanchez (changes requested):
The links are no longer broken but I only see 1 player when instantiating more than one (using multiple tabs).

Nico-Sanchez (approved):

ncontinanza (approved):
LGTM!

PR #413, opened 2024-03-25
Skills config

ncontinanza:
Part of [Config](https://github.com/lambdaclass/champions_of_mirra/issues/251)
Closes [Add Skills JSON import](https://github.com/lambdaclass/champions_of_mirra/issues/276)

Skills are now configurable in the skills.json file, instead of being defined in the seeds. As with the characters, the configuration file is parsed in the seeds when starting the app, and the skills are upserted.

To test this PR you can edit the skills.json file and start the app. To check that the changes were applied you can directly query the Skills table in the db, or run the following in your terminal:

```
make start
GameBackend.Units.Skills.get_skill_by_name(skill_name)

```
Using the name of the skill you want to get in the second command should retrieve the Skill, were you can check that the value of each field corresponds to the ones you set in the configuration file.


lotuuu (changes requested):
We have an unused alias in seeds

```
    warning: unused alias Characters
    │
  8 │ alias GameBackend.Units.Characters
    │ ~
    │
    └─ priv/repo/seeds.exs:8:1
```

lotuuu (approved):

PR #417, opened 2024-03-25
[GH 414 add bush mechanic

agustinesco:


apps/game_client/lib/game_client/protobuf/messages.pb.ex
```diff
@@ -397,6 +408,8 @@ defmodule GameClient.Protobuf.Player do
   field(:effects, 10, repeated: true, type: GameClient.Protobuf.Effect)
   field(:inventory, 11, type: GameClient.Protobuf.Item)
   field(:cooldowns, 12, repeated: true, type: GameClient.Protobuf.Player.CooldownsEntry, map: true)
+  field(:visible_players, 13, repeated: true, type: :uint64, json_name: "visiblePlayers")
```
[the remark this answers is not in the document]
agustinesco: This is something that will benefit us in the future, now we're sending every player to the clients and that's really easy to cheat. In the short term the idea is to use this list to send the players in the update already filtered so it's not easily cheatable. I think we should keep this imo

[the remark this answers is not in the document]
lotuuu: Makes sense!

apps/arena/lib/arena/game_updater.ex
```diff
@@ -1210,6 +1230,41 @@ defmodule Arena.GameUpdater do
...
+
+          players_in_same_bush? =
+            Enum.any?(bush_collisions, fn collided_id -> collided_id in candidate_bush_collisions end)
+
+          players_close_enough? =
+            Physics.distance_between_entities(player, candidate_player) <= game_config.bushes.field_of_view_inside_bush
+
+          if candidate_player.id != player_id and
+               (Enum.empty?(candidate_bush_collisions) or (players_in_same_bush? and players_close_enough?)) do
+            acc ++ [candicandidate_player_id]
```
lotuuu, line 1255: I think there's no reason not to use `[candidate_player_id | acc]` here, right?

It's the better option out of the two, [as per ++ docs](https://hexdocs.pm/elixir/1.12/Kernel.html#++/2), and I don't think we care about what order these players are on

lotuuu (changes requested):
Functionality looked good by following the steps you gave! Leaving a small code comment

lotuuu:
Also please add the steps for testing the PR that you sent me via slack in the description!

apps/arena/lib/arena/game_updater.ex
agustinesco: nope, not reason at all. Changed!

Also i noticed we're doing extra checks since we're running the checks for the same player, i moved the ifs a little bit so we dont do extra calculations

lotuuu (approved):

apps/serialization/messages.proto
```diff
@@ -209,6 +213,9 @@ message Pool {
   uint64 owner_id = 1;
 }
 
+message Bush {
```
tvillegas98, line 216: Why do we need this empty message struct? is it for the addional_info oneof? 🤔

agustinesco: yeah! exactly that

PR #425, opened 2024-03-26
Effect refactor: Reset and reapply player effects on each tick (and some other cool things)

AminArria:
## Current problem

Effects are a mess, we have some skills directly modifying attributes of a player and then having their own (repeated) special logic to cleanup their changes. At the same time we have special handling for attributes that could be modified by things, for example power ups, we have the magical `Player.calculate_real_damage/2` to apply damage modifiers

Even worse, which is sort of happening, their is the annoying problem around "undoing" effects, once the effects runs out and gets removed how do we properly undo it. Think the following scenario: `effect_1` increases attribute by 30, `effect_2` increases it by 20, and `effect_3` increases it by 10%, in this scenario the final increase is 55 (if applied as they show up), but what happens if `effect_2` is removed? Should the new value be 35 or 33? This is only one simple example, but undoing things when we have percentage increases can get messy real quick

## Goal

Have a simple and straightforward way to manage effects that modify attributes, if you are setting an effect you should not care about how to undo it, this should be automagically handled (less bugs)

## Proposed solution

The idea here is to have the effects reapplied on every game update. On each we reset every player to their base stats (or whatever is appropiate as a base) and then reapply all effects

By doing it like this we have no need to "undo" effects as they will not be there to apply

Some caveats that will exists
- Effects will not be applied until next tick (they will be in the player, just not applied)
- Effects will last one more tick than expected (could be fixed by properly ordering in the sequence)

Pending to look into
- [x] Effects list in players is a map right now and that will mess up the ordering between applications, we would need either to use and ordered map or a list 

## Actual changes

This is me trying to summarise the changes made in this PR

- Add logic to reset and reapply stat based effects on each tick
- Refactor items to use effects, before the changes on each was hardcoded
- Effects are removed as part of game tick rather than by messages sent when putting them
- Make `invisible` a proper effect mechanic rather than relaying on effect name
- Effects that need to be remove after an action (e.g. invisible) are removed on game tick
- Refactor how pool add/remove effects so it goes through `Effect` module
- Refactor `Player.effects` to be a list instead of a map (**IMPORTANT**)

## Testing

For testing you will need to use this client PR https://github.com/lambdaclass/curse_of_mirra/pull/1660


apps/arena/lib/arena/game/player.ex
```diff
@@ -261,27 +262,59 @@ defmodule Arena.Game.Player do
...
+    put_in(player, [:aditional_info, :effects], effects)
+  end
+
+  def reset_effects(player, game_config) do
+    character = Enum.find(game_config.characters, fn %{name: name} -> name == player.aditional_info.character_name end)
+
+    player =
+      player
+      |> put_in([:speed], character.base_speed)
+      |> put_in([:aditional_info, :stamina_interval], character.stamina_interval)
```
agustinesco, line 313: maybe we could avoid this character fetch if we add the base_stats to the player and have them separated from the altereded ones, e.g. having base some variable called :base_speed and to reset it you just do `|> put_in([:speed], player.aditional_info.base_speed)`

apps/arena/lib/arena/game/skill.ex
```diff
@@ -330,22 +317,12 @@ defmodule Arena.Game.Skill do
...
-            :aditional_info,
-            :effects,
-            effect.id,
-            :effect_mechanics,
-            mechanic_name,
-            :last_application_time
-          ],
-          now
-        )
+        |> Effect.put_in_effect(effect, [:effect_mechanics, mechanic_name, :last_application_time], now)
```
agustinesco, line 325: what would happend if we need to apply two effect with the same name at the same time? for example when someone get inside two valtimer's pools, the name will be the same but we should do the damage in two different ocurrences

apps/arena/lib/arena/game/effect.ex
```diff
@@ -0,0 +1,103 @@
+defmodule Arena.Game.Effect do
+  @moduledoc """
+  This module contains all the functionality related to effects
+  """
+
+  def put_effect(game_state, player_id, owner_id, effect) do
+    last_id = game_state.last_id + 1
+    expires_at = System.monotonic_time(:millisecond) + effect.duration_ms
```
agustinesco, line 8: not every effect have expiration time, for example the effect that the pool applies doesn't, they're added/deleted when a player collides or stop colliding with a pool, in fact when you try to ult with valtimer this pr breaks because it doesn't have a duration:

```
** (KeyError) key :duration_ms not found in: %{
  name: "singularity",
  remove_on_action: false,
  effect_mechanics: %{
    pull: %{force: 1.0, effect_delay_ms: 0},
    damage: %{damage: 0.01, effect_delay_ms: 400}
  }
}
```

agustinesco:
The uma invisibility is no longer working as expected:

https://github.com/lambdaclass/mirra_backend/assets/106101218/a3f18129-2727-4a11-b8d9-758132324644

apps/arena/lib/arena/game/skill.ex
AminArria: There are no restrictions in the general case, but there is `Effect.put_non_owner_stackable_effect/4` which does check for something like that, it checks owner and effect name, so to answer you it will depend on how we are putting the effects on the player and if we are talking same owner or different owners

apps/arena/lib/arena/game/effect.ex
AminArria: Fixed on https://github.com/lambdaclass/mirra_backend/pull/425/commits/c57e497b3d9d1be691023f3b7c325bc0896d92f5

apps/arena/lib/arena/game/player.ex
AminArria: Changed in https://github.com/lambdaclass/mirra_backend/pull/425/commits/5fc4e9270b784b3237f16c2d601cf18b978b1e10

agustinesco (changes requested):
The valtimer tp is broken 
`lib/arena/game_updater.ex:104: Arena.GameUpdater.handle_info({:remove_damage_immunity, 1},`

for some reason the dash is not working as expected, the leap is also wrong

https://github.com/lambdaclass/mirra_backend/assets/106101218/7295b2bb-886c-4895-9ca8-86afb425fd27

apps/arena/lib/arena/game/skill.ex
agustinesco: nice!

apps/arena/lib/arena/entities.ex
```diff
@@ -39,8 +41,9 @@ defmodule Arena.Entities do
         power_up_damage_modifier: config.power_ups.power_up.power_up_damage_modifier,
         inventory: nil,
         damage_immunity: false,
-        effects: %{},
-        cooldowns: %{}
+        effects: [],
+        cooldowns: %{},
+        bonus_damage: 0
```
agustinesco, line 46: just to comment this but this variable sounds like an additive modifier, maybe `bonus_damage_multiplier` would be more suitable

agustinesco (approved):
LGTM!

ncontinanza (approved):
LGTM! Added the do not merge tag so we can wait for the front-end PR to be ready to merge

PR #427, opened 2024-03-27
[Rewrite battle] Battle tests

ncontinanza:

Our battles now have a lot of functionality, so some tests are needed to ensure the correct behavior of these mechanics, at least in a vaccum


agustinesco (approved):

Nico-Sanchez (changes requested):
Looking good!
We could add some utils module or auxiliary functions to create the mocked schemas, we're defining several times a basic attack or a character that are almost the same but for 1 or 2 fields.

apps/champions/mix.exs
```diff
@@ -23,6 +24,10 @@ defmodule Champions.MixProject do
     ]
   end
 
+  # Specifies which paths to compile per environment.
+  defp elixirc_paths(:test), do: ["lib", "test/support"]
+  defp elixirc_paths(_), do: ["lib"]
```
Nico-Sanchez, line 29: Nice! 💎

Nico-Sanchez (approved):

lotuuu (approved):

PR #428, opened 2024-03-27
Fix deployment bug

Nico-Sanchez:
## Motivation
Deployment was being executed successfully but the service kept restaring because the configurator app was being launched alongside arena (we don't want this).

## Changes
Remove configurator as arena dependency

### DIFF
--- a/apps/arena/mix.exs
+++ b/apps/arena/mix.exs
@@ -51,8 +51,9 @@ defmodule Arena.MixProject do
       {:exbase58, "~> 1.0.2"},
       {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
       {:plug_cowboy, "~> 2.5"},
-      {:toxiproxy_ex, "~> 1.1.1"},
-      {:configurator, in_umbrella: true}
+      {:toxiproxy_ex, "~> 1.1.1"}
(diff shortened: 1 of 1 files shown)


apps/arena/mix.exs
```diff
@@ -51,8 +51,9 @@ defmodule Arena.MixProject do
       {:exbase58, "~> 1.0.2"},
       {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
       {:plug_cowboy, "~> 2.5"},
-      {:toxiproxy_ex, "~> 1.1.1"},
-      {:configurator, in_umbrella: true}
+      {:toxiproxy_ex, "~> 1.1.1"}
+      # TODO We will find a way to solve this dependency.
```
lotuuu, line 55: As was stated in slack, we should delete the dependency altogether instead of adding this TODO

lotuuu (changes requested):

apps/arena/mix.exs
Nico-Sanchez: Yep, Done!

lotuuu (approved):

agustinesco (approved):
🔪

PR #430, opened 2024-03-27
Move advance_level to Champions app and add experience reward

lotuuu:
Fixes https://github.com/lambdaclass/champions_of_mirra/issues/216

Adds experience rewads to levels. These are applied automatically to a user on level win.

We had to move advance_level from GameBackend to Champions since we needed a Champions function (add_experience) to make it work, which means that we were putting Champions logic inside GameBackend.

# To Test

Start the client so that the test user gets created, then run in iex:
`{:ok, user} = Champions.Users.get_user_by_username("testUser")`
Check your level and xp by running:
`Champions.Users.get_user("d53d6129-d69f-4762-8077-9d84ab25b13b") |> elem(1) |> Map.take([:experience, :level])`
Now win a level in the campaign and then check your xp again. It should have grown.

### DIFF
--- a/apps/champions/lib/champions/battle.ex
+++ b/apps/champions/lib/champions/battle.ex
@@ -23,7 +23,7 @@ defmodule Champions.Battle do
       units = Units.get_selected_units(user_id)
 
       if battle(units, level.units) == :team_1 do
-        case Users.advance_level(user_id, level.campaign_id) do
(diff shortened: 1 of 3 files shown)


ncontinanza (approved):
LGTM!

agustinesco (approved):
LGTM!

PR #432, opened 2024-04-03
Improve loadtest app

Nico-Sanchez:
## Motivation

Loadtest app 1st iteration was working ok, but here I give it some love to it.

## Changes

- [Terminate websocket when backend tells me that the game started](https://github.com/lambdaclass/mirra_backend/commit/88f95411efe4b4791cccc412babc5dbb7afab2ac)
- [Make ws messages less verbose](https://github.com/lambdaclass/mirra_backend/commit/edd407aec7cfbae2c250265f656f28fc75588b8d)
- [Create ets table to persist and log the amount of connected clients](https://github.com/lambdaclass/mirra_backend/commit/45bcdd96c7ba0e30a6c726f4f7d75d73c68970f3)
- [Add missing env vars in deploy script for loadtest app](https://github.com/lambdaclass/mirra_backend/commit/9c9cb560f0ea832460ccef198f2fbdf636f7c8e6)


ncontinanza (approved):
LGTM!

apps/arena_load_test/lib/arena_load_test/game_socket_handler.ex
```diff
@@ -83,6 +86,19 @@ defmodule ArenaLoadTest.GameSocketHandler do
...
   end
 
+  def terminate(_, state) do
+    case :ets.lookup(:clients, state.client_id) do
+      [{client_id, _}] ->
+        :ets.delete(:clients, client_id)
+
+      [] ->
+        raise KeyError, message: "Client with ID #{state.client_id} doesn't exist."
+    end
```
agustinesco, line 96: Why not calling the `:ets.delete(:clients, client_id)` directly? do we want to stop everything when we do not find the client in the :clients table due to this custom rise?

apps/arena_load_test/lib/arena_load_test/socket_supervisor.ex
```diff
@@ -31,9 +32,17 @@ defmodule ArenaLoadTest.SocketSupervisor do
 
   # Creates `num_clients` clients to join a game
   def spawn_players(num_clients) do
-    for i <- 1..num_clients do
-      {:ok, _pid} = add_new_client(i)
+    case :ets.whereis(:clients) do
```
agustinesco, line 35: this is the same as an if statement 
```
if :ets.whereis(:clients) == :undefinded do
 :ets.new(:clients, [:set, :named_table, :public])
end
```

apps/arena_load_test/lib/arena_load_test/socket_supervisor.ex
```diff
@@ -31,9 +32,17 @@ defmodule ArenaLoadTest.SocketSupervisor do
...
+      :undefined -> :ets.new(:clients, [:set, :named_table, :public])
+      _table_exists_already -> nil
     end
+
+    Enum.each(1..num_clients, fn client_number ->
+      Logger.info("Iteration: #{client_number}")
+      {:ok, _pid} = add_new_client(client_number)
+      true = :ets.insert(:clients, {client_number, "1"})
```
agustinesco, line 43: you could also save the pid in the table, just to replace the magic 1

agustinesco (approved):
🪨

apps/arena_load_test/lib/arena_load_test/socket_supervisor.ex
Nico-Sanchez: Yeah, I just wanted to make more explicit that when is not undefined is because there's no table called `:clients`

apps/arena_load_test/lib/arena_load_test/socket_supervisor.ex
Nico-Sanchez: You're right. However, I don't really need any value, I just want to store client_id's and know how many clients are in the table.

PR #438, opened 2024-04-03
Champions skills config

ncontinanza:
Closes [Config](https://github.com/lambdaclass/champions_of_mirra/issues/251)

This PR implements the basic and ultimate skill of each character following the specification found in [this design document](https://docs.google.com/spreadsheets/d/177mvJS75LecaAEpyYotQEcrmhGJWI424UnkE2JHLmyY/edit?usp=sharing).

You'll notice that the targeting strategies are hardcoded to be always the random strategy. That'll remain like that until the  ([[Targeting]](https://github.com/lambdaclass/champions_of_mirra/issues/325#top)) issue is solved.

To test it, you can start the application and run:
```
{:ok, user} = Champions.Users.register("Nico4")
GameBackend.Repo.preload(user, units: [character: [:basic_skill, :ultimate_skill]])
```
There you can check that all your units have a basic and an ultimate id


apps/game_backend/lib/game_backend/units/skills/effects/target_strategy.ex
```diff
@@ -22,11 +27,13 @@ defmodule GameBackend.Units.Skills.Effects.TargetStrategy do
 
   def dump(strategy), do: {:ok, target_strategy_to_string(strategy)}
 
+  defp target_strategy_to_string("all"), do: "random"
```
lotuuu, line 30: Rather than storing it wrongly on the DB, we should keep this as it was and change `create_effect_map` in `Simulator` to always set target_strategy to "random".

Same design decision should be followed by other fallbacks

apps/champions/test/support/test_utils.ex
```diff
@@ -75,10 +75,10 @@ defmodule Champions.TestUtils do
       animation_trigger: 0,
       effects: [
         %{
-          type: "instant",
+          type: %{type: "instant"},
           initial_delay: 0,
           components: [],
-          modifier: [],
+          modifiers: [],
```
lotuuu, line 81: If this isn't absolutely necessary, restore it so that it will generate less conflicts against #464

apps/champions/priv/skills.json
```diff
@@ -1,12 +1,1651 @@
...
+  {
+    "name": "Valcher Basic",
+    "energy_regen": 100,
+    "animation_duration": 0,
+    "animation_trigger": 0,
+    "effects": [
+      {
+        "type": {
+   "type": "instant"
+},
```
lotuuu, line 126: Format these entries!

lotuuu: Just running the whole skills.json file through a json formatter should do the trick

apps/champions/lib/champions/config.ex
```diff
@@ -42,9 +53,17 @@ defmodule Champions.Config do
...
         game_id: Utils.game_id(),
+        basic_skill_id: get_skill(basic_skill).id,
+        ultimate_skill_id: get_skill(ultimate_skill).id,
         active: true
       }
     end)
     |> Characters.upsert_characters()
   end
+
+  defp get_skill(skill_name) do
```
lotuuu, line 64: Why do we need this function? Just call Skills.get_skill_by_name(skill_name).
If you're afraid that a Skill with `skill_name` may not exist:
a. you're still not handling it since you don't have an else clause in your with
b. you can just use a default value `Map.get(basic_skill_that_could_be_nil, :id, default_value)`

lotuuu (changes requested):

apps/champions/priv/skills.json
```diff
@@ -1,12 +1,1651 @@
...
+            "type": "DealDamage",
+            "attack_ratio": 1,
+            "energy_recharge": 50,
+            "delay": 0
+          }
+        ],
+        "target_strategy": "nearest",
+        "target_count": 1,
+        "target_allies": false,
+        "target_attribute": "Health"
```
lotuuu, line 368: target_attribute like this one are wrong. "Nearest" strategy should not take a `target_attribute`. Only Min and Max.

lotuuu: Also, isn't target_attribute as a whole outdated? We already have this
```
  def cast(%{"lowest" => attribute}), do: {:ok, {"lowest", attribute}}
  def cast(%{"highest" => attribute}), do: {:ok, {"highest", attribute}}

in our TargetStratgy type
```

lotuuu (changes requested):

apps/champions/lib/champions/config.ex
ncontinanza: Right, didn't consider that.
In any case, if the skill_name doesn't exist Skills.get_skill_by_name/1 will return nil, which I guess is ok if we don't have a skill in the json for that character.

apps/champions/lib/champions/config.ex
```diff
@@ -53,17 +53,11 @@ defmodule Champions.Config do
         base_health: Integer.parse(health) |> elem(0),
         base_defense: Integer.parse(defense) |> elem(0),
         game_id: Utils.game_id(),
-        basic_skill_id: get_skill(basic_skill).id,
-        ultimate_skill_id: get_skill(ultimate_skill).id,
+        basic_skill_id: Skills.get_skill_by_name(basic_skill).id,
```
lotuuu, line 56: use the `[:id]` syntax so that it won't fail if the skill doesn't exist
```
iex(2)> nil[:a]
nil
```

lotuuu (changes requested):

apps/champions/priv/skills.json
```diff
@@ -1,12 +1,1655 @@
...
+      {
+        "type": {
+          "duration": 6,
+          "period": 1
+        },
+        "initial_delay": 0,
+        "components": [],
+        "modifiers": [
+          {
+            "attribute": "damage reduction",
```
lotuuu, line 1233: use an underscore instead of a space

apps/champions/priv/skills.json
```diff
@@ -20,35 +1663,339 @@
...
+    "animation_trigger": 0,
+    "effects": [
+      {
+        "type": {
+          "duration": 2,
+          "period": 1
+        },
+        "initial_delay": 0,
+        "components": [],
+        "modifiers": [
```
lotuuu, line 1887: Let's not handle paralyze-stun effects via speed, we should create a tag for this instead

apps/champions/priv/skills.json
```diff
@@ -20,35 +1663,339 @@
...
+            "type": "DealDamage",
+            "attack_ratio": 0.8,
+            "energy_recharge": 50,
+            "delay": 0
+          },
+          {
+            "type": "DealDamage",
+            "attack_ratio": 0.8,
+            "energy_recharge": 50,
+            "delay": 0
```
lotuuu, line 1933: We can ask for the correct value later, but we should set these delays to increasing numbers like 0, 1 to correctly emulate the spirit of the skill

apps/champions/priv/skills.json
```diff
@@ -20,35 +1663,339 @@
...
+          {
+            "type": "DealDamage",
+            "attack_ratio": 0.59,
+            "energy_recharge": 50,
+            "delay": 0
+          },
+          {
+            "type": "DealDamage",
+            "attack_ratio": 0.83,
+            "energy_recharge": 50,
```
lotuuu, line 1972: We can ask for the correct value later, but we should set these delays to increasing numbers like 0, 1, 2 to correctly emulate the spirit of the skill

lotuuu (changes requested):

apps/champions/priv/skills.json
lotuuu: You did it correctly with Muflus!

apps/champions/priv/skills.json
ncontinanza: Yeah, that's something to be defined yet, I left them in 0 so they don't interfere with what we're currently testing

lotuuu: What I mean is that we can leave some default "1, 2, 3" values, we can change fix them again later

lotuuu: They won't interfere with anything because delay is ignored for DealDamage executions

lotuuu (approved):

lotuuu (approved):

agustinesco (approved):

PR #442, opened 2024-04-04
Replays for Champions Battle

lotuuu:
Closes https://github.com/lambdaclass/champions_of_mirra/issues/317
First commit of this PR's changes is [Implement history in the simulator](https://github.com/lambdaclass/mirra_backend/pull/442/commits/1da73222ff0b3c046b9bb9070b9e3c447c58a676)
Implements battle log functionality. The idea is for this to be able to be replayed (animated) by the client.

# Battle

Adds another accumulator to the battle called `history`, where we store battle history according to the format specified as required by the client for animation. We constantly pass it around to update it when one of the actions we want to log takes place. When this is returned to the champions_socket_handler.ex we make the translation into Protobuf structs for encoding.

# Protobuf

The protobuf message for battle replays has been created together with @tkz00 and @ncontinanza. A more in-depth explanation can be found here: https://www.notion.so/lambdaclass/Notes-on-animation-messages-d826b0f0ca864e77ac8a563723e94472?pvs=4.

We also make steps 0-indexed since it makes more sense: when you start a chronometer, it starts from 0 to 1, not from 1 to 2.

# To test

As with other things that require protobuf testing, you can use the `apps/gateway/test/support/socket_tester.ex` file. You will need to move it under `apps/gateway/lib/gateway` to be able to use it in iex. Then, you can run:
```
IEx.configure(inspect: [limit: :infinity])
{:ok, pid} = Gateway.SocketTester.start_link
Gateway.SocketTester.battle_test(pid, nil)
send(pid, {:last_message, self()})
flush()
```

This will get you the response that would be sent to the client

If you want to test modifier application/expiration actions you can change muflus' ultimate_skill in the seeds to:
```
ultimate_skill: %{
      name: "Muflus Ultimate",
      effects: [
        %{
          type: %{"period" => 0, "duration" => 10},
          initial_delay: 0,
          components: [],
          modifiers: [
            %{
              attribute: "damage",
              modifier_operation: "Multiply",
              float_magnitude: 0.75
            }
          ],
          executions: [],
          target_strategy: "random",
          target_count: 2,
          target_allies: false,
          target_attribute: "Health"
        }
      ]
    }
```


apps/serialization/gateway.proto
```diff
@@ -291,3 +297,95 @@ syntax = "proto3";
...
+    EFFECT_TRIGGER = 1;
+    EFFECT_HIT = 2;
+    EFFECT_MISS = 3;
+  }
+
+  enum Stat {
+    health = 0;
+    energy = 1;
+    damage = 2;
+    defense = 3;
```
ncontinanza, line 314: Is there any reason to have these in lower case and the values of SkillActionType in uppercase?

ncontinanza (dismissed):
LGTM!

ncontinanza (approved):

apps/champions/test/battle_test.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Champions.Test.BattleTest do
+defmodule Champions.Test.Battle do
```
agustinesco, line 1:
```suggestion
defmodule Champions.Test.BattleTest do
```

ncontinanza (approved):

agustinesco (approved):

PR #447, opened 2024-04-08
Add bot missing env vars

agustinesco:
The bot manager application was unable to start since we're using the `BOT_MANAGER_PORT` env var to start the http process, and we forgot to parse this value to integer that's required by the Plug library

This pr will fix 2 more things:
- The arena server will send his url in the get request to the bot manager
- Now you can turn on and off the bot spawning

You can check if this pr actually fixed the bots starting a match in brazil and testing that the bots are moving.
I'll add examples of the deploys runs:
- [Arena deploy](https://github.com/lambdaclass/mirra_backend/actions/runs/8605019476)
- [Load Tests deploy](https://github.com/lambdaclass/mirra_backend/actions/runs/8605099183)


lotuuu (approved):

apps/bot_manager/lib/application.ex
```diff
@@ -7,21 +7,23 @@ defmodule BotManager.Application do
 
   @impl true
   def start(_type, _args) do
-    http_server =
-      Plug.Cowboy.child_spec(
-        scheme: :http,
-        plug: BotManager.Endpoint,
-        options: Application.get_env(:bot_manager, :end_point)[:port]
-      )
+    options = [port: get_bot_manager_port()]
```
AminArria, line 10: Don't get the reasoning for this instead of configs, can you explain a bit?

agustinesco: basically i didn't understood the benefit of having the configs there instead of just creating them here, i moved this to the runtime.ex file and it worked 😄 i just moved the whole endpoint config there so we centralize that

lotuuu (approved):

AminArria (approved):

PR #448, opened 2024-04-08
Items modifiers

ncontinanza:
Closes https://github.com/lambdaclass/champions_of_mirra/issues/249

Use the already existing modifiers to implement the items stats bonuses. 
We used the Decimal library to handle the operation with floating point numbers, as the way in which the floats are represented was leading us to problems when truncating and rounding at the moment of calculating a unit's attack, defense and health. This affected the pre-existing implementation.

To test, you can equip the existing items in the seeds (or create new ones!) and check the character's stats before and after, to see how they are modified. After that, unequip the item and verify that the affected stats go back to normal.

The steps you can follow to do so are:

1. Run `{:ok, user} = Champions.Users.register("User")` to create a new user.
2. Pick a unit from the units list, and check its current attack by running `user.units |> hd() |> Champions.Units.get_attack()` (or any other stat, like defense or health)
3. Pick an item from the items list that affects the attack (or whatever stat you chose for this test).
4. Run `Champions.Items.equip_item(user.id, item_id, unit_id)` to equip the item you chose.
5. Run `{:ok, updated_user} = Champions.Users.get_user(user.id)` to get the user with the updated units, and look for the unit that equipped the item. Here you should see the item equipped to the unit, and that the attack was modified by the item by running `updated_user.units |> hd() |> GameBackend.Repo.preload(items: :template) |> Champions.Units.get_attack()`

You can also check the tests in this PR and play with the values and assertions to go further on this functionality.
Another useful test to perform is to start a game with the client, equip an item to an unit, and fight a battle using that unit. The battle should work properly, and you should see the items bonus in the server logs.


lotuuu:
Please mention a bit about the decision to replace Float operations with Decimals in the description!

lotuuu (changes requested):
Let's implement something similar to what we have with Modifiers, but with a custom Item.BaseModifier, which ItemTemplates have an array of. These will possess the base_modifier value for each skill, as well as the operation to apply to the user with the calculated value resulting of `F(base_value)` where `F` is the formula for item scaling (for now, we can just make it scale linearly)

martinacantaro:
Ran the commands in the description and it worked as intended.
Attack of the first unit increased from 113 to 180 as expected with the 1.6 multiplier of the item.
attack 113
Also ran the test suite, although that is checked with each commit automatically, because they include a test with addition as the "modifier_operation"

martinacantaro (approved):

martinacantaro (approved):
tested it in battle and the sword item increases the attack power

lotuuu (changes requested):

lotuuu (approved):

PR #449, opened 2024-04-08
Removed RELEASE workflow input on GitHub Actions.

samoht9277:
This PR aims to facilitate the deploying sequence on each individual server.

By removing the abiity to choose what release to deploy, there is no possible confusion while deploying a certain release, since this PR no longer supports the ability to choose the deploy.

```
Arena    Server        -> Only arena
Loadtest Server        -> Only arena
Loadtest Client        -> Loadest and/or bots
```


devops/deploy.sh
samoht9277, line 1: By removing this, we no longer have a way of confirming if the release we are deploying is the same as the already deployed (if any).

Since we no longer have the option to choose which release we want, these lines are no longer in need.

However, the `loadtest client` servers still have the choice to select between `loadtest` and `bots manager`. In this case, the script would generate two different services that run concurrently.

Nico-Sanchez (approved):

lotuuu (approved):
Changes look good. As we talked, deploy is failing because of server issues and not because of the code modifications

Nico-Sanchez (approved):

lotuuu (approved):

PR #451, opened 2024-04-08
Block tier ups by level

lotuuu:
Closes [#208](https://github.com/lambdaclass/champions_of_mirra/issues/208)

Don't allow units to be tiered up unless they cannot be leveled up further


lotuuu:
Sorry! I forgot! Also, this is an useful command to get more gold to level up your unit:
`GameBackend.Users.Currencies.add_currency_by_name!("3b90ac7d-1dcc-4ba1-b3d9-b40985014319", "Gold", 99999)`

martinacantaro (approved):
Tried tiering up with and without reaching max level, and it returned OK and error accordingly

PR #456, opened 2024-04-09
[GH-409] Leave matchmaking lobby

### LINKED ISSUE #409: [Feature] Allow a player to unjoin a game (game_launcher)

AminArria:
Closes #409 

Test with https://github.com/lambdaclass/curse_of_mirra/pull/1664


apps/arena/lib/arena/socket_handler.ex
```diff
@@ -32,6 +32,11 @@ defmodule Arena.SocketHandler do
   end
 
   @impl true
+  def websocket_handle({:text, "leave_lobby"}, state) do
```
AminArria, line 35: I would like opinions on this, receive the action as a plain text rather than protobuf. On one hand this is waaayy simpler and works, on the other it kinda goes against having everything in protobuf

So looking for opinions

AngieDutra (dismissed):
I playtested client and back PRs and had a weird bug:
I used Unity dektop and an android build with an ngrok and noticed I could control my Unity desktop character with my android build. If you need help replicating this bug let me know

AminArria:
> I playtested client and back PRs and had a weird bug...

Problem was in the client, it has been fixed now

apps/arena/lib/arena/socket_handler.ex
agustinesco: i don't see it as a big issue, i think protobuf as a way to transport structs between the different places, a leave lobby signal is not a struct so i don't see it as an issue

Nico-Sanchez (approved):

manucamejo (approved):

lotuuu (dismissed):

agustinesco (approved):
LGTM!

manucamejo (dismissed):

manucamejo (approved):

agustinesco (approved):
LGTM!

PR #459, opened 2024-04-09
Fix valtimer singularity pulling you inside obstacles

agustinesco:
Valtimer was pulling people inside the colliders and with his tp you were able to go inside rocks

to test this add this to the config file in the "obstacles" list and check that valtimer don't go inside of the obstacle and move a player close to it and try to move it inside of the obstacle:

<details>
<summary>Emerald</summary>

``` json
     {
        "position": {
          "x": 0.0,
          "y": 0.0
        },
        "vertices": [
          {
            "x": 0.0,
            "y": 3000.0
          },
          {
            "x": 2000.0,
            "y": 1000.0
          },
          {
            "x": 2000.0,
            "y": -1000.0
          },
          {
            "x": 0.0,
            "y": -3000.0
          },
          {
            "x": -2000.0,
            "y": -1000.0
          },
          {
            "x": -2000.0,
            "y": 1000.0
          }
        ],
        "radius": 0.0,
        "shape": "polygon"
      }
```
</details>

### DIFF
--- a/apps/arena/native/physics/src/lib.rs
+++ b/apps/arena/native/physics/src/lib.rs
@@ -77,12 +77,23 @@ fn move_entity_to_position(
     entity: Entity,
     new_position: Position,
     external_wall: Entity,
+    obstacles: HashMap<u64, Entity>,
 ) -> Entity {
     let mut entity: Entity = entity;
     entity.position = new_position;
 
     if entity.category == Category::Player && !entity.is_inside_map(&external_wall) {
         entity.move_to_next_valid_position_inside(&external_wall);
+    }
+
+    let collides_with = entity.collides_with(obstacles.clone().into_values().collect());
+
+    if entity.category == Category::Player && !collides_with.is_empty() {
+        let collided_with: Vec<&Entity> = collides_with
+            .iter()
+            .map(|id| obstacles.get(id).unwrap())
+            .collect();
+        entity.move_to_next_valid_position_outside(collided_with);
     }
     entity
 }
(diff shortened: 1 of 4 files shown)


apps/arena/native/physics/src/lib.rs
```diff
@@ -77,20 +77,50 @@ fn move_entity_to_position(
...
+
+    let collides_with = entity.collides_with(obstacles.clone().into_values().collect());
+
+    if entity.category == Category::Player && !collides_with.is_empty() {
+        let collided_with: Vec<&Entity> = collides_with
+            .iter()
+            .map(|id| obstacles.get(id).unwrap())
+            .collect();
+        entity.move_to_next_valid_position_outside(collided_with);
+    }
```
Nico-Sanchez, line 123: This is done twice, can we move it to a separate function?

Nico-Sanchez (changes requested):
PR works amazingly well! Congrats!

Left a comment to refactor some code.

agustinesco:
While testing the corrections i noticed that if you try to tp inside an obstacle valtimer is walking to the destination instead of teleportong : 

https://github.com/lambdaclass/mirra_backend/assets/106101218/c46175f9-b3a3-41ef-8f0a-97e06b9ca54e

This is due to how the client handle the teleport and the result of the first destination that we sent to the cleint dont being the same as the final destination. To fix this i calculated the final destination and then store it when you first execute the skill


https://github.com/lambdaclass/mirra_backend/assets/106101218/e31bcabe-d11c-4495-b01a-192e9f2b835c

apps/arena/native/physics/src/lib.rs
agustinesco: this is not just done twice but four times, i moved all this repeated code to a separated function

Nico-Sanchez (approved):

PR #461, opened 2024-04-10
Balance Valtimer

marco-paulucci:
Teleport: 
- faster execution
- longer teleport range
- cooldown to 6s (most char dashes are 5s) due to strong mechanic

Basic
- fat chunk projectile
- reduced splash
- adjusted damage
- faster execution

Pileta
- sliiightly reduced damage. I think it shouldn't be a lot better than h4cks ulti because both deal similar damage in total, but since Singu does it over a longer period, it's easier to dash away.


AngieDutra (approved):
Tested in playtest

BertovDev (approved):

PR #465, opened 2024-04-11
Make bot get request async to fix server timeout

agustinesco:
### Error
the join messages from the clients were timouting

### Cause
the game_launcher genserver was stuck every time we tried to spawn a bot because we did a GET request, and if the bot server is down or it took it a while to answer the game_launcher didn't answer to the client messages

### Solution
Make the bot request to the bot server async

### DIFF
--- a/apps/arena/lib/arena/game_launcher.ex
+++ b/apps/arena/lib/arena/game_launcher.ex
@@ -87,7 +87,7 @@ defmodule Arena.GameLauncher do
 
   def handle_info({:spawn_bot_for_player, bot_client, game_pid}, state) do
     Finch.build(:get, build_bot_url(game_pid, bot_client))
-    |> Finch.request(Arena.Finch)
+    |> Finch.async_request(Arena.Finch)
 
     {:noreply, state}
   end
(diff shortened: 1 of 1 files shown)


lotuuu (changes requested):
Works great!

I receive the following error (when trying with main in `client` and without running the `bot_manager` app):

```
[error] GenServer Arena.GameLauncher terminating
** (FunctionClauseError) no function clause matching in Arena.GameLauncher.handle_info/2
    (arena 0.1.0) lib/arena/game_launcher.ex:55: Arena.GameLauncher.handle_info({{Finch.HTTP1.Pool, #PID<0.1218.0>}, {:error, %Mint.TransportError{reason: :econnrefused}}}, %{clients: [], batch_start_at: -576460714887})
    (stdlib 5.2) gen_server.erl:1095: :gen_server.try_handle_info/3
    (stdlib 5.2) gen_server.erl:1183: :gen_server.handle_msg/6
    (stdlib 5.2) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
Last message: {{Finch.HTTP1.Pool, #PID<0.1218.0>}, {:error, %Mint.TransportError{reason: :econnrefused}}}
State: %{clients: [], batch_start_at: -576460714887}
```

which as we discussed is caused by a message `{:error, %Mint.TransportError{reason: :econnrefused}` we're not handling in our genserver. Doc: https://hexdocs.pm/finch/Finch.html#async_request/3

Let's handle it so our genserver doesn't die!

lotuuu (approved):

Nico-Sanchez (approved):

PR #468, opened 2024-04-12
Integrate battles and levels

lotuuu:
Frontend PR: https://github.com/lambdaclass/champions_of_mirra/pull/324 (Waiting on)

Closes https://github.com/lambdaclass/champions_of_mirra/issues/351

- Get rid of the BattleTest endpoint
- Use Simulator.battle() in fights
- Add handler for unimplemented executions
  - Unimplemented stats Modifiers (`damage_reduction`) and Components (`AddTags`) don't need any extra handling because of how we implemented them :D
  - Edit: Unimplemented stat modifiers don't need extra handling _on Simulator_, but they do require it in the Stat enum for the Protobuf message :c

# To test
In the frontend, go to the campaign and fight some levels! any level you fight should be playable without errors. Pro tip: If you're struggling with a level, try leveling up your units ;) or even get lucky and roll a Muflus in Summon

### DIFF
--- a/apps/gateway/test/champions_test.exs
+++ b/apps/gateway/test/champions_test.exs
@@ -339,11 +339,11 @@ defmodule Gateway.Test.Champions do
       fetch_last_message(socket_tester)
 
       assert_receive %WebSocketResponse{
-        response_type: {:battle_result, _ = battle_result}
+        response_type: {:battle_result, battle_result}
       }
 
       # Battle result should be either win or loss
-      assert battle_result.result == "win" or battle_result.result == "loss"
+      assert battle_result.result in ["team_1", "team_2", "draw", "timeout"]
 
       # TODO: check rewards [#CHoM-341]
     end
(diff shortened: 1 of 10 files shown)


apps/gateway/test/champions_test.exs
```diff
@@ -339,11 +339,11 @@ defmodule Gateway.Test.Champions do
       fetch_last_message(socket_tester)
 
       assert_receive %WebSocketResponse{
-        response_type: {:battle_result, _ = battle_result}
+        response_type: {:battle_result, battle_result}
       }
 
       # Battle result should be either win or loss
```
ncontinanza, line 345:
```suggestion
      # Battle result should be either team_1, team_2, draw or timeout
```

ncontinanza (changes requested):

ncontinanza (approved):

agustinesco (approved):

PR #469, opened 2024-04-15
Delete unused deploy script and generate protos

agustinesco:
This pr fix two things:

- The bot manager had an unused deploy script
- The proto were generated with an old elixir version causing the `syntax: :proto3` to be in a wrong position  and changing every time you generated the protos
- Some application didn't have the `messages.pb` files generated, so this pr will update them by running `make generate protos`

To tests this pr simply run `make generate protos` in your local computer and see that nothing changes


apps/bot_manager/lib/protobuf/messages.pb.ex
```diff
@@ -299,7 +308,14 @@ defmodule BotManager.Protobuf.GameState do
...
   field(:items, 13, repeated: true, type: BotManager.Protobuf.GameState.ItemsEntry, map: true)
-  field(:pools, 14, repeated: true, type: BotManager.Protobuf.GameState.PoolsEntry, map: true)
+
+  field(:obstacles, 14,
+    repeated: true,
+    type: BotManager.Protobuf.GameState.ObstaclesEntry,
+    map: true
+  )
+
+  field(:pools, 15, repeated: true, type: BotManager.Protobuf.GameState.PoolsEntry, map: true)
```
Nico-Sanchez, line 318: Hmmm you're pushing :pools from 14th to 15th position. Isn't that what we should avoid?

apps/bot_manager/lib/protobuf/messages.pb.ex
```diff
@@ -328,13 +344,13 @@ defmodule BotManager.Protobuf.Entity do
   field(:pool, 17, type: BotManager.Protobuf.Pool, oneof: 0)
 end
 
-defmodule BotManager.Protobuf.Player.EffectsEntry do
+defmodule BotManager.Protobuf.Player.CooldownsEntry do
```
Nico-Sanchez, line 347: What about the EffectsEntry?

Nico-Sanchez (changes requested):

apps/bot_manager/lib/protobuf/messages.pb.ex
Nico-Sanchez: Answered in Slack. What happened is that the protobuf files generated by messages.proto file were outdated.
Just ran `make generate-protos` then pushed.

apps/bot_manager/lib/protobuf/messages.pb.ex
Nico-Sanchez: Answered above.

Nico-Sanchez (approved):

AminArria (approved):

PR #470, opened 2024-04-15
[GH-454] implement crates feature

### LINKED ISSUE #454: Implement Crates
Crates are breakable boxes that have HP=250 and don't move. They drop a single power-up when destroyed.

Crates are market with pink in the map. Ignore the black squares (I put too many crates at the start)

image

agustinesco:
closes #454 

Added crate entities to the game logic, they will work like this:
- They should receive damage just as another player
- When they health drop to 0 they should be destroyed and drop an x amount of power ups

Test with [This client branch](https://github.com/lambdaclass/curse_of_mirra/pull/1676)

To test the crates add this to your config.json file in the crates lists

```json
{
      "position": {
        "x": 0.0,
        "y": 0.0
      },
      "shape": "circle",
      "radius": 250.0,
      "health": 250,
      "vertices": [],
      "amount_of_power_ups": 5
    },
    {
      "position": {
        "x": -2000.0,
        "y": -2000.0
      },
      "shape": "circle",
      "radius": 250.0,
      "health": 250,
      "vertices": [],
      "amount_of_power_ups": 5
    }
```


apps/arena/lib/arena/game/crate.ex
```diff
@@ -0,0 +1,16 @@
...
+  def take_damage(crate, damage) do
+    update_in(crate, [:aditional_info, :health], fn current_health -> max(current_health - damage, 0) end)
+  end
+
+  def alive?(crate) do
+    crate.aditional_info.health > 0
+  end
+
+  def interactable_crates(crates) do
+    Map.filter(crates, fn {_crate_id, crate} -> crate.aditional_info.status != :DESTROYED end)
```
AminArria, line 14: Any reason for this to not be based on `alive?/1` ?

apps/arena/lib/arena/game/skill.ex
```diff
@@ -40,7 +41,28 @@ defmodule Arena.Game.Skill do
...
+      Physics.check_collisions(circular_damage_area, interactable_crates)
+      |> Enum.reduce(game_state.crates, fn crate_id, crates_acc ->
+        real_damage = Player.calculate_real_damage(entity_player_owner, circle_hit.damage)
+
+        crate =
+          Map.get(crates_acc, crate_id)
+          |> Crate.take_damage(real_damage)
+
+        unless Crate.alive?(crate) do
+          send(self(), {:destroy_crate, crate_id})
```
AminArria, line 59: I understand why doing this, but how about no? Instead we can do nothing and in the game tick check for crates that died and mark them as destroyed or whatever we want

apps/arena/lib/arena/game_updater.ex
```diff
@@ -924,34 +985,71 @@ defmodule Arena.GameUpdater do
...
+
+  # Projectile collided a crate
+  defp process_projectile_collision(
+         projectile,
+         %{category: :crate} = crate,
+         _,
+         _,
+         {projectiles_acc, players_acc, crates_acc}
+       )
+       when not is_nil(crate) do
```
AminArria, line 1001: This guard is useless, if the pattern `%{category: :crate} = crate` matches then it can't ever be `nil`

apps/arena/priv/config.json
```diff
@@ -553,5 +552,31 @@
         }
       }
     }
+  ],
+  "crates": [
+    {
+      "id": "1",
```
AminArria, line 558: Remove this

AminArria (changes requested):

apps/arena/lib/arena/game/crate.ex
agustinesco: no reason at all, changed!

apps/arena/lib/arena/game_updater.ex
agustinesco: you're right! the player pattern match had something similar so i deleted that too

apps/arena/priv/config.json
agustinesco: deleted!

apps/arena/lib/arena/game/skill.ex
agustinesco: hahaha i liked that way of rejecting something 😛  

sure thing! i moved this spawn power up handling to a handler in the main tick, also this makes mi wornders, what stops us to make this with the player? 🤔 

Also i added some sections to the main tick pipeline because it started to get a bit messy

apps/arena/lib/arena/game_updater.ex
```diff
@@ -821,6 +859,20 @@ defmodule Arena.GameUpdater do
     %{game_state | players: updated_players}
   end
 
+  defp handle_destroyed_crates(%{crates: crates} = game_state, game_config) do
+    Enum.reduce(crates, game_state, fn {crate_id, crate}, game_state ->
+      if Crate.alive?(crate) or crate.aditional_info.status == :DESTROYED do
```
AminArria, line 864: Is this removing the crates? Or are we not supposed to?

apps/arena/priv/config.json
```diff
@@ -553,5 +552,6 @@
         }
       }
     }
-  ]
+  ],
+  "crates": []
```
AminArria, line 556: Is this correct?

apps/arena/lib/arena/game_updater.ex
agustinesco: nope, this is not removing the crates from the game, just checking if it should spawn a power up and destroying it. Would you rather remove destroyed crates from the state?

AminArria: Aaahh got it, honestly no strong opinions about it. We do have problems with the size of the binary, but a couple crates is the least of our problems and we can tackle it later on

AminArria (approved):

AminArria (dismissed):

AminArria (dismissed):

Nico-Sanchez (approved):

Nico-Sanchez (approved):

AminArria (approved):

PR #464, opened 2024-04-11
Implement defense

lotuuu:
[Issue](https://github.com/lambdaclass/champions_of_mirra/issues/326)

Adds defense to DealDamage Executions according to the [game design](https://www.notion.so/lambdaclass/Mechanics-fda91b86b0cf450d92ff08fd08ea8813).

A necessary change has been added where we abstract the TestUtils, so that we could use them for our new Defense test (and future ones as well)

# To Test

Run a battle with Muflus like detailed on `battle.md`. Check the history output for EFFECT_HIT of basic skills. They should be dealing less damage now:
```
          action_type: {:skill_action,
           %{
             caster_id: "e119b3b4-e189-42f0-9967-373bf2639881",
             skill_id: "037ec7ec-466c-4f47-998e-5bbae261cae7",
             skill_action_type: :EFFECT_HIT,
             stats_affected: [%{stat: :HEALTH, amount: -28}],
             target_ids: ["e66eec75-17ca-4c04-a0ac-2b3865c98580"]
           }}
        }
```
They were dealing 50 damage before.

PR #471, opened 2024-04-15
Implement Tags for Champions Battle

lotuuu:
- Implements tags behaviour with `ApplyTags` components
- Implements "Stun" tag custom behavior (unit won't be able to attack)
- Implements `TargetTagRequirements` component, that determines whether a skill hits or not based on if the target has a skill

Tests haven't been added in this PR so as to not generate conflicts with #464 . Issue is [here](https://github.com/lambdaclass/champions_of_mirra/issues/356).

# To Test

Battle in battle.md should work. Additionally, you should see many `Applying tag "Stun" to unit Muflus-e76` logs, which render units unable to cast skills.

If you want to do more in-depth testing, something you can do is play around with `ApplyTags` and `TargetTagRequirements`. For example, you can give Muflus' ultimate a TargetTagRequirement of a random tag that obviously won't exist. If you do so, then no one will get stunned during battle. However, if you make his basic apply the tag you set (remember to turn the effect into a "Duration" type) then stuns will happen because the tag is found now.

Feel free to find your own ways to experiment with this!

### DIFF
--- a/apps/champions/lib/champions/battle/simulator.ex
+++ b/apps/champions/lib/champions/battle/simulator.ex
@@ -243,6 +243,9 @@ defmodule Champions.Battle.Simulator do
         overrides: overrides
       }
 
+    # Reduce tags remaining timers & remove expired ones
+    {new_tags, new_history} = reduce_tag_timers(unit, new_history)
+
     # Reduce basic skill cooldown
     new_state =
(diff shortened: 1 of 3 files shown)


ncontinanza (approved):
LGTM!

PR #472, opened 2024-04-15
[gh-433] Fix killfeed not displaying names

### LINKED ISSUE #433: [BUG] Names not appearing in killfeed message
Sometimes the killfeed message does not contain the player name

agustinesco:
Closes #433 

The killfeed were not displaying character names sometimes, this was caused bu the rust `Entity` structure not containing this field thus causing it to be lost when you called the `physics` module to process an entity calculation 

Fixed added the name to the rust entity


Nico-Sanchez (approved):

lotuuu (approved):
Works great!

PR #473, opened 2024-04-15
Refactor Gateway Messages

lotuuu:
⚠️ Waiting on: [Frontend PR
]()
The design we chose for battle replays was flawed. This aims to fix it by separating responsibilities of 
a. notifying when a skill hits
and
b. notifying damage/heals/energising via executions

# To test

Battles through the client should work as expected


martinacantaro (approved):
Tested the battle and it works as it should

ncontinanza (approved):
LGTM!

martinacantaro (approved):
tried the battle and it works normally from the Client

PR #476, opened 2024-04-16
Fix sample user and add doc to Characters

lotuuu:
add_sample_units was using get_characters_by_rank but we want get_characters_by_quality.

Also adds docs for Characters

# To test

Create a user through the client succesfully (just run the app)


ncontinanza (approved):
LGTM

martinacantaro (approved):
Tested it, created a user and the initial units were:
 - Otix * 2
 - D'Agna
 - Kenzu
 - H4ck
 - Uren
What was the reason behind this change?

PR #477, opened 2024-04-16
Fix dead units targeting in delayed effects

lotuuu:
Fixes the targeting of dead units in effects with delays

# To test

Change create_effect_map to
```
defp create_effect_map(%Effect{} = effect, skill_id),
    do: %{
      type: effect.type,
      delay: 15,
...
```

hardcoding a high delay value. This will make it so that sometimes units miss because their target died before the effect's delay ends. Battle should still work fine


apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -415,24 +415,26 @@ defmodule Champions.Battle.Simulator do
...
-              update_in(current_state, [:units], fn units -> Map.merge(units, targets_after_effect) end)
+            new_state = update_in(current_state, [:units], fn units -> Map.merge(units, targets_after_effect) end)
 
             # We don't add this effect to the new_pending_effects list because it has already been applied
             {{new_pending_effects, new_state}, new_history}
 
           effect ->
             # If the effect isn't ready to be processed, we reduce its remaining delay.
-            {{[%{effect | delay: effect.delay - 1} | new_pending_effects], current_state}, new_history}
+            {{[%{effect | delay: effect.delay - 1} | new_pending_effects], current_state}, history}
```
ncontinanza, line 428:
```suggestion
            {{[%{effect | delay: effect.delay - 1} | new_pending_effects], current_state}, new_history}
```

ncontinanza (changes requested):

ncontinanza (approved):

martinacantaro (approved):
functional review approved!

PR #479, opened 2024-04-16
Fix tag encoding

lotuuu:
Fixes an atom that was uppercased

# To test

Run a battle through the client, it should work fine!


ncontinanza (approved):

martinacantaro:
it works!

martinacantaro (approved):

PR #480, opened 2024-04-16
[GH-404] Allow to configure item radius size

### LINKED ISSUE #404: Increase items and power ups pickup area
Now it's difficult to pick up items or power-ups; you need to walk over them exactly, and it doesn't feel right

image

agustinesco:
Closes #404 

The item size wasn't accurate and it was not easily configurable, this pr will add the item size to the config file


apps/arena/priv/config.json
```diff
@@ -399,6 +399,9 @@
       ]
     }
   ],
+  "items_params": {
+    "radius": 200.0
+  },
```
AminArria, line 404: Please move this into each item config

AminArria (changes requested):

apps/arena/priv/config.json
agustinesco: Sounds good!

apps/arena/priv/config.json
```diff
@@ -521,8 +524,7 @@
         "reduce_stamina_interval": {
           "decrease_by": 0.3
         },
-        "refresh_stamina": {
-        }
+        "refresh_stamina": {}
```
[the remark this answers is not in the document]
agustinesco: this is happening in various prs, i think it's another issue with the formatter? 🤔

Nico-Sanchez (approved):

PR #481, opened 2024-04-16
Fix history occasionally removing entries

lotuuu:
We were using the wrong history variable, which made it so that sometimes our actions were removed from the accumulator. I think this happened everytime we had effects being processed while other effects were having their delay reduced

# To Test

We noticed this issue when running a fight between 1 Muflus and 1 Uma (low level). When this happened, step 25 showed no hits in the history but it did show logs in the battle. This PR should show logs in step 25 in said battle.


ncontinanza (dismissed):
LGTM!

tkz00 (approved):
Awesome work!

PR #482, opened 2024-04-17
Fix protobuf message names for tags and modifiers

tkz00:
The protobuf messages didn't match with the names of the fields being returned in the history of the battles. In the protobuf messages "tag_name" has been changed to "tag" and now the history returns "target_id" instead of "unit_id" or "target_ids" for modifiers and tags received and expired.

All four of those messages should have the target_id like this:
```
%{
  action_type: {:tag_received,
   %{
     tag: "Stun",
     skill_id: "8cb48f51-0719-4d88-9b4f-6581efabf680",
     target_id: "64514657-0dc5-431b-ac19-5bb7cacb4a11"
   }}
}
```

And the `tag_received` message should send `tag` field instead of `tag_name`.

To test these changes run a battle and check that the history returned by the simulator complies with these conditions.

### DIFF
--- a/apps/champions/lib/champions/battle/simulator.ex
+++ b/apps/champions/lib/champions/battle/simulator.ex
@@ -106,7 +106,6 @@ defmodule Champions.Battle.Simulator do
         end
       )
 
-    IO.inspect(Enum.reverse(history), limit: :infinity)
     %{initial_state: transform_initial_state_for_replay(initial_state), steps: Enum.reverse(history), result: result}
   end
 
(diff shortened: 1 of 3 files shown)


lotuuu (changes requested):
You missed changing the `tag_name` in `TagExpired` to `tag` (`gateway.proto`)!

tkz00:
> You missed changing the `tag_name` in `TagExpired` to `tag` (`gateway.proto`)!

Solved!

ncontinanza (approved):

PR #489, opened 2024-04-17
[GH-422] increase health when picking up power up

### LINKED ISSUE #422: [BUG] Power-ups should increase HP as well
by a percentage (10%) of BASE hp (not current nor modified, i.e. it is additive with itself)

this works as a heal too.

agustinesco:
Closes #422

Now when picking up power ups you also get a configurable amount of health that scales based on the base health

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -10,6 +10,7 @@ defmodule Arena.GameUpdater do
   alias Arena.Game.{Player, Skill}
   alias Arena.Serialization.{GameEvent, GameState, GameFinished}
   alias Phoenix.PubSub
+  alias Arena.Utils
 
   ##########################
   # API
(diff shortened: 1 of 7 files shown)


apps/arena/lib/arena/game_updater.ex
```diff
@@ -1015,7 +1015,17 @@ defmodule Arena.GameUpdater do
...
+      updated_player =
+        update_in(player, [:aditional_info, :power_ups], fn amount -> amount + 1 end)
+        |> update_in([:aditional_info], fn additional_info ->
+          Map.update(additional_info, :health, additional_info.health, fn current_health ->
+            (current_health + additional_info.base_health * power_up.aditional_info.power_up_health_modifier)
+            |> round()
+          end)
+          |> Map.update(:max_health, additional_info.max_health, fn max_health ->
+            (max_health + additional_info.base_health * power_up.aditional_info.power_up_health_modifier) |> round()
+          end)
```
Nico-Sanchez, line 1027: These updates are the same. I think we can create a Utils function like `increase_health_by_base_health_fraction(health, base_health, modifier)` What do you think?

AminArria: Even better a more generic `increase_current_by_percentage(current_value, base_value, percentage)`  and in theory you can event pattern match `is_integer(current_value)` to decide if to round or not

agustinesco: really nice suggestions! changed

agustinesco: great suggestions guys! changed

Nico-Sanchez (approved):

AminArria (approved):

PR #491, opened 2024-04-17
[GH-486] Spawn bot when player disconnect 

### LINKED ISSUE #467: Change handling of opaque type of finch to spawn bots
Currently we're spawning bots using finch `async_request` function, this was chaged from a sync request due to and error that caused the game_launcher process to get stuck waiting for spawn bots. This send a message to the game_launcher server that we need to handle in order for the game launcher to not crash. This is bad since the match that we're using to identify that the message comes from finch is using the opaque type `request_ref()` and if they change that in the future our application will break

We can fix this by changing the async request to a sync one again but handling the sync request but [Spawning](https://hexdocs.pm/elixir/processes.html) a new process to use it

### LINKED ISSUE #486: Replace player with bot when he disconnect
When a player leaves a match it should be replaced with a bot. This will be done instantly.
Player won't be able to rejoin (we could create another issue for this).

agustinesco:
Closes #486 
Closes #467 

### Replace player with bot when he disconnects

Now when the `Arena.GameSocketHandler` websocket get termianted and the game still running we'll call the bot server to spawn a bot and control that character

### Replace opaque finch handling

We'll replace the handler for finch async_request calling a elixir process to call the bot spawn get request  in the `game_launcher.ex` module

### Testing

To test this you can enter a match though [this link ](http://localhost:3000/) and start 2 players, when the match starts you can close one of the tabs and the player that belonged to that tab should start to be controlled by a bot

Also this can be done with the unity client, exiting the match with the "exit" button in the options menu.
Disclaimer: the "play" button from the unity application doesn't work for testing this


apps/bot_manager/lib/game_socket_handler.ex
```diff
@@ -46,9 +46,6 @@ defmodule BotManager.GameSocketHandler do
       %{event: {:joined, joined}} ->
         {:ok, Map.merge(state, joined)}
 
-      %{event: {:finished, _}} ->
```
Nico-Sanchez, line 49: Why do you remove this? If this is correct you should delete the terminate/1 callback aswell.

Nico-Sanchez (changes requested):
There's a `BOTS_ACTIVE` env var that does nothing really. Please delete them from the code and deployment script/workflows. That way we can test this easily.

apps/bot_manager/lib/game_socket_handler.ex
Nico-Sanchez: > Now the bot socket will be closed when the game closes after 10 seconds of the game being finished

This ws closes when the game finishes and you send a message to its dead process.
That's because we don't close the websockets from the server (we have an issue for that), so you should close the client ws gracefully or you could trigger several errors on Arena.
Try to restore this and stop the spawning of the bots from Arena.

Nico-Sanchez (approved):

lotuuu (approved):
Works great! Couldn't get it to work when closing the game execution from within Unity (pressing the play button). Seems more like a problem with Unity than our implementation, since exiting match from the settings screen, from game_client, and by `cmd+q`ing Unity works fine!

PR #523, opened 2024-04-23
Update araban colliders

AngieDutra:
This PR:
- adds colliders
- adjusts spawn positions


apps/arena/priv/config.json
```diff
@@ -65,11 +65,391 @@
...
+            "y": -1130.45
+          },
+          {
+            "x": -1248.0,
+            "y": -299.0
+          }
+        ],
+        "radius": 0.0,
+        "shape": "polygon",
+        "id": "collider 4 - a"
```
manucamejo, line 291: Is this "id" necessary? Should we add it to the other obstacles?

AngieDutra: I thought doing a cheat sheet to edit the colliders could be helpful to later on change individually a section of vectors head on, instead of debugging by changing random values to get the vector I want and then start to work on it. I was starting to do this:
Screenshot 2024-04-29 at 08 48 01
But since the colliders needed to be re-done later on I just paused this since it was just and idea and not a formal request. I planned on resume this whenever I had to do the colliders again. But, as I said, the IDs could be deleted and this idea dismissed if you see fit

manucamejo:
I'm closing it because there is a [newer one](https://github.com/lambdaclass/mirra_backend/pull/547).

PR #528, opened 2024-04-24
Add Class Diagram

lotuuu:
Adds a class diagram as documentation for GameBackend.

[Here's](https://github.com/lambdaclass/mirra_backend/blob/add-graph/apps/game_backend/README.md) how it looks like!


martinacantaro (approved):
Approved!

mgrunwaldt (approved):

PR #530, opened 2024-04-25
Item config & level up logic

lotuuu:
Implements ItemTemplate configuration and tier up (item fusion) functionalities.

Item tier ups work by combining many items into a new one. Unlike units, where we delete all units but one and raise the latter's tier, here we create a new item altogether. Reason is we don't have any more data other than the unit it's equipped to: The template defines the item now, they don't hold a level themselves. So really there's not really an issue with removing all items entirely.

Source of truth: https://docs.google.com/spreadsheets/d/177mvJS75LecaAEpyYotQEcrmhGJWI424UnkE2JHLmyY/edit#gid=1542073627

# To Review
Most of the change lines come from the item_templates.json file. The main file changes on this PR are:
- `apps/champions/lib/champions/config.ex`, where we use some new utility functions defined in `GameBackend.Items` to import the item template config
- `apps/champions/lib/champions/items.ex`, where we define the logic for item fusion
- `apps/game_backend/lib/game_backend/items/item_template.ex` and `apps/game_backend/priv/repo/migrations/20240425171046_add_item_template_fields.exs`, where we define the new fields for the ItemTemplate schema.

Of course, this is just meant to help make PR review process a bit easier given its large amount of files and lines changed, but do check all the of them anyways!

# To test

You can test item fusion by running `Champions.Items.fuse(user_id, [item_1.id, item_2.id, item_3.id])`. Remember that when you define your items' template, you also need to define the template for what they will upgrade into! Check the tests added [here](https://github.com/lambdaclass/mirra_backend/blob/d796ffe9b1c0a5d25a2ed88e6d6f1a8a0b2a5fbc/apps/gateway/test/champions_test.exs#L731) for an example.


ncontinanza (changes requested):

ncontinanza (approved):

apps/champions/lib/champions/items.ex
```diff
@@ -12,36 +14,94 @@ defmodule Champions.Items do
...
+          Currencies.substract_currencies(user_id, new_template.upgrade_costs)
+        end)
+        |> Transaction.run()
+
+      case result do
+        {:error, _, _, _} ->
+          {:error, :transaction}
+
+        {:ok, %{item: item}} ->
+          Items.get_item(item.id)
```
agustinesco, line 68: why not simply returning the item?

apps/champions/lib/champions/items.ex
```diff
@@ -12,36 +14,94 @@ defmodule Champions.Items do
...
+      end
+    else
+      {:consumed_items_owned, false} ->
+        {:error, :consumed_items_not_found}
```
agustinesco, line 72:
```suggestion
        {:error, :consumed_items_not_owned}
```

apps/champions/lib/champions/items.ex
```diff
@@ -12,36 +14,94 @@ defmodule Champions.Items do
...
+
+      {:consumed_items_valid, {false, _}} ->
+        {:error, :consumed_items_invalid}
+    end
+  end
+
+  defp delete_consumed_items(item_ids) do
+    {amount_deleted, _return} = Items.delete_items(item_ids)
+
+    if Enum.count(item_ids) == amount_deleted, do: {:ok, amount_deleted}, else: {:error, "failed"}
```
agustinesco, line 85: In which situations this would fail?

apps/champions/lib/champions/items.ex
lotuuu: I chose to hide the fact that an item with the given id actually exists when the user doesn't own it. It's none of their business!

apps/champions/lib/champions/items.ex
lotuuu: This is to get the preloads that protobuf messages need

tkz00 (commented):
Do you think adding `Champions.Items.fuse` function to the readme file would be a good idea?

tkz00 (approved):

apps/champions/lib/champions/items.ex
agustinesco: right!

agustinesco (approved):

PR #532, opened 2024-04-26
[GH-531] Add google JWT validation

### LINKED ISSUE #531: Add google token validation
This is part of https://github.com/lambdaclass/curse_of_mirra/issues/961
The sign in will be handled in the client, we'll provide an endpoint for google JWT (Jason web token) verify and validation.

Nico-Sanchez:
## Motivation
We need a way to handle the google sign in validation from the backend server.
Closes #531 

## Changes
- [Add secret_key_base to gateway endpoint](https://github.com/lambdaclass/mirra_backend/pull/532/commits/3d3841c804091d2b550e2da8383e300d43e7de17)
- [Add GoogleUser table and schema](https://github.com/lambdaclass/mirra_backend/pull/532/commits/c213623347b882aa3b669f30f8c8ef6dfb3e3e41)
- [Add find and create functions for GoogleUser](https://github.com/lambdaclass/mirra_backend/pull/532/commits/220b881a996b7e8550a5041ce26ce25958516e48)
- [Add find_or_create_google_user_by_email/1 function doc](https://github.com/lambdaclass/mirra_backend/pull/532/commits/f2dffb238061a38caff24d12e5046d3270fab806)
- [Add Joken and GoogleCertificates dependencies](https://github.com/lambdaclass/mirra_backend/pull/532/commits/98a7f873287b54e8c5db2833c9d91e509c5eecb5)
- [Add Joken hook to fetch google public keys and create a joken signer …](https://github.com/lambdaclass/mirra_backend/pull/532/commits/7b9761d743e6a695ef40d8a702ce1ba3064662d3)
- [Add GoogleTokenManager module to handle tokens verification and valid…](https://github.com/lambdaclass/mirra_backend/pull/532/commits/47a5e9b7fb89470ff6279052fe8a745a545003a6)
- [Start google certificates fetcher genserver with the application](https://github.com/lambdaclass/mirra_backend/pull/532/commits/0f99bfeae1ded49cbf921b3dfcffbaa7c8d463d2)
- [Add validate token endpoint](https://github.com/lambdaclass/mirra_backend/pull/532/commits/28b5aa319940d2f1304ff30442a4cc0d76c2a1b7)

## Docs
- [Google documentation](https://developers.google.com/identity/sign-in/web/backend-auth)
- Dependencies added:
  - [Joken](https://hexdocs.pm/joken/introduction.html) -> Dependency to verify and validate JWT.
  - [Google Certificates](https://hexdocs.pm/google_certs/readme.html) -> Dependency to fetch google JWK (public keys to validate signature).

## How to test it?
- Just get a tokenID and do a GET request to `http://localhost:4001/auth/google/token/:token_id`
- To get a tokenID ->
  - Create a google cloud dummy credential at https://console.cloud.google.com/.
  - Open cloud shell (see attached image).
  - Use the following command: `gcloud auth print-identity-token`
Screenshot 2024-04-30 at 5 01 15 PM


apps/gateway/lib/gateway/controllers/auth_controller.ex
```diff
@@ -0,0 +1,21 @@
...
+  def validate_token(conn, %{"token_id" => token_id}) do
+    conn = put_resp_content_type(conn, "application/json")
+
+    case GoogleTokenManager.verify_and_validate(token_id) do
+      {:ok, response} ->
+        Users.find_or_create_google_user_by_email(response["email"])
+        send_resp(conn, 200, Jason.encode!(response))
+
+      {:error, error} ->
+        send_resp(conn, 401, Jason.encode!(Map.new(error)))
```
AminArria, line 18: I would expect 400 in this case, why 401?

apps/gateway/lib/gateway/controllers/auth_controller.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Gateway.Controllers.AuthController do
+  @moduledoc """
+  Controller for users authentication.
+  """
+  use Gateway, :controller
+  alias Gateway.Auth.GoogleTokenManager
+  alias GameBackend.Users
+
+  def validate_token(conn, %{"token_id" => token_id}) do
+    conn = put_resp_content_type(conn, "application/json")
```
AminArria, line 10: Is this really necessary? Cause I'm pretty sure its not

apps/gateway/lib/gateway/auth/google_token_manager.ex
```diff
@@ -0,0 +1,23 @@
...
+  Module responsible to verify and validate the Google TokenIDs.
+  Its logic has been done following the GoogleCertificates docs:
+  https://hexdocs.pm/google_certs/readme.html
+  """
+
+  use Joken.Config, default_signer: nil
+
+  @iss "https://accounts.google.com"
+
+  defp aud, do: System.get_env("GOOGLE_CLIENT_ID")
```
AminArria, line 12: Let's get this from configuration rather than env vars

apps/gateway/lib/gateway/router.ex
```diff
@@ -9,6 +9,12 @@ defmodule Gateway.Router do
     pipe_through :api
   end
 
+  scope "/auth", Gateway do
+    pipe_through :api
+
+    get "/:provider/token/:token_id", Controllers.AuthController, :validate_token
```
AminArria, line 15: Either pattern match `:provider` in `validate_token/2` or remove it

AminArria (changes requested):

apps/gateway/lib/gateway/controllers/auth_controller.ex
Nico-Sanchez: It seems more accurate to me since 401 is "unauthorized" but I can change it.

apps/gateway/lib/gateway/controllers/auth_controller.ex
Nico-Sanchez: Yeah, its not. Deleted!

apps/gateway/lib/gateway/auth/google_token_manager.ex
Nico-Sanchez: Done! Let me know if it should be different.

apps/gateway/lib/gateway/router.ex
Nico-Sanchez: Done!

apps/gateway/lib/gateway/controllers/auth_controller.ex
AminArria: But that's the thing, the request is not unauthorized, it happens to be an invalid token (or something). A 401 refers to the authorization to make the request (do we allow this client to call this endpoint/path/resource?) and in this case we allow any client to call us (for now at least)

Based on that a 401 is incorrect (and by extension a 403), from our options remaining it would be either a 400 or a 422 that properly reflect the erro. I think a 400 is good enough because we are likely grouping a bunch of different errors together so it would be generic enough, but if we had details on the specific errors then we can differentiate between a malformed token which would be a 400 and an invalid token which would be a 422

Nico-Sanchez: Thanks a lot for the explanation, changed!

AminArria (approved):

agustinesco:
The application crashes when i send random things as token:
```
curl "http://localhost:4001/auth/google/token/1234"
{"errors":{"detail":"Internal Server Error"}}%
```

we should handle this right?

agustinesco:
Can you specify how can i get a google token_id so i can make a get request an create a google_user struct in the db?

Nico-Sanchez:
> The application crashes when i send random things as token:
> 
> ```
> curl "http://localhost:4001/auth/google/token/1234"
> {"errors":{"detail":"Internal Server Error"}}%
> ```
> 
> ```
> Request: GET /auth/google/token/1234
> ** (exit) an exception was raised:
>     ** (Protocol.UndefinedError) protocol Enumerable not implemented for :no_signer of type Atom. This protocol is implemented for the following type(s): DBConnection.PrepareStream, DBConnection.Stream, Date.Range, Ecto.Adapters.SQL.Stream, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, Jason.OrderedObject, List, Map, MapSet, Phoenix.LiveView.LiveStream, Postgrex.Stream, Range, Stream
> ```
> 
> we should handle this right?

If the application crashes, yeah, we should. However I think is just an error and the app is ok (smiles in Elixir).

Nico-Sanchez:
> Can you specify how can i get a google token_id so i can make a get request an create a google_user struct in the db?

Added to the descripion!

config/runtime.exs
```diff
@@ -315,3 +315,7 @@ config :bot_manager, :end_point_configuration,
   scheme: :http,
   plug: BotManager.Endpoint,
   options: [port: bot_manager_port]
+
+###################################
+
+config :joken, issuer: "https://accounts.google.com", audience: System.get_env("GOOGLE_CLIENT_ID")
```
agustinesco, line 321: Testing this is local the issuer was wrong, for some reason it expects "accounts.google.com" without the https, i replaced it in the code and could create a google user 🥳

agustinesco: This was an issue related to the way i generated the token

agustinesco (approved):

PR #543, opened 2024-04-29
[GH-479] Change calls for casts

### LINKED ISSUE #497: [REFACTOR] Use cast instead of call
Image

Image

manucamejo:
Closes #497 

Use cast instead of calls for move, attack and use item, because we are not expecting any response.
We keep call in the join message because we are using that response.

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -21,15 +21,15 @@ defmodule Arena.GameUpdater do
   end
 
   def move(game_pid, player_id, direction, timestamp) do
-    GenServer.call(game_pid, {:move, player_id, direction, timestamp})
+    GenServer.cast(game_pid, {:move, player_id, direction, timestamp})
   end
 
   def attack(game_pid, player_id, skill, skill_params, timestamp) do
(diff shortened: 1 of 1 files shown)


Nico-Sanchez (commented):
Is this properly tested? maybe we could run some loadtests using this, to see if this degrades our performance.
Generally speaking, a cast is better for scenarios where you don't want a reply, but maybe the sync call task is preventing some parts of our code to overload.
It's just an idea, I agree that cast seems a better option, but we could check it doing a loadtest with this changes.

manucamejo:
Yesterday I did a load test (1000 players) and it works fine!

Nico-Sanchez:
> Yesterday I did a load test (1000 players) and it works fine!

Can you attach some pictures regarding the server use (specially cpu usage)?
If you don't have any but you checked it and remember the number is fine as well. I just want to make sure that it stays the same (or better) than before.

manucamejo:
Yeah, I did it again, here is the video
https://github.com/lambdaclass/mirra_backend/assets/20927859/fbc688e2-fc55-430a-832d-623cc6d655c7

Nico-Sanchez (approved):

agustinesco (approved):
LGTM!

PR #544, opened 2024-04-29
[GH-411] The item spawning position shouldn't collide with an obstacle

### LINKED ISSUE #411: [BUG] Items should spawn in reachable positions
- [x] Check if collides with some obstacle.

manucamejo:
Create some obstacles to test it.

closes #411


agustinesco (approved):
LGTM! i let it run for a while and i creating this piece of modern art
image

If we have really bad luck the `random_position_in_map` could run forever if the position always land inside an obstacle, very improbable tho. Another solution could be pick a random location inside the map and if it collides with an obstacle resolve the collision and move it

Nico-Sanchez (approved):

PR #547, opened 2024-04-29
[GH-541] Add map collisionables version 3

### LINKED ISSUE #541: [REFACTOR] Colliders position according to concept

agustinesco:
Closes #541

Added new map obstacles according to new map design and adjust spawn positions


apps/arena/priv/config.json
```diff
@@ -70,6 +70,772 @@
...
+            "y": 1485.0
+          },
+          {
+            "x": -425.0,
+            "y": 1416.0
+          }
+        ]
+      },
+      {
+        "collider": "carapace 1",
```
AminArria, line 143: What is this field?

AminArria (approved):

apps/arena/priv/config.json
agustinesco: this a field that we'll use to identify the rocks in the map, it's really hard to identify where a polygon belongs without this field

manucamejo (dismissed):

apps/arena/priv/config.json
Nico-Sanchez: Can't it be the "name" instead of "collider"? I think that's a word of unity business logic, not ours

agustinesco: sure! yeah the name isn't important, we just want a way to recognize which polygon correspond with which collider

Nico-Sanchez (approved):

manucamejo (approved):

PR #552, opened 2024-04-30
Fix item tests

lotuuu:
## Motivation

We had a failing test

Closes https://github.com/lambdaclass/champions_of_mirra/issues/403

## Summary of changes

- Make assertion range more permissive 
## How to test it?

Run tests with the same seed with the following command:

`mix test --seed 444725`


agustinesco (approved):
LGTM!

PR #555, opened 2024-04-30
[GH-549] Fix valtimer teleporting inside polygons

### LINKED ISSUE #549: [BUG] Fix valtimer teleporting inside polygons
When you tp with valtimer inside a composed obstacle you could get inside of it

agustinesco:
## Motivation

Whenever you used valtimers teleport ability between two polygon obstacles you could get inside of a collider in the map

In a nutshell this error was due to how we process collision resolution with polygon and the fact that colliders in the map are composed of several polygons due to the fact that we can't have concave polygons. 
Since the collition resolution algorithm always return the closest vector that would solve the collition this could result in a movement that gets you inside another polygon, resulting in ending up inside the obstacles.

The fix will work by filtering movements that would result in ending up inside another obstacle, something like:

image

This was the bug present on main:

https://github.com/lambdaclass/mirra_backend/assets/67444519/d97bfeaa-5d2c-42cc-9e2a-ec8ec70d5a37

Closes #549 

08/05 update:
There was another case were an axis was invalid if the result would make the player to go outside the map 

## Summary of changes

- Added line to polygon collision detection
- Filter invalid axis when processing collision resolution with sat
- Fixed cannon stones polygon vertex to remove hole in the middle

## How to test it?
Start a game with valtimer and try to use the teleport inside a rock
Also try to walk to obstacles and check that it feels correct

## Checklist


manucamejo:
@agustinesco please do a loadtest and paste the results

agustinesco:
Ran a loadtest with 1000 players running and the cpu peaks at 20%


https://github.com/lambdaclass/mirra_backend/assets/106101218/5f179609-a3d7-47ea-81ef-8cfeeadcde78

apps/arena/native/physics/src/collision_detection/sat.rs
```diff
@@ -245,7 +257,7 @@ fn project_circle(circle: &Entity, axis: Position) -> (f32, f32) {
         y: axis.y * circle.radius,
     };
 
-    let position_plus_radius = Position::add(circle.position, direction_radius);
+    let position_plus_radius = Position::add(&circle.position, &direction_radius);
```
tvillegas98, line 260: Since the add function is receiving a reference, lets make all other Positions functions receive a reference as parameter, to have our code consistent

tvillegas98 (changes requested):
I found this bug when I was trying the PR, do you have a hint of what could be the cause?

https://github.com/lambdaclass/mirra_backend/assets/67444519/b4844f18-db58-4602-bad6-f32da0cdcc62

Also I left a comment about the rust code

apps/arena/native/physics/src/collision_detection/sat.rs
agustinesco: sounds good! changed

agustinesco:
> I found this bug when I was testing the PR, do you have a hint of what could be the cause?
> 
>  Screen.Recording.2024-05-07.at.14.21.46.mov 
> Also I left a comment about the rust code

hmmm nice catch there, i think this could be caused by two reasons:
1. This fix is not taking care of the external wall collision, so an edge outside of the map is considered as valid when it should
2. the specific stone you're bugged inside have a whole in it, so we should fix the polygon

agustinesco:
@tvillegas98 yeah in fact the bug was that we were considering axis that were outside of the play area as valid to resolve the collision with a polygon, which is wrong. Fixed both rocks on the edges of the map! also the stairs on the left side of the map had the same error so i fixed them too

tvillegas98 (approved):
LGTM!

manucamejo:
Hey @agustinesco I found this bug.
But let's merge this feature an please create a new issue for fixing it next week.

https://github.com/lambdaclass/mirra_backend/assets/20927859/177dd525-5fc4-4cae-9da0-f64280b924b4

apps/arena/priv/config.json
```diff
@@ -879,7 +848,7 @@
     "end_game_interval_ms": 1000,
     "shutdown_game_wait_ms": 10000,
     "natural_healing_interval_ms": 300,
-    "zone_shrink_start_ms": 35000,
+    "zone_shrink_start_ms": 999999,
```
manucamejo, line 851: Restore this!

manucamejo (changes requested):

apps/arena/priv/config.json
agustinesco: done!

manucamejo (approved):

tvillegas98 (approved):

PR #556, opened 2024-05-02
[GH-421] Golden clock improvements

### LINKED ISSUE #421: [REFACTOR] Golden clock (Item) improvements
Golden clock: 

- [ ] "interval_decrease_by" is this a flat amount deduction? It would be better as a percentage. For example, 100 here would mean it recharges twice as fast.

- [ ] It should halve all cooldowns (based on BASE value and not current) as well while effect is active (if a 10s cooldown skill has 5s left of cooldown and this item is taken, it should be available immediately)

- [ ] In the future when #394 is implemented it should grant mana as well.

manucamejo:
## Motivation

We need to improve the golden clock effects according to the specifications.
Closes #421

## Summary of changes
Stamina
- Recharge stamina
- Reduce stamina interval

Cooldowns
- Restore cooldowns
- Reduce cooldowns duration

## How to test it?
Use the golden clock item and you should feel a real boost now.

## Checklist


apps/arena/lib/arena/game/player.ex
```diff
@@ -375,7 +380,9 @@ defmodule Arena.Game.Player do
   end
 
   defp apply_skill_cooldown(player, skill_key, %{cooldown_mechanism: "time", cooldown_ms: cooldown_ms}) do
-    put_in(player, [:aditional_info, :cooldowns, skill_key], cooldown_ms)
+    cooldown_multiplier = player |> get_golden_clock_effect |> get_cooldown_multiplier
```
agustinesco, line 383: Why not having something similar to what the stamina is doing? have a field in the player additional info that determine a cooldown multiplier and change that when you gain the `golden_clock_effect`

agustinesco: As we talked, we can do something similar to what `apply_stat_modifier` is doing when it has to reduce the stamina interval, and use that with another field in the player called `cooldown_multiplier`

manucamejo: fixed

apps/arena/priv/config.json
```diff
@@ -423,6 +423,7 @@
       "base_size": 110.0,
       "base_health": 440,
       "base_stamina": 3,
+      "base_cooldown_multiplier": 1,
```
agustinesco, line 426: i wouldn't add this to the character, if we start to spread around this values maybe we'll lost track of what is altering the duration of cooldows, what do you think about just leave is as a player entity value and have 1 as default for everyone?

manucamejo: But I think that have a hardcoded value isn't a good idea

agustinesco (approved):
LGTM!

apps/arena/priv/config.json
agustinesco: it would be something similar to what `bonus_damage` is doing, since this is a entity modifier that should not come from the character since it's unrelated, maybe we could refactor this in another pr to put all the hardcoded player entity data in the config file

PR #558, opened 2024-05-02
Heal executions

lotuuu:
## Motivation

Add Heal executions.
Closes [#401](https://github.com/lambdaclass/champions_of_mirra/issues/401)

## Important Note

This breaks the client animation a little bit, under certain circumstances. Reason is that we are not notifying the client about health caps, so as far as it knows, a unit with `Max HP = HP = 100`, that in one step a `Heal of 20` and a `Damage of 10`, ends the step with `HP = 110`, resulting in a discrepancy with the backend (source of truth) where `HP = 100`. To fix this, an OverrideStat message will be implemented in an upcoming PR ([issue](https://github.com/lambdaclass/champions_of_mirra/issues/280)).

## Summary of changes

- Add heal execution
- Cap unit health points after step finishes processing (to account for simultaneous damages and heals)
- Add a test for Heal

## How to test it?

You can start the client and check your units. Choose one and edit their basic skill so that it heals allies instead of dealing damage. You can also change its cooldown so that it's very strong. Remember to run `Champions.Config.import_skill_config()` You will now end fights with super healthy units thanks to your healer!

Alternatively, just start a small fight with some healing unit in the console (check `battle.md` for examples, and `skills.json` to look for a healing character or to create your own). You should see a message getting logged when that unit casts its heal: "Healing N hp to X (HP -> HP+N)". Remember that if your [info] logs are getting cut off, you can swap `Logger.info` calls with `IO.inspect` to debug

### DIFF
--- a/apps/champions/test/support/test_utils.ex
+++ b/apps/champions/test/support/test_utils.ex
@@ -13,7 +13,7 @@ defmodule Champions.TestUtils do
         class: "Assassin",
         base_attack: 100,
         base_health: 100,
-        base_defense: 100,
+        base_defense: 0,
         basic_skill: build_skill(%{name: "Default Basic Skill"}),
         ultimate_skill: build_skill(%{name: "Default Ultimate Skill"})
       },
--- a/apps/champions/test/battle_test.exs
+++ b/apps/champions/test/battle_test.exs
@@ -390,5 +390,125 @@ defmodule Champions.Test.BattleTest do
       assert "team_1" ==
                Champions.Battle.Simulator.run_battle([unit], [target_dummy], maximum_steps: maximum_steps).result
     end
+
+    test "Execution-Heal", %{target_dummy: target_dummy} do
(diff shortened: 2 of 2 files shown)


apps/champions/test/support/test_utils.ex
```diff
@@ -13,7 +13,7 @@ defmodule Champions.TestUtils do
         class: "Assassin",
         base_attack: 100,
         base_health: 100,
-        base_defense: 100,
+        base_defense: 0,
```
agustinesco, line 16: why changing this?

lotuuu: It makes more sense for default values to be as "useless" as possible (skills that never get to be executed, for example). The neutral value for defense is 0 since it just leaves the damage received unmodified, so it's a more natural fit

agustinesco: nice

PR #559, opened 2024-05-02
Fix Modifiers

lotuuu:
Modifiers were working incorrectly: They were fetching the duration with a `"duration"` key instead of a `:duration` key. This was making all modifiers permanent (as it always defaulted to -1). We also were not transforming config duration ms to steps

Also, the units with "self" strategies were not defaulting to "random" correctly, as they did not define a `count` and a `targets_allies`. We also add these values.

# To Test

Run a battle with Grimstein and check that somewhere in the logs the following shows up (with different numbers):
`Modifier [Add damage_reduction by 0.2] remaining time reduced for Grimstein-790.`

### DIFF
--- a/apps/champions/docs/battle.md
+++ b/apps/champions/docs/battle.md
@@ -3,8 +3,8 @@ Sample code to create two teams of Mufluses and make them battle. Change [1..n] 
 ```
 {:ok, user_1} = Champions.Users.register("User1")
 {:ok, user_2} = Champions.Users.register("User2")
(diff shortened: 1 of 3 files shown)


tkz00 (changes requested):
Running a battle with 6 Grimstein's vs 6 Grimstein's I only get one ``dmg-red%`` modifier applied in the logs, which I find weird, but the error is that that modifier is never shown in the history of the battle.

tkz00 (approved):
Functional approve!

ncontinanza (approved):
LGTM!

PR #561, opened 2024-05-03
[GH-560] add currency endpoint

### LINKED ISSUE #560: Add gateway endpoint to add currency to user
part of #542 

Create a endpoint in the gateway application that communicates with the game backend application and add an X amount of N currency to a given user_id

agustinesco:
## Motivation

We need a way to modify users currencies through an endpoint

Closes #560

## Summary of changes

- Added endpoint to modify users currencies
- Refactored currencies code to consider game_id and not just name

## How to test it?

You need to get an user id , you can create an user in the backend with the following command:

```Elixir
{:ok, %{user: user}} = GameBackend.Users.find_or_create_google_user_by_email("a")


user.id
```

Copy the user id from the one generated and perform a put request to the gateway server to change the amount of currency, 

First you need to create a new json file called `currency.json` and add a map like this:

```json
{
  "currency_name": "Gold",
  "amount": 4
}
```

then you can copy paste the following command in a terminal:

```
curl -X PUT -H "Content-Type: application/json" -d @currency.json http://localhost:4001/curse/users/YOUR_USER_ID/currency
```

## Checklist


apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
```diff
@@ -0,0 +1,21 @@
...
+
+  alias GameBackend.Users.Currencies.UserCurrency
+  alias GameBackend.Users.Currencies
+
+  @moduledoc """
+    Controller to control currency changes in users
+  """
+
+  def modify_currency(conn, %{"currency_name" => currencty_name, "amount" => amount, "user_id" => user_id}) do
+    with {:value, {amount, _}} <- {:value, Integer.parse(amount)},
```
AminArria, line 12: This should not be needed, the JSON decoder in `:api` will take care of this

AminArria (changes requested):
Please make it work with a JSON payload like
```
{
  "user_id": "3847ec3e-b38d-4b55-82b1-0c2ea5b7d289",
  "currency_name": "Gold",
  "amount": 4
}
```

And this curl request
```
curl -X PUT -H "Content-Type: application/json" -d @currency.json http://localhost:4001/users/currency/modify
```

`currency.json` is the payload in a file

agustinesco:
> Please make it work with a JSON payload like
> 
> ```
> {
>   "user_id": "3847ec3e-b38d-4b55-82b1-0c2ea5b7d289",
>   "currency_name": "Gold",
>   "amount": 4
> }
> ```
> 
> And this curl request
> 
> ```
> curl -X PUT -H "Content-Type: application/json" -d @currency.json http://localhost:4001/users/currency/modify
> ```
> 
> `currency.json` is the payload in a file

I think with the current implementation it already accepts json files, at least i tested with the code you provided and it worked, the thing is that the integer parser is not working now, i just removed it

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
agustinesco: deleted!

AminArria (approved):

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Gateway.Controllers.Users.CurrencyController do
+  use Gateway, :controller
+
+  alias GameBackend.Utils
+  alias GameBackend.Users.Currencies.UserCurrency
+  alias GameBackend.Users.Currencies
+
+  @moduledoc """
```
lotuuu, line 8: Why the moduledoc here? It looks so weird 😂

apps/game_backend/lib/game_backend/users/currencies.ex
```diff
@@ -169,8 +185,28 @@ defmodule GameBackend.Users.Currencies do
...
+      |> add_currency(get_currency_by_name_and_game!(currency_name, game_id).id, amount)
+
+  @doc """
+  Add the specified amount of currency to an user by it's name
+
+  Returns nil if the currency doesn't exists
+
+  ## Examples
+
+      iex> add_currency_by_name_and_game(user_id, currency_name, amount)
```
lotuuu, line 200: Missed the new parameter here!

lotuuu (changes requested):
Left some dumb docs comments.

I have a question on the implementation of the fix on get_currency_by_name_and_game: why did we choose to add the game id as a parameter every time, instead of taking the game_id from the user (user.game_id)? Not saying it's wrong how we did it now, also I think it saves us a query, I just wanted to know if this was the reason or if there's something else I'm missing

lotuuu (changes requested):
I think we should also add a validation for the user id that's sent. If you send an unexisting user id now you get an elixir crash because of unexisting user_id fkey in new UserCurrency. While we are at it we could add more descriptive errors in "Couldn't add currency"

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
agustinesco: ahahhahahaha i didn't noticed! moved

apps/game_backend/lib/game_backend/users/currencies.ex
agustinesco: Whoops, fixed

apps/gateway/lib/gateway/router.ex
```diff
@@ -19,7 +19,13 @@ defmodule Gateway.Router do
     pipe_through :api
 
     get "/auth/:provider/token/:token_id", Controllers.AuthController, :validate_token
-    put "/users/:user_id", Controllers.UserController, :update
+
+    scope "/users" do
```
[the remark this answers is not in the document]
agustinesco: Yeah, it have sense! i'll move curse to a different scope and ensure that the user is from curse so we don't interact with other games

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
```diff
@@ -0,0 +1,25 @@
...
+  def modify_currency(conn, %{
+        "currency_name" => currencty_name,
+        "amount" => amount,
+        "user_id" => user_id,
+        "game_name" => game_name
+      }) do
+    game_id = Utils.get_game_id(String.to_atom(game_name))
+
+    case Currencies.add_currency_by_name_and_game(user_id, currencty_name, game_id, amount) do
+      {:ok, %UserCurrency{}} -> send_resp(conn, 200, "Currency added")
```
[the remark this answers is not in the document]
agustinesco: hmmm i don't know if better but we can do that yeah, i can't sent the whole struct here but i'll send some fields

agustinesco:
> I think we should also add a validation for the user id that's sent. If you send an unexisting user id now you get an elixir crash because of unexisting user_id fkey in new UserCurrency. While we are at it we could add more descriptive errors in "Couldn't add currency" (returning either "Currency not found" or "User not found")

It have sense to validate the user yes, we don't want to change champions users by mistake

agustinesco:
> Left some dumb docs comments.
> 
> I have a question on the implementation of the fix on get_currency_by_name_and_game: why did we choose to add the game id as a parameter every time, instead of taking the game_id from the user (user.game_id)? Not saying it's wrong how we did it now, also I think it saves us a query, I just wanted to know if this was the reason or if there's something else I'm missing

Not particular reason, it was just the way i implemented it, i'll use the game_id from the user instead of taking it from the url

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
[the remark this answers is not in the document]
lotuuu: Yeah I was thinking not even any more structues just a plain integer value

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
```diff
@@ -1,25 +1,29 @@
-defmodule Gateway.Controllers.Users.CurrencyController do
+defmodule Gateway.Curse.Controllers.Users.CurrencyController do
+  @moduledoc """
+    Controller to control currency changes in users
```
lotuuu, line 3:
```suggestion
  Controller to control currency changes in users
```

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
```diff
@@ -1,25 +1,29 @@
...
-
   def modify_currency(conn, %{
         "currency_name" => currencty_name,
         "amount" => amount,
-        "user_id" => user_id,
-        "game_name" => game_name
+        "user_id" => user_id
       }) do
-    game_id = Utils.get_game_id(String.to_atom(game_name))
+    {:ok, user} = Users.get_user(user_id)
```
lotuuu, line 17: This will cause a crash if the user we send doesn't exist. Let's do this inside the `with` block instead!

lotuuu (changes requested):

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
agustinesco: Sure thing

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
```diff
@@ -22,7 +22,7 @@ defmodule Gateway.Curse.Controllers.Users.CurrencyController do
            {:add_currency, Currencies.add_currency_by_name_and_game(user_id, currencty_name, user.game_id, amount)} do
       send_resp(conn, 200, Jason.encode!(Map.take(user_currency, [:amount, :user_id, :currency_id])))
     else
-      {:get_user, _} -> send_resp(conn, 400, "User not found")
+      {:get_user, _} -> send_resp(conn, 404, "User not found")
       {:curse_user, _} -> send_resp(conn, 400, "User from another game")
       {:add_currency, _} -> send_resp(conn, 400, "Couldn't add currency")
```
lotuuu, line 27:
```suggestion
      {:add_currency, _} -> send_resp(conn, 404, "Currency not found")
```

This is the only way this path can be reached! Makes for a clearer error :P

lotuuu (dismissed):

apps/gateway/lib/gateway/controllers/arena/currency_controller.ex
agustinesco: we can also have an error in the  user currency changeset, so i would let this to be more general

lotuuu: Right but now we cannot have an error in the UserCurrency changeset though, there's no other possible error because we're capturing the non-existing user beforehand. If you're afraid this might change in the future, that's valid, but I think it'd be a nice addition

agustinesco: Let's change it here and iterate if we ever validate currencies then

lotuuu (approved):

apps/champions/lib/champions/units.ex
```diff
@@ -118,7 +119,7 @@ defmodule Champions.Units do
   def calculate_level_up_cost(unit),
     do: [
       %CurrencyCost{
-        currency_id: Currencies.get_currency_by_name!("Gold").id,
+        currency_id: Currencies.get_currency_by_name_and_game!("Gold", Utils.get_game_id(:champions_of_mirra)).id,
```
tvillegas98, line 122: It is not necessary to address this comment here but I think we should put these strings into a module attribute

tvillegas98 (approved):

apps/champions/lib/champions/units.ex
agustinesco: I think this is a placeholder and i would not change this

PR #568, opened 2024-05-06
[GH-565] add trophies currency

### LINKED ISSUE #565: Develop thropies feature
Part of #542 

use matches to give trophies to users that win matches

The character that won first get 4 trophies
The character that won first get 2 trophies
The character that won first get 1 trophy

This should be handled by a new applicataion called curse_of_mirra, that will act as an intermediary between arena and game_backend

the role of the application of curse of mirra is to process messages coming from arena and interact with the game_backend application to insert/delte/update registries 

Game backend should ONLY have db operations

agustinesco:
## Motivation
When a match ends you should receive or loss x amount of trophies depending on which positions you ended in the match and your current amount of trophies

Closes #565

the levels and amounts especification can be found here: https://www.notion.so/lambdaclass/Ranking-System-c04164a466b64e27ba9f9c8e81d755c6

## Summary of changes
- Added position check to the match result struct
- Now every time a player dies the result will be stored in a map 

## How to test it?

1. Create a google user with: `{:ok, user} = GameBackend.Users.find_or_create_google_user_by_email("a")`
2. use the user id generated in 1 and play a match with the following link in the browser: `http://localhost:3000/board/YUOR_USER_ID/uma/jorge/join`
3. When the match the user that you created should receive x amount of trophies depending in how many you had previously and your position. The rules for this can be found in `apps/game_backend/priv/currencies_rules.json` and you can check how many trophies you have with the following command: `Repo.get_by(GameBackend.Users.GoogleUser, id: YOUR_USER_ID ) |> Repo.preload([user: [currencies: :currency]])`

## Checklist


agustinesco:
Changed base

apps/game_backend/priv/currencies_rules.json
```diff
@@ -0,0 +1,160 @@
+{
+    "rank": [
+        {
+            "maximum_rank": 50,
```
AminArria, line 4: Can you explain what `maximum_rank` means?

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -2,21 +2,48 @@ defmodule GameBackend.Matches do
...
-      end)
+    currency_config =
+      Jason.decode!(conrrency_config_json)
 
-    Repo.transaction(transaction)
+    Enum.reduce(results, Multi.new(), fn result, transaction_acc ->
+      changeset = ArenaMatchResult.changeset(%ArenaMatchResult{}, result)
+      {:ok, google_user} = Users.get_google_user(result["user_id"])
+
+      amount_of_trophies = Currencies.get_amount_of_currency_by_name(google_user.user.id, "Trophies")
```
AminArria, line 23: Not to fix in this PR, but is it expected/right to need that `google_user.user`? Shouldn't the currency be in the meta-user?

priv/repo/seeds.exs
```diff
@@ -41,6 +42,9 @@ units_per_level = 5
 {:ok, _fertilizer_currency} =
   Users.Currencies.insert_currency(%{game_id: champions_of_mirra_id, name: "Fertilizer"})
 
+{:ok, _trophies_currency} =
+  Users.Currencies.insert_currency(%{game_id: curse_of_mirra_id, name: "Trophies"})
```
AminArria, line 46: FYI I don't think the seeds run when this is deployed, so you will not have this unless manually added or part of the migration

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -2,21 +2,48 @@ defmodule GameBackend.Matches do
...
+    |> Repo.transaction()
+  end
+
+  def get_amount_of_trophies_to_modify(current_trophies, position, currencies_config) do
+    Enum.sort_by(currencies_config["rank"], fn %{"maximum_rank" => maximum} -> maximum end, :asc)
+    |> Enum.find(fn %{"maximum_rank" => maximum} ->
+      maximum > current_trophies
+    end)
+    |> case do
+      nil -> 0
```
AminArria, line 45: `nil` can't be a valid result, like this it will happen that if you have more than 9999 trophies you can no longer loose or win as it will always return 0 as the amount here

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -2,21 +2,48 @@ defmodule GameBackend.Matches do
...
 
-    transaction =
-      Enum.reduce(results, transaction, fn result, transaction_acc ->
-        attrs = Map.put(result, "match_id", match_id)
-        changeset = ArenaMatchResult.changeset(%ArenaMatchResult{}, attrs)
-        transaction_acc = Multi.insert(transaction_acc, {:insert, attrs["user_id"]}, changeset)
-        transaction_acc
-      end)
+    currency_config =
+      Jason.decode!(conrrency_config_json)
```
AminArria, line 17: Having to read and JSON parse on every request is not ideal, if we could read and process this config during compilation (in `config.exs`) or at startup we could have a nicer presentation and save some milliseconds on this

For example, we could end up saving the config as a map like
```elixir
%{
  :distributions => [50, 250, 9999],
  50 => %{1 => 10, 2 => 8, ...},
  250 => %{1 => 8, 2 => 6, ...},
  9999 => %{1 => 1, 2 => -4, ...},
  :infinity => %{1 => 0, 2 => -1, ...},
}
 ```
 
 So we have a `distributions` entry that lists the max amounts, then a key with that distribution which has a map where every position is a key. Also, we have the special case `:infinity` to handle distributions after whatever max amount we set

File: apps/game_backend/lib/game_backend/matches.ex [hunk 6]
```diff
@@ -2,21 +2,48 @@ defmodule GameBackend.Matches do
...
+        {:add_trophies_to, result["user_id"]},
+        fn _, _ ->
+          Currencies.add_currency_by_name!(google_user.user.id, "Trophies", amount)
+        end
+      )
+    end)
+    |> Repo.transaction()
+  end
+
+  def get_amount_of_trophies_to_modify(current_trophies, position, currencies_config) do
```
AminArria, line 39: Kinda tied to my comment on the pre-processing of the configuration, if we did that and have it in a easier format like that then we can simplify some of the logic here and have something like

```elixir
  def get_amount_of_trophies_to_modify(current_trophies, position) do
    trophies_config = Application.get_env(:arena, GameBackend.Currencies)[:trophies]
    distribution = Enum.find(trophies_config[:distributions], trophies_config[:infinity], fn max_amount -> max_amount > current_trophies end)
    get_in(trophies_config, [distribution, position])
  end
```

AminArria (changes requested):

apps/game_backend/lib/game_backend/matches.ex
agustinesco: Imo the currencies should be in the User, since it's different for each game, in the future we can have a feature to move currencies between users bellow the same meta-users so we can have gold in curse and move it to champions a rate

apps/arena/lib/arena/game_updater.ex
```diff
@@ -1377,6 +1386,15 @@ defmodule Arena.GameUpdater do
     end)
   end
 
+  defp put_player_position(%{positions: positions} = game_state, player_id) do
+    next_position = 10 - Enum.count(positions)
```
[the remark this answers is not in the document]
agustinesco: It's the `@clients_needed 10` from the game launcher, when we move that to a config file we should move this 10 too. Maybe adding a fixme is required right?

apps/game_backend/lib/game_backend/matches.ex
agustinesco: right, a fix could be having an "infinite" rank and default to that instead of 0

apps/arena/lib/arena/game_updater.ex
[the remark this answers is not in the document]
AminArria: Ah right, maybe add a simple config in `config.exs` for it?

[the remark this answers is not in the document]
agustinesco: sounds good! done

apps/game_backend/lib/game_backend/matches.ex
agustinesco: Yeah i didn't quite liked parsing the config file on every request, i didn't though about add it to the config file thx! i moved this to the `runtime.exs` file and we can simply read it from there! i keep the old system format tho do you think we should change it?

apps/game_backend/lib/game_backend/matches.ex
agustinesco: this was changed thanks to the previous comment!

priv/repo/seeds.exs
agustinesco: right! i didn't though about that. I'll do the same as what the champions team is doing, running the seed manually when this is deployed to the server

AminArria (dismissed):

priv/repo/seeds.exs
```diff
@@ -10,6 +10,7 @@ alias GameBackend.Units.Unit
 alias GameBackend.Users
 alias GameBackend.Users.KalineTreeLevel
 
+curse_of_mirra_id = 1
 champions_of_mirra_id = 2
```
Nico-Sanchez, line 14: Replace this with Utils function please 😄

config/runtime.exs
```diff
@@ -137,6 +137,11 @@ end
 ###################################
 # App configuration: game_backend #
 ###################################
+{:ok, conrrency_config_json} =
```
Nico-Sanchez, line 140: Conrrency. Con-who?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -491,6 +495,8 @@ defmodule Arena.GameUpdater do
   end
 
   defp report_game_results(state, winner_id) do
+    match_id = Ecto.UUID.generate()
```
Nico-Sanchez, line 498: You can remove this variable and just call generate/0 inside the map.

apps/arena/lib/arena/game_launcher.ex
```diff
@@ -68,15 +66,16 @@ defmodule Arena.GameLauncher do
     Process.send_after(self(), :launch_game?, 300)
     diff = System.monotonic_time(:millisecond) - state.batch_start_at
 
-    if length(clients) >= @clients_needed or (diff >= @start_timeout_ms and length(clients) > 0) do
+    if length(clients) >= Application.get_env(:arena, :clients_needed_in_match) or
```
Nico-Sanchez, line 69: Players instead of clients seems more accurate imo.
```suggestion
    if length(clients) >= Application.get_env(:arena, :players_needed_in_match) or
```

Nico-Sanchez (changes requested):
Just some minor comments and we're ready to go. Congrats 🚀

config/runtime.exs
agustinesco: 😂  fixed

apps/arena/lib/arena/game_updater.ex
agustinesco: detroyed

Nico-Sanchez (approved):

AminArria (approved):

PR #569, opened 2024-05-06
Remove unused Gateway modules

lotuuu:
## Motivation

We have these deprecated modules that were used in our previous HTML implementation. Now that we use websockets they are unused.

## Summary of changes

Remove controllers and utils

## How to test it?

Play around with the client. If nothing breaks in the elixir shell then we're good!


AminArria (dismissed):

AminArria (approved):

tvillegas98 (approved):
LGTM!

PR #575, opened 2024-05-06
[GH-322] Implement untargetable tag

ncontinanza:
## Motivation

We want some skills to leave their targets as Untargetables. That is, no other unit will be able to use their skills on them.

Closes https://github.com/lambdaclass/champions_of_mirra/issues/322

## Summary of changes

Before choosing the targets of a skill, filter the units that have the Untargetable tag.

## How to test it?

The easiest way to test this PR is to follow the steps in the test that is added, where a unit is always marked as untargetable and is never damaged since it receives the tag.

To test with the client, you can follow these steps:
1. Start the client and choose one of your units
2. Edit the `skills.json` file adding a tag component to one of your unit's skills. You should add something like this to an effect's `components` list
```
{
    "type": "ApplyTags",
    "tags": [
      "Untargetable"
    ]
}
```
3. Run `Champions.Config.import_skill_config()` in your terminal to update your unit's skills
4. Run a battle and check the log. You should see how the Untargetable tag is applied each time the skill you edited is used. Remember to change all Logger.info for IO.inspect to have the complete log.

## Checklist


apps/champions/test/battle_test.exs
```diff
@@ -514,6 +514,115 @@ defmodule Champions.Test.BattleTest do
     end
   end
 
+  describe "Components" do
+    test "Untargetable tag" do
+      # We will set up a battle between a unit with an untargetable skill and a unit that targets the nearest enemy, performing a DealDamage execution.
+      # The first unit will attack and tag the second unit as untargeteable each time it hits, so the second unit will only be hit once.
+      # The second unit is very weak, but the first unit will never be able to hit it, so the battle will end in a victory for the second unit.
+
+      backline_slot = 6
```
lotuuu, line 523: We don't need to set the slot, it doesn't change anything

lotuuu (changes requested):
The strategy you chose for the implementation is excellent, since we won't need to worry about implementing untargetable units for each target strategy.

I think it would be nice to add a test that checks that an attacking unit goes for the next one available if the one it would target was untargetable. I'm thinking something like:
- Units A and B vs unit C
- Unit C has one-shot power and gets to attack once. Unit B would be its target because it's nearest.
- Unit B's skill makes it untargetable before C hits it.
- Unit A also has one-shot power, but since its' cooldown is longer than C, it will die before getting to cast.
- Since A doesn't manage to kill C, the battle ends in a draw. If B didn't turn untargetable, it would end in Team1 win.

Of course you can think of your own test if you want to, but I think this scenario is going to be so likely during our battles that it deserves a test.

ncontinanza:
> The strategy you chose for the implementation is excellent, since we won't need to worry about implementing untargetable units for each target strategy.
> 
> I think it would be nice to add a test that checks that an attacking unit goes for the next one available if the one it would target was untargetable. I'm thinking something like:
> 
> * Units A and B vs unit C
> * Unit C has one-shot power and gets to attack once. Unit B would be its target because it's nearest.
> * Unit B's skill makes it untargetable before C hits it.
> * Unit A also has one-shot power, but since its' cooldown is longer than C, it will die before getting to cast.
> * Since A doesn't manage to kill C, the battle ends in a draw. If B didn't turn untargetable, it would end in Team1 win (C kills B, then A kills C).
> 
> Of course you can think of your own test if you want to, but I think this scenario is going to be so likely during our battles that it deserves a test.

I added a test with some similarities to the one you suggest, I think we'll be able to make a better one when the Self targeting strategy is implemented

apps/champions/test/battle_test.exs
```diff
@@ -566,7 +567,7 @@ defmodule Champions.Test.BattleTest do
         |> Characters.insert_character()
 
       {:ok, loser_unit} =
-        TestUtils.build_unit(%{character_id: loser_character.id, slot: 1}) |> Units.insert_unit()
+        TestUtils.build_unit(%{character_id: loser_character.id}) |> Units.insert_unit()
```
lotuuu:
```suggestion
      {:ok, loser_unit} = TestUtils.build_unit(%{character_id: loser_character.id}) |> Units.insert_unit()
```

lotuuu (changes requested):
Let's wait until #585 is merged and implement the test you wanted! It's only missing 1 review

lotuuu (approved):

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -520,16 +520,27 @@ defmodule Champions.Battle.Simulator do
...
 
+  defp choose_targets(caster, targeting_strategy, state) do
+    targeteable_units =
+      state.units
+      |> Enum.filter(fn {_, unit} -> not Enum.any?(unit.tags, &(&1.tag == "Untargetable")) end)
+
+    state_with_targeteable_units =
+      Map.put(state, :units, targeteable_units)
+
+    choose_targets_by_strategy(caster, targeting_strategy, state_with_targeteable_units)
```
agustinesco, line 531: Is it really necessary to include the targetable units inside the state? maybe it could be a new parameter of the `choose_targets_by_strategy`, just wondering

PR #578, opened 2024-05-07
[GH-525] Track match data 

### LINKED ISSUE #525: [FEATURE] Data Tracking System
**NOTE: See bottom [comment](https://github.com/lambdaclass/mirra_backend/issues/525#issuecomment-2105145623)**

## 🚀 Feature

Depends on lambdaclass/curse_of_mirra#905

We need to implement tracking of various events that happen inside and outside matches. Various subsystems will eventually rely on this feature: daily/weekly quests, events, and in the future: challenges/achievements. 

For now, the primary goal should be to figure out a proper architecture for tracking this information, i.e. how will the data be structured and how are events going to be triggered through the code, in a way that is versatile and expandable. 

**Some example stats for a basic system:**
- Kills (also storing killer and killed classes, and skill that was used) (also save if the kill is against a bot or player)
- Deaths (see above parentheses)
- Damage dealt
- Damage received
- Heal
- Times the user placed 1st... all the way to 10th
- Match count (per class and total)
- Time spent inside the game app
- Time spent inside matches
- Win-streak (negative number for losses streak)
- Highest kill-streak (streaks reset at every game)
- Crate abilities used

**Future use cases:**
- Game currencies and skin shards
- More specific/arbitrary stats for achievements and challenges
- Derivated stats (win rate%, time in combat%, K/D ratio, etc)

AminArria:
## Motivation
Closes #525 

We have to main reasons to track things that happen in-game during a match
1. Statistics and analysis for us (e.g. Why are people picking so much Muflus? Do the spam only certain skills? etc)
2. Post-match bonuses (e.g. daily quests, currency gains, etc)

## Summary of changes

Main change here is the introduction of `game_tracker.ex` this module will act as a collector of events, in a way its architected similar to how metrics/telemetry works and how their reporters works. Then is a matter of `game_updater.ex` emitting the events and properly tracking them, for now we are only emitting a kill event that keep tracks of the characters  you have killed and the character that killed you (saving your resulting position as well)

For this current implementation several decisions have been made that we I would like to make clear or talk about
- `GameTracker` is a single gen_server handling the tracking of all matches in the server, this might not scale, but its easily solvable by either having a pool of trackers or by having a dedicated tracker per match. Let's measure first before changing this
- Error handling for reporting of match results we should have a retry mechanism to prevent players from loosing game results, eventually this could be some "persistent" storage locally or another DB directly connected. For now this will be addressed in a follow-up #601 

## How to test it?

To test you can run a game through the browser (`game_client`) by going to [localhost:3000](http://localhost:3000/) and in the console see the result after the match ends

```
GameBackend.Repo.all(GameBackend.Matches.ArenaMatchResult)
```

## Checklist


apps/arena/lib/arena/game_updater.ex
```diff
@@ -342,6 +355,12 @@ defmodule Arena.GameUpdater do
 
     broadcast_player_dead(state.game_state.game_id, victim_id)
 
+    GameTracker.push_event(
+      self(),
+      {:kill, %{id: killer.id, character_name: killer.aditional_info.character_name},
+       %{id: victim.id, character_name: victim.aditional_info.character_name}}
+    )
+
```
Nico-Sanchez, line 363: Why the change of storing the whole match at the end to send every kill event in time?

apps/arena/lib/arena/game_tracker.ex
```diff
@@ -0,0 +1,118 @@
...
+    GenServer.cast(__MODULE__, {:push_event, match_pid, event})
+  end
+
+  ##########################
+  # Callbacks
+  ##########################
+  @impl true
+  def init(_) do
+    state = %{matches: %{}}
+    {:ok, state}
```
Nico-Sanchez, line 34:
```suggestion
    {:ok, %{matches: %{}}
```

apps/arena/lib/arena/game_tracker.ex
```diff
@@ -0,0 +1,118 @@
+defmodule Arena.GameTracker do
+  @moduledoc """
+  This module is in charge of tracking the different game changes that we consider
+  datapoints and use for quests, analytics, etc.
+
+  `GameTracker` will behave similar to a metrics collector, but with a push model rather than pull.
+  Games will push the data to it and every X time it will push the data to Gateway for storing and updating quests
```
Nico-Sanchez, line 7: Every X time? I think every time a match ends, it sends the data to Gateway, which is desirable.

Nico-Sanchez (changes requested):
I left a few questions.
Also, can you create issues for every todo in the code? Then please put the link in the TODO tag in code.

apps/arena/lib/arena/game_updater.ex
AminArria: Essentially we keep doing that, but rather than `GameUpdater` keeping track of it we have this specialised module for it. So now we push this events towards it instead of saving it in the game state, a benefit of this pushing is that we can order things as they happen

The main reason for the change was that I had made for when this stats updates where meant to be sent every X seconds during a game, not only at the end. Afterwards I still liked the design cause it allows to decouple things a bit and in the case when `gateway` is down/unabe to process the request we will have the retry logic here. Yes, we could have it in `GameUpdater` instead, but that process is "heavy" as it contains a lot of state so keeping it around only for retries could be problematic for the resource consumptions of the server

apps/arena/priv/config.json
```diff
@@ -1190,7 +1190,7 @@
       "active": true,
       "base_speed": 27.0,
       "base_size": 110.0,
-      "base_health": 440,
+      "base_health": 44,
```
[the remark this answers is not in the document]
AminArria: Done [6f72d52](https://github.com/lambdaclass/mirra_backend/pull/578/commits/6f72d529f1d4ed852a9f01d124c3c86cef842814)

apps/arena/lib/arena/game_tracker.ex
AminArria: Yep, fixed the words. They are from before changing to only the final update

apps/arena/lib/arena/game_updater.ex
Nico-Sanchez: Perfectly explained. Thank you 😄

Nico-Sanchez (approved):

apps/arena/lib/arena/game_tracker.ex
```diff
@@ -0,0 +1,118 @@
...
+  def handle_cast({:push_event, match_pid, event}, state) do
+    state = update_in(state, [:matches, match_pid], fn data -> update_data(data, event) end)
+    {:noreply, state}
+  end
+
+  defp update_data(data, {:kill, killer, victim}) do
+    data
+    |> update_in([:players, killer.id, :kills], fn kills -> kills ++ [victim.character_name] end)
+    |> put_in([:players, victim.id, :death], killer.character_name)
+    |> put_in([:players, victim.id, :position], data.position_on_death)
```
agustinesco, line 88: this kinda invalidate the `put_player_position/2` logic in the updater right? i'll create an issue to delete it

agustinesco: Created #615

Issue #581, opened 2024-05-08
[FEATURE] Daily quest functionality


manucamejo:
- [x] Random N quests per day -> #590
- [x] Be able to change a quest by paying a currency -> #624
- [x] Give reward and close daily quest when daily quest is achieved -> #633

PR #590, opened 2024-05-10
[GH-581] implement daily quest feature

agustinesco:
## Motivation

*Explain why not how.*
Closes #issue_number

## Summary of changes

*List your changes here.*

## How to test it?

## Checklist

### DIFF
--- a/apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
+++ b/apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
@@ -1,0 +1,61 @@
+defmodule GameBackend.CurseOfMirra.Quests do
+  @moduledoc """
+
+  """
+  alias GameBackend.Repo
+  alias GameBackend.Quests.QuestDescription
+  alias GameBackend.Matches.ArenaMatchResult
+  alias Ecto.Multi
+  import Ecto.Query
+
+  defmacrop custom_where_macro(queryable, binding, field_name, operator, value) do
+    {:where, [],
+     [
+       queryable,
+       [
+         {{:^, [], [binding]}, {:relation, [], Elixir}}
+       ],
+       {operator, [context: Elixir, import: Kernel],
+        [
+          {:field, [], [{:relation, [], Elixir}, {:^, [], [field_name]}]},
+          {:^, [], [value]}
+        ]}
+     ]}
+  end
+
+  def upsert_quests(quests_params) do
+    {multi, _acc} =
+      Enum.reduce(quests_params, {Multi.new(), 0}, fn quest_param, {multi, acc} ->
+        quest_changeset = QuestDescription.changeset(%QuestDescription{}, quest_param)
+
+        multi = Multi.insert(multi, {:insert_quest, acc}, quest_changeset)
+
+        {multi, acc + 1}
+      end)
+
+    Repo.transaction(multi)
+  end
+
+  def build_query_from_quest_description(%QuestDescription{} = quest_description) do
+    base_query =
+      from(amr in ArenaMatchResult,
+        as: :amr
+      )
+
+    Enum.reduce(quest_description.quest_objectives, base_query, fn quest_objective, query ->
+      type = String.to_atom(quest_objective["type"])
+      comparator = quest_objective["comparison"] |> parse_comparator()
+      value = quest_objective["value"]
+      custom_where(query, :amr, type, comparator, value)
+    end)
+  end
+
+  defp parse_comparator("equal"), do: :==
+  defp parse_comparator("greater"), do: :>
+
+  for operator <- [:!=, :<, :<=, :==, :>, :>=, :ilike, :in, :like] do
+    defp custom_where(queryable, binding, field_name, unquote(operator), value) do
+      custom_where_macro(queryable, binding, field_name, unquote(operator), value)
+    end
+  end
+end
--- a/apps/game_backend/lib/game_backend/quests/quest_description.ex
+++ b/apps/game_backend/lib/game_backend/quests/quest_description.ex
@@ -1,0 +1,28 @@
+defmodule GameBackend.Quests.QuestDescription do
+  @moduledoc """
+
+  """
+
+  use GameBackend.Schema
+  import Ecto.Changeset
+
+  schema "quest_descriptions" do
+    field(:description, :string)
+    field(:quest_objectives, {:array, :map})
+
+    timestamps()
+  end
+
+  @required [
+    :description,
+    :quest_objectives
+  ]
+
+  @permitted [] ++ @required
+
+  def changeset(changeset, attrs) do
+    changeset
+    |> cast(attrs, @permitted)
+    |> validate_required(@required)
+  end
+end
--- a/apps/game_backend/lib/game_backend/quests/user_daily_quest.ex
+++ b/apps/game_backend/lib/game_backend/quests/user_daily_quest.ex
@@ -1,0 +1,40 @@
+defmodule GameBackend.Quests.UserDailyQuest do
+  @moduledoc """
+
+  """
+
+  use GameBackend.Schema
+  import Ecto.Changeset
+
+  alias GameBackend.Quests.QuestDescription
+  alias GameBackend.Users.User
+
+  schema "user_daily_quest" do
+    field(:target, :integer)
+    field(:progress, :integer)
+
+    belongs_to(
+      :quest_description,
+      QuestDescription
+    )
+
+    belongs_to(:user, User)
+
+    timestamps()
+  end
+
+  @required [
+    :quest_description_id,
+    :user_id,
+    :target,
+    :progress
+  ]
+
+  @permitted [] ++ @required
+
+  def changeset(changeset, attrs) do
+    changeset
+    |> cast(attrs, @permitted)
+    |> validate_required(@required)
+  end
+end
--- a/apps/game_backend/priv/curse_of_mirra/quests_descriptions.json
+++ b/apps/game_backend/priv/curse_of_mirra/quests_descriptions.json
@@ -1,0 +1,22 @@
+[
+    {
+        "description": "Tests description",
+        "quest_objectives": [
+            {
+                "type": "kills",
+                "value": 3,
+                "comparison": "greater"
+            },
+            {
+                "type": "character",
+                "value": "muflus",
+                "comparison": "equal"
+            },
+            {
+                "type": "result",
+                "value": "abandon",
+                "comparison": "equal"
+            }
+        ]
+    }
+]
--- a/apps/game_backend/priv/repo/migrations/20240509225708_add_quest_description_table.exs
+++ b/apps/game_backend/priv/repo/migrations/20240509225708_add_quest_description_table.exs
@@ -1,0 +1,11 @@
+defmodule GameBackend.Repo.Migrations.AddQuestDescriptionTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:quest_descriptions) do
+      add :description, :string, default: ""
+      add :quest_objectives, {:array, :map}
+      timestamps()
+    end
+  end
+end
--- a/apps/game_backend/priv/repo/migrations/20240509225945_add_user_daily_quest.exs
+++ b/apps/game_backend/priv/repo/migrations/20240509225945_add_user_daily_quest.exs
@@ -1,0 +1,13 @@
+defmodule GameBackend.Repo.Migrations.AddUserDailyQuest do
+  use Ecto.Migration
+
+  def change do
+    create table(:user_daily_quest) do
+      add :user_id, references(:users, on_delete: :delete_all), null: false
+      add :quest_description_id, references(:quest_descriptions, on_delete: :delete_all), null: false
+      add :target, :integer
+      add :progress, :integer
+      timestamps()
+    end
+  end
+end
--- a/apps/game_backend/lib/game_backend/curse_of_mirra/config.ex
+++ b/apps/game_backend/lib/game_backend/curse_of_mirra/config.ex
@@ -1,0 +1,16 @@
+defmodule GameBackend.CurseOfMirra.Config do
+  @moduledoc """
+
+  """
+
+  alias GameBackend.CurseOfMirra.Quests
+
+  def import_quest_descriptions_config() do
+    {:ok, skills_json} =
+      Application.app_dir(:game_backend, "priv/curse_of_mirra/quests_descriptions.json")
+      |> File.read()
+
+    Jason.decode!(skills_json, [{:keys, :atoms}])
+    |> Quests.upsert_quests()
+  end
+end


apps/game_backend/lib/game_backend/quests/user_daily_quest.ex
```diff
@@ -0,0 +1,40 @@
...
+  alias GameBackend.Users.User
+
+  schema "user_daily_quest" do
+    field(:target, :integer)
+    field(:progress, :integer)
+
+    belongs_to(
+      :quest_description,
+      QuestDescription
+    )
```
AminArria, line 19: Please put this in one line

apps/game_backend/lib/game_backend/quests/quest_description.ex
```diff
@@ -0,0 +1,34 @@
...
+
+  schema "quest_descriptions" do
+    field(:description, :string)
+    field(:type, :string)
+    field(:quest_objectives, {:array, :map})
+
+    timestamps()
+  end
+
+  @types ["daily"]
```
AminArria, line 17: Unless we actually use this outside the module don't do this and put it directly in `validate_inclusion/3`

AminArria (changes requested):

apps/game_backend/priv/repo/migrations/20240509225945_add_user_daily_quest.exs
```diff
@@ -0,0 +1,13 @@
+defmodule GameBackend.Repo.Migrations.AddUserDailyQuest do
+  use Ecto.Migration
+
+  def change do
+    create table(:user_daily_quest) do
```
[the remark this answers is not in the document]
agustinesco: Done!

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_description_table.exs
```diff
@@ -0,0 +1,12 @@
+defmodule GameBackend.Repo.Migrations.AddQuestDescriptionTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:quest_descriptions) do
```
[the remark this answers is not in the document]
agustinesco: Done!

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_description_table.exs
```diff
@@ -0,0 +1,12 @@
...
+      add :description, :string, default: ""
+      add :type, :string
+      add :quest_objectives, {:array, :map}
```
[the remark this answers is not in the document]
agustinesco: postgres ends up using _jsonb to save this on the db, i checked this using dbeaver:

image


image

apps/game_backend/lib/game_backend/quests/user_daily_quest.ex
agustinesco: Done!

apps/game_backend/lib/game_backend/quests/quest_description.ex
agustinesco: not used at all, removed

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_description_table.exs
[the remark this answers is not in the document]
AminArria: Cool cool, that's good

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -0,0 +1,111 @@
...
+
+        multi = Multi.insert(multi, {:insert_quest, acc}, quest_changeset)
+
+        {multi, acc + 1}
+      end)
+
+    Repo.transaction(multi)
+  end
+
+  def build_query_from_quest_description(%Quest{} = quest_description, user_id) do
```
AminArria, line 40: Need to understand a bit more why we need this and how we are going to use it, is the idea here to make 1 query per user per quest? If yes, then we are talking 30 queries to the DB (assuming 3 daily quests) per game ending. Honestly I'm preemptively worried about how this will perform under load

What do you think?

agustinesco: yeah of course! i was testing different appoaches of a result , now i have something that's worth looking :P

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_table.exs
```diff
@@ -0,0 +1,19 @@
+defmodule GameBackend.Repo.Migrations.AddQuestTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:quests) do
+      add :description, :string, default: ""
```
AminArria, line 6: Don't use `default: ""` if there is no string then `nil` should be the correct value. Having this `default` could cause a quest to be saved with `descriptor: ""` and then prevent modifications to other fields (not touching `descriptor`) due to our Ecto schema having `validate_required(..., [..., :descriptor, ...])`

Why do you wanted `default: ""`?

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_table.exs
```diff
@@ -0,0 +1,19 @@
...
+      add :type, :string
+      add :config_id, :integer
+      add :target, :integer
+      add :objective,  :map
+      add :reward, :map
```
AminArria, line 11: Looking at the config seems it could be 2 fields here `reward_currency_id` and `reward_amount`, why not that?

apps/game_backend/priv/curse_of_mirra/quests_descriptions.json
```diff
@@ -0,0 +1,45 @@
...
+        "config_id": 1,
+        "description": "Win three matches with Muflus",
+        "type": "daily",
+        "objective": {
+            "field": "result",
+            "value": 3,
+            "comparison": "greater",
+            "scope": "match"
+        },
+        "conditions": [
```
AminArria, line 12: Can you explain a bit on the distinction between `objective` and `conditions`? Like why can't we have the `objective` as one more condition?

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -0,0 +1,164 @@
...
+  defp parse_comparator("lesser"), do: :<
+  defp parse_comparator("lesser_or_equal"), do: :<=
+  defp parse_comparator(comparator), do: raise("Comparator not implemented yet #{comparator}")
+
+  defp acumulate_objective_progress_by_scope("day", value), do: value
+  defp acumulate_objective_progress_by_scope("match", _value), do: 1
+
+  for operator <- [:!=, :<, :<=, :==, :>, :>=, :in] do
+    defp custom_comparator(unquote(operator), first_value, second_value) do
+      custom_comparator_macro(unquote(operator), first_value, second_value)
```
AminArria, line 135: Feels like you are overcomplicating this with a macro and metaprogramming when functional things should be more than enough

You should leverage functional programming a bit
```elixir
  defp parse_comparator("equal"), do: &Kernel.==/2
  defp parse_comparator("distinct"), do: &Kernel.!=/2
  defp parse_comparator("greater_or_equal"), do: &Kernel.>=/2

  defp completed_daily_quest?(%DailyQuest{quest: %Quest{} = quest}, arena_match_results) do
    progess =
      Enum.filter(arena_match_results, fn arena_match_result ->
        Enum.all?(quest.conditions, fn condition ->
          ## Replace
          type = String.to_atom(condition["field"])
          comparator = condition["comparison"] |> parse_comparator()
          value = condition["value"]
          arena_match_result_value = Map.get(arena_match_result, type)
          custom_comparator(comparator, arena_match_result_value, value)
          
          ## With
          value = condition["value"]
          arena_match_result_value = Map.get(arena_match_result, type)
          comparator = parse_comparator(condition["comparison"])
          comparator.(arena_match_result_value, value)
        end)
      end)
      |> ...
  end
```

AminArria (changes requested):

apps/game_backend/priv/curse_of_mirra/quests_descriptions.json
agustinesco: Sure! the objective will determine the amount of progress you need to have in order to consider a quest completed, the conditions are a set of comparators to see if a match result is a valid candidate to contribute to quest completion

> Like why can't we have the objective as one more condition?
About this, how would define progression in a quest if everything is an objective? this was the solution i came up with but im open to ideas

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_table.exs
agustinesco: This was force of habit, we don't need it at all

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_table.exs
agustinesco: Is there any benefit of having this columns over a map?

apps/game_backend/priv/repo/migrations/20240425171037_add_arena_match_results.exs
```diff
@@ -3,7 +3,7 @@ defmodule GameBackend.Repo.Migrations.AddArenaMatchResults do
 
   def change do
     create table(:arena_match_results) do
-      add :user_id, references(:google_users, on_delete: :delete_all), null: false
+      add :google_user_id, references(:google_users, on_delete: :delete_all), null: false
```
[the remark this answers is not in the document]
agustinesco: Whoops, yeah this was something i added to tests this and was left here 😂  created his proper migration changing this! also i wonder if is worth to rename this or change it to belong to a user instead of a google user, what do you think?

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: Yeah i'm overcomplicating things, this started by creating a macro to use a custom Ecto.where  and was left to use kernel comparitions, i really like your suggestion! i didn't know you could do that, i'll be moving this to a different pr tho since this one scalated too much

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_table.exs
```diff
@@ -0,0 +1,19 @@
+defmodule GameBackend.Repo.Migrations.AddQuestTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:quests) do
+      add :description, :string
```
AminArria, line 6: If you happen to make other changes, would be good to put `null: false` in the columns here otherwise no big deal we can do that later

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -0,0 +1,98 @@
...
+
+  """
+  def change_quest(quest, attrs) do
+    Quest.changeset(quest, attrs)
+  end
+
+  @doc """
+  Insert or update config quests present in the "quests_descriptions.json" file
+  """
+  def upsert_quests(quests_params) do
```
AminArria, line 58: If you feel curious you can try the actual [Ecto Upsert](https://hexdocs.pm/ecto/constraints-and-upserts.html#upserts) rather than implement your own, but full disclosure I have never used it so no idea if it will work, but would be a nice improvement here if it does

AminArria (approved):
LGTM, left some comments, but nothing that needs immediate solutions just nice to have, but we can tackle later (if possible create an issue about them before merging)

apps/game_backend/priv/repo/migrations/20240509225708_add_quest_table.exs
agustinesco: I had to fix some conflicts so i'll do this

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: hmmm if we use the Ecto upsert we should update every other upsert already implemented in the repo, maybe we could do this change in a proper pr

Nico-Sanchez (changes requested):
I ran `GameBackend.CurseOfMirra.Quests.add_quest_to_user_id/3` with an amount greater than 3 and it always insert 3. It should return error and just insert the specified amount.
Also, please run a spell checker for the documentation (quests.md) because I found several typos.

apps/game_backend/lib/game_backend/quests/daily_quest.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  use GameBackend.Schema
+  import Ecto.Changeset
+
+  alias GameBackend.Quests.Quest
+  alias GameBackend.Users.User
+
+  schema "daily_quest" do
+    field(:completed_at, :utc_datetime)
+    field(:completed, :boolean, default: false)
```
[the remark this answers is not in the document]
agustinesco: Sure, this will reduce redundancy 💯

apps/game_backend/docs/configuration/quests.md
```diff
@@ -0,0 +1,21 @@
...
+
+Quests are a set of conditions that the players need to meet by playing matches in order to get rewards (for the time being currencies) you can specify an objective that's the final goal
+
+## Configuration
+
+- `config_id`: Unique id for the quests in the config file,
+- `description`: A description about the quest,
+- `type`: The duration of the quest, possible values can be ["daily"] 
+- `objective`: What the player need to achieve in order to take the quest as complete (Can be left empty)
+  - `field`: The field of the arena match result table that will be used as target
```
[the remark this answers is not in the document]
agustinesco: Sure thing

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -22,17 +22,30 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+    end_of_date = NaiveDateTime.end_of_day(naive_today)
+
     q =
-      from(qd in Quest,
-        where: qd.type == ^type
+      from(q in Quest,
+        left_join: dq in DailyQuest,
+        on: q.id == dq.quest_id and dq.user_id == ^user_id,
+        where:
+          (is_nil(dq) or dq.inserted_at < ^start_of_date or dq.inserted_at > ^end_of_date or not is_nil(dq.completed_at)) and
```
Nico-Sanchez, line 46: Documentation says the daily quests shouldn't be completed but here you're checking that the `:completed_at` field is not nil, that means that the quest has been completed.

Nico-Sanchez (changes requested):

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: that's right, the thing of this is that we're negating the valid quests:
`that doesn't have a valid daily quest`
so we're looking for quests from other days or completed quests

Nico-Sanchez (approved):

AminArria (approved):

PR #605, opened 2024-05-10
Fix route to currencies json in runtime.ex

agustinesco:
When we deployed the server the runtime file is not causing the application to crash since we're not always raising every application

succesful deploy runs: 
Arena: https://github.com/lambdaclass/mirra_backend/actions/runs/9038004110
Central Game Backend: https://github.com/lambdaclass/mirra_backend/actions/runs/9038138173

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -138,7 +138,7 @@ end
 # App configuration: game_backend #
 ###################################
 
-if System.get_env("RELEASE") == :central_backend or config_env() == :dev do
+if System.get_env("RELEASE") == "central_backend" or config_env() == :dev do
   {:ok, currency_config_json} =
     "./apps/game_backend/priv/currencies_rules.json"
     |> File.read()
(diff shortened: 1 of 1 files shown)


config/runtime.exs
```diff
@@ -137,11 +137,14 @@ end
 ###################################
 # App configuration: game_backend #
 ###################################
-{:ok, currency_config_json} =
-  Application.app_dir(:game_backend, "priv/currencies_rules.json")
-  |> File.read()
 
-config :game_backend, :currencies_config, Jason.decode!(currency_config_json)
+if System.get_env("RELEASE") == :central_backend or config_env() == :dev do
```
tvillegas98, line 141: `:central_backend` should be an string, shouldn't it?

agustinesco: you're right! the deploy didn't failed because it didn't ran at all :P

Nico-Sanchez (approved):
This is ok for now. But let's keep an eye to this for future configurations https://elixirforum.com/t/import-runtime-config-from-apps-in-umbrella-app-root/37662

tvillegas98 (approved):

PR #608, opened 2024-05-13
[GH-607] Only spawn bots if player is alive

### LINKED ISSUE #607: [Bug] Dead players are being controlled by bots

agustinesco:
## Motivation

We were spawning bots that could move when a dead player leaves a match.

Closes #607

## Summary of changes

- Now we'll only spawn bots if the player is alive, we're updating the enable field when a player dies so we'll use that

## How to test it?

1. Change the players_needed_in_match variable to 3
2. Join a match with the browser client and the unity client
3. Die to the spawned bot in the unity client and leave the match
4. In main you'll see that the dead players is moving
5. In this branch that shouldn't happen

## Checklist

### DIFF
--- a/apps/arena/lib/arena/game_socket_handler.ex
+++ b/apps/arena/lib/arena/game_socket_handler.ex
@@ -153,7 +153,7 @@ defmodule Arena.GameSocketHandler do
   end
 
   @impl true
-  def terminate(_reason, _req, %{game_finished: false} = state) do
+  def terminate(_reason, _req, %{game_finished: false, enable: true} = state) do
     spawn(fn ->
       Finch.build(:get, Utils.get_bot_connection_url(state.game_id, state.client_id))
       |> Finch.request(Arena.Finch)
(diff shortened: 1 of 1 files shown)


apps/arena/lib/arena/game_socket_handler.ex
```diff
@@ -153,7 +153,7 @@ defmodule Arena.GameSocketHandler do
   end
 
   @impl true
-  def terminate(_reason, _req, %{game_finished: false} = state) do
+  def terminate(_reason, _req, %{game_finished: false, enable: true} = state) do
```
Nico-Sanchez, line 156: Let's change this to check if the player is alive. That way if we change the `:enable` functionality we don't break this.

Nico-Sanchez (changes requested):

apps/arena/lib/arena/game_socket_handler.ex
agustinesco: Sounds good

Nico-Sanchez (approved):

BertovDev (approved):

PR #609, opened 2024-05-13
[GH-601] Track more match data

### LINKED ISSUE #601: [FEATURE] Tracking more data
This is the continuation of #525 

This time we should do 2 main things
- Add retry mechanism when reporting match results (see TODO in `game_tracker.ex`)
- Track more match data, a non exhaustive list and ideas
  - What characer killed you and if it was a bot or human
  - Damage dealt
  - Damage received
  - Heal
  - Times the user placed 1st... all the way to 10th
  - Crate abilities used
  - Skills used
- (optional, we can do more one follow up) Design a tracking of meta data, non exhaustive list
  - Time spent inside the game app
  - Time spent inside matches
- (optional) Any other TODO in `game_tracker.ex`

AminArria:
## Motivation

Closes #601 

## Summary of changes

- [x] Rebase after #578 is merged
- [x] Retry mechanism when reporting result
- More things to track
  - [x] Damage done
  - [x] Damage taken
  - [x] Health healed
  - [x] What character killed you
  - [x] Killed by human or bot?
  - [x] Match duration

## How to test it?

To test you can run a game through the browser (`game_client`) by going to [localhost:3000](http://localhost:3000/) and login or if you don't have the Google credentials you can create a user `GameBackend.Users.find_or_create_google_user_by_email("tester")` and then go to `http://localhost:3000/board/<YOUR_USER_ID>/muflus/player_name/quick_game`

Then after the match ends you can go to the console and see the results by doing

```
GameBackend.Repo.all(GameBackend.Matches.ArenaMatchResult)
```

## Checklist


manucamejo (approved):

apps/arena/lib/arena/game_updater.ex
```diff
@@ -353,11 +351,17 @@ defmodule Arena.GameUpdater do
 
     broadcast_player_dead(state.game_state.game_id, victim_id)
 
-    GameTracker.push_event(
-      self(),
-      {:kill, %{id: killer.id, character_name: killer.aditional_info.character_name},
-       %{id: victim.id, character_name: victim.aditional_info.character_name}}
-    )
+    case Map.get(game_state.players, killer_id) do
```
tvillegas98, line 354: This is the fix for the killing zone bug, right?

apps/game_backend/priv/repo/migrations/20240513180434_add_more_match_results_data.exs
```diff
@@ -0,0 +1,33 @@
...
+      add :killed_by, :string
+      add :killed_by_bot, :boolean
+      add :duration_ms, :integer
+    end
+
+    ## This is to allow us to later set `null: false` on all the new columns
+    ## without having to set a default value when adding the column
+    ## Having a default value in the DB could lead to us having a hidden bug or wrong values if
+    ## it where to be used by accident
+    query = """
```
tvillegas98, line 18: Help me understand this 😄 . This is only for previous records, right?

apps/arena/lib/arena/game_updater.ex
AminArria: Yes, this commit specifically [a6c7000](https://github.com/lambdaclass/mirra_backend/pull/609/commits/a6c70007dca97f389428662604e6513cb481ee79)

apps/game_backend/priv/repo/migrations/20240513180434_add_more_match_results_data.exs
AminArria: Yep exactly, I want to enforce `null: false` on this columns, but I can't do it if they are previous records unless I put a default value which I don't want to as mentioned in the comment

So this query essentially sets some one-time default values for existing records so then after we can enforce `null: false` on the new columns

tvillegas98: Thanks for the explanation!

tvillegas98 (approved):
Nice work!

Issue #431, opened 2024-03-27
Send all skill config fields to the client


BertovDev:
We have the problem that not all the skill config fields are send to the client, so in the moment we try for example to draw the skill indicators we miss some values

for example:
If we want to draw the dash indicator we would need

- Speed
- Duration

and this values are not been send, this happens the same for the projectiles.

PR #617, opened 2024-05-14
[GH-431] Add currency cost to levels

ncontinanza:
Closes: https://github.com/lambdaclass/champions_of_mirra/issues/431

Associated PR: https://github.com/lambdaclass/champions_of_mirra/issues/463

## Motivation

We need to be able to charge the user with a currency cost in order to get access to fight some of the levels. For example, the Dungeon levels will have a cost in Shovels.

## Summary of changes
- Add an embedded array of CurrencyCost to the Level schema.
- Before starting a battle, check that the user can afford it.
- Substract the level's cost in currency when fighting the battle.

## How to test it?

Run the backend and start the client in the main branch. For starters, everything should be working as usual, when none of the existing levels have a CurrencyCost associated.

Add this line to the map used for level creation in seeds.ex: `currency_costs: [%CurrencyCost{currency_id: gold_currency.id, amount: 13}],`
Also, you'll need to add this alias to that file: `alias GameBackend.Users.Currencies.CurrencyCost`

You should end up with something like this:

```elixir
levels =
  Enum.flat_map(Enum.with_index(rules, 1), fn {campaign_rules, campaign_index} ->
    {_, campaign} =
      Campaigns.insert_campaign(
        %{
          game_id: champions_of_mirra_id,
          super_campaign_id: super_campaign.id,
          campaign_number: campaign_index
        },
        returning: true
      )

    Enum.map(1..campaign_rules.length, fn level_index ->
      %{
        game_id: champions_of_mirra_id,
        campaign_id: campaign.id,
        level_number: level_index,
        experience_reward: 100 * level_index,
        attempt_cost: [%CurrencyCost{currency_id: gold_currency.id, amount: 13}],
        inserted_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second),
        updated_at: NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
      }
    end)
  end)
```

Now all the levels cost 13 Gold. Before fighting a new level, check your current amount of gold. If you fight the level and lose the battle, your amount of gold should have decreased by 13. If you won, your gold amount should now be `previous amount + level reward - level cost`.

## Checklist


priv/repo/seeds.exs
```diff
@@ -251,7 +252,7 @@ currency_rewards =
 
 currency_rewards =
   currency_rewards ++
-    Enum.map(levels_without_units, fn level ->
+    Enum.map(levels_without_embeds, fn level ->
```
lotuuu, line 255: Let's fix this double iteration while we're at it! We can use a [flat_map](https://hexdocs.pm/elixir/1.12/Enum.html#flat_map/2)

apps/serialization/gateway.proto
```diff
@@ -297,7 +297,8 @@ syntax = "proto3";
     uint32 level_number = 3;
     repeated Unit units = 4;
     repeated CurrencyReward currency_rewards = 5;
-    uint32 experience_reward = 6;
+    uint32 experience_reward = 6;    
+    repeated CurrencyCost cost = 7;
```
lotuuu, line 301: The names of this variable in the schema/DB and Protobuf messages differ, this will probably break. Also, let's use a more descriptive name like "attempt_cost" or "play_cost" to explain better what this represents!

lotuuu (changes requested):

lotuuu (dismissed):
Just as a nitpicky comment, I like the uncountable version of `cost` better for the cost of playing a level. Instead of `attempt_costs`, i'd go for `attempt_cost`.

Also, the description is now outdated as it uses the previous "`currency_costs`" attribute.

Lastly, is there an issue already in the client for implementing handling of "cant afford" errors? Because of course right now this breaks in the client

tkz00 (approved):

lotuuu (approved):

PR #620, opened 2024-05-14
Fix gateway.proto

lotuuu:
## Motivation

We introduced a change that was breaking tests, but since we forgot to run make generate-protos we didn't notice it until someone else ran it for a different PR.

Since this id is needed for tests, it makes no sense to remove it, so we will restore the changes.


Nico-Sanchez (approved):

agustinesco (approved):
LGTM!

PR #625, opened 2024-05-15
[GH-624] Endpoint to reroll quests

### LINKED ISSUE #624: [Feature] Add end point to reroll quests
Add an endpoint to reroll a daily_quest at the cost of a configurable currency

agustinesco:
## Motivation

Added endpoint to reroll a daily quest
you must have enough currency to reroll quests and each time it will be more expensive

This pr will have ambiguous endpoints responses since we don't have a requisite to follow

Closes #624

## Summary of changes

Added endpoint to reroll quests

### Create user
`{:ok, google_user} = GameBackend.Users.find_or_create_google_user_by_email("a")`
### Import quests configs
`GameBackend.CurseOfMirra.Config.import_quest_descriptions_config`

### Add quest to your user
`quests = GameBackend.CurseOfMirra.Quests.add_quest_to_user_id(google_user.user.id, 2, "daily")`

from the quests list grab a daily quest id and run the following command

`curl -X GET localhost:4001/curse/users/YOUR_USER_ID/quest/YOUR_QUEST_ID/reroll_quest`

you should receive a non enough currency error

### Add currency to user
`GameBackend.Users.Currencies.add_currency_by_name_and_game(google_user.user.id, "Gold", 1, 444)`

Run the curl method again and you should receive an Added! message 
If you try to use the curl method again you should receive a "Quest already rerolled" error message

## Checklist


apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -65,6 +70,43 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+    q = from(dq in DailyQuest, preload: [:quest], where: dq.id == ^daily_quest_id)
+
+    Repo.one(q)
+  end
+
+  def get_users_daily_quests(user_id) do
+    q = from(dq in DailyQuest, preload: [:quest], where: dq.user_id == ^user_id)
+
+    Repo.all(q)
+  end
```
Nico-Sanchez, line 83: Add @doc tags for these functions please.

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -65,6 +70,43 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+
+  @doc """
+  Get all %DailyQuest{} for the user_id that were inserted today and were rerolled
```
Nico-Sanchez, line 86: I think this is a better doc.
```suggestion
  Receives a user id and a daily quest type.
  Returns a list of DailyQuest for the given user and type that were inserted today an their status are :rerolled.
```

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -65,6 +70,43 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+      iex>get_user_today_rerolled_daily_quests_by_type(user_id, "daily")
+      [%Quest{type: "daily"}]
+  """
+  def get_user_today_rerolled_daily_quests_by_type(user_id, type) do
+    naive_today = NaiveDateTime.utc_now()
+    start_of_date = NaiveDateTime.beginning_of_day(naive_today)
+    end_of_date = NaiveDateTime.end_of_day(naive_today)
+
+    q =
+      from(dq in DailyQuest,
```
Nico-Sanchez, line 99: If you're taking only DailyQuest into account, why do you have to explicit the type in the function? I think we should either remove the type argument or make the pattern-matching functions for each type.
Also remember to fix the @doc and function name if you change this.

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -109,4 +152,77 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+            Multi.new()
+            # Deduct currency
+            |> Multi.run(:deduct_currencies, fn _, _ ->
+              Currencies.substract_currencies(daily_quest.user_id, reroll_costs)
+            end)
+            # Update old daily quest
+            |> Multi.update(:change_previous_quest, finish_previous_quest_changeset)
+            # Add new daily quest
+            |> Multi.insert(:insert_quest, new_quest_changeset)
+            |> Repo.transaction()
```
Nico-Sanchez, line 224: The comments are redundant to the multi operation description, maybe you could replace the description with the comment (then delete the comment) to be more explicit and accurate.

Nico-Sanchez (changes requested):

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: Done!

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: yeah I think the best approach is to pattern match the quest  type and when we have more quest types like "weekly" or "match" create a proper match to retrieve them

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: I just deleted them, the multi key should be enough to know what's happening imo

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -101,7 +121,7 @@ defmodule GameBackend.CurseOfMirra.Quests do
         preload: [:quest],
         where:
           dq.user_id == ^user_id and dq.inserted_at > ^start_of_date and dq.inserted_at < ^end_of_date and
-            dq.status == ^"rerolled" and q.type == ^type
+            dq.status == ^"rerolled" and q.type == ^"daily"
```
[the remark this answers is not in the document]
Nico-Sanchez: As discussed by huddle, we'll just talk about DailyQuests in this PR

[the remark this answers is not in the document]
agustinesco: Done!

Nico-Sanchez (dismissed):

tvillegas98 (commented):
As we talked, lets put the reroll_quest endpoint under the users scope

tvillegas98 (approved):
Nice work @agustinesco 🚀

Nico-Sanchez (approved):

PR #627, opened 2024-05-16
[GH-618] Triangulate concave polygons

### LINKED ISSUE #618: [FEATURE] Triangulate concave polygons from json config
When we're parsing obstacles from the config file we should admit concave ones and triangulate them

We can use this algorithm https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf

agustinesco:
## Motivation

We had a recurrent problem where every time we wanted to add new obstacles to the game we had to manually split concave shapes into various concave polygons, this was tedious and time-consuming. The idea is to split the convex polygon into a series of triangles using an algorithm and use that to build the game obstacles.

Closes #618

## Summary of changes

The triangulation algorithm used is called *Ear Clipping* you can find more formal information about this in the following sources:
- [Video by two bit coding](https://www.youtube.com/watch?v=QAdfkylpYwc)
- [Geomtric tools article](https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf)

- Added polygon triangulation, this pr will not introduce the algorithm to the match creation process, it will only include the algorithm into the backend

### Cross product usage
On several checks, we're using the cross product between two bi-dimensional vectors to get the angle between them. You can get a visual representation of this operation here: https://www.geogebra.org/m/psMTGDgc
The value you need to see is this one:

Screenshot 2024-05-29 at 10 55 06 AM

# DISCLAIMER: The collisions with the triangles will be weird and will be addressed in future prs

the best way of testing this is using the game client since you can see the triangles.

You can test this on the current config  but we don't have any concave polygons

To run the triangulation you need to change this:
On the `game_updater.ex` file replace the `intialize_obstacles/2` with this:

```Elixir
defp initialize_obstacles(obstacles, last_id) do
    obstacles =
      obstacles
      |> Enum.map(fn obstacle ->
        Entities.new_obstacle(
          1,
          obstacle
        )
      end)
      |> Physics.maybe_triangulate_concave_entities()

    Enum.reduce(obstacles, {Map.new(), last_id}, fn obstacle, {obstacles_acc, last_id} ->
      last_id = last_id + 1

      obstacles_acc =
        Map.put(
          obstacles_acc,
          last_id,
          Map.put(obstacle, :id, last_id)
        )

      {obstacles_acc, last_id}
    end)
  end
```

here are some examples that I tested, to use them replace the "obstacles" list with some of them:

<details>
<summary>shuriken</summary>

``` json 
 {
        "position": {
          "x": 0.0,
          "y": 0.0
        },
        "vertices": [
          {
          "x": 0.0,
          "y": 500.0
          },
          {
          "x": 1000.0,
          "y": 1000.0
          },
          {
          "x": 500.0,
          "y": 0.0
          },
          {
          "x": 1000.0,
          "y": -1000.0
          },
          {
          "x": 0.0,
          "y": -500.0
          },
          {
          "x": -1000.0,
          "y": -1000.0
          },
          {
          "x": -500.0,
          "y": 0.0
          },
          {
          "x": -1000.0,
          "y": 1000.0
          }
        ],
        "radius": 0.0,
        "shape": "polygon"
      }
```
</details>

<details>
<summary>Weird thing</summary>

``` json 
 {
        "position": {
          "x": 0.0,
          "y": 0.0
        },
        "vertices": [
          {
        "x": 0.0,
        "y": 500.0
          },
          {
            "x": 250.0,
            "y": 750.0
          },
          {
            "x": 2000.0,
            "y": 0.0
          },
          {
            "x": 1750.0,
            "y": -1500.0
          },
          {
            "x": 750.0,
            "y": 250.0
          },
          {
            "x": 0.0,
            "y": -1000.0
          },
          {
            "x": -2500.0,
            "y": 0.0
          },
          {
            "x": -500.0,
            "y": 250.0
          },
          {
            "x": -1000.0,
            "y": 750.0
          }
        ],
        "radius": 0.0,
        "shape": "polygon"
      }
```
</details>

<details>
<summary>Weird thing 2</summary>

``` json 
 {
        "position": {
          "x": 0.0,
          "y": 0.0
        },
        "vertices": [
        {
          "x": -1000.0,
          "y": 1000.0
        },
        {
          "x": 2200.0,
          "y": 1000.0
        },
        {
          "x": 2000.0,
          "y": 3500.0
        },
        {
          "x": 2500.0,
          "y": 4500.0
        },
        {
          "x": 3200.0,
          "y": 3500.0
        },
        {
          "x": 3600.0,
          "y": 4000.0
        },
        {
          "x": 3750.0,
          "y": 2000.0
        },    
        {
          "x": 3500.0,
          "y": 500.0
        },
        {
          "x": 3000.0,
          "y": 750.0
        },
        {
          "x": 2200.0,
          "y": 1000.0
        },
        {
          "x": 4000.0,
          "y": -1500.0
        },
        {
          "x": 2500.0,
          "y": -1000.0
        },
        {
          "x": -300.0,
          "y": -3000.0
        },
        {
          "x": -2250.0,
          "y": -1000.0
        },
        {
          "x": -2500.0,
          "y": 2500.0
        },
        {
          "x": -2000.0,
          "y": 3000.0
        }
        ],
        "radius": 0.0,
        "shape": "polygon"
      }
```
</details>

## Checklist
- [x] This change requires new documentation.
  - [x] Documentation has been added/updated.


apps/arena/lib/physics.ex
```diff
@@ -39,4 +39,6 @@ defmodule Physics do
   def distance_between_entities(_entity, _entities), do: :erlang.nif_error(:nif_not_loaded)
 
   def nearest_entity_position_in_range(_entity, _entities, _range), do: :erlang.nif_error(:nif_not_loaded)
+
+  def process_polygon_triangulation(_obstacles), do: :erlang.nif_error(:nif_not_loaded)
```
Nico-Sanchez, line 43: I think `triangulate_concave_entities` is a more descriptive name.

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+
+The logic of the algorithm goes like this:
+1. Traverse the polygon looking for ears
+2. Once we find an ear do the following:
+   - Save the triangle formed by the current previous and next vertex
+   - Delete the current vertex from the available vertex list
+   - Start again with the cleaned up vertex list until the list length is 3
+3. Make a final triangle with the 3 remaining vertices
+
+Even tho the resolution is different the logic is the same as the following article:
```
Nico-Sanchez, line 14:
```suggestion
Even though the resolution is different, the logic is the same as the following article:
```

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+- [Geometric tools article](https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf)
+
+*/
+
+pub(crate) fn maybe_triangulate_polygon(mut polygon: Entity) -> Vec<Entity> {
+    clean_extra_vertices(&mut polygon);
+    if is_polygon_concave(&polygon) {
+        return vec![polygon.clone()];
+    }
```
Nico-Sanchez, line 23: Shouldn't this be
```suggestion
    if is_polygon_convex(&polygon) {
        return vec![polygon.clone()];
    }
```

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+    triangulate_polygon(&polygon)
+}
+
+// Here we'll remove any vertices are shouldn't be part of the polygon or are useless
+// - If three vertices are in a line that's a useless connections that we should remove
+fn clean_extra_vertices(polygon: &mut Entity) {
```
Nico-Sanchez, line 29: I think this comment is more descriptive and precise.
If you change it, remember to change the variables inside the function.
```suggestion
// Remove redundant vertices.
// A redundant vertex is one that is in the middle of a straight line drawn by two other vertices.
fn remove_redundant_vertices(polygon: &mut Entity) {
```

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+
+fn is_vertex_consecutive(
+    previous_vertex: &Position,
+    current_vertex: &Position,
+    next_vertex: &Position,
+) -> bool {
+    let first_vector = Position::sub(previous_vertex, current_vertex);
+    let second_vector = Position::sub(next_vertex, current_vertex);
+
+    get_cross_product_scalar(&first_vector, &second_vector) == 0.0
```
Nico-Sanchez, line 56: What's a `cross product scalar`?
I know there's a dot (scalar) product and a cross product. But why are you mixing these 2 concepts?

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+    polygon.vertices = cleaned_vertices;
+}
+
+fn is_vertex_consecutive(
+    previous_vertex: &Position,
+    current_vertex: &Position,
+    next_vertex: &Position,
+) -> bool {
+    let first_vector = Position::sub(previous_vertex, current_vertex);
+    let second_vector = Position::sub(next_vertex, current_vertex);
```
Nico-Sanchez, line 54: I think we should avoid using "first", "second". Maybe "previous_distance_between_vertices" and "current_distance_between_vertices" is a more descriptive option.
Also, that option seems long. Maybe we should give a name to the straight line linking two vertices and write it down somewhere. You're using "vector" for that but it's too generic.

Please give it a minute or two to come up with better variable naming, taking into account the things I mentioned.

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    get_cross_product_scalar(&first_vector, &second_vector) == 0.0
+}
+
+fn triangulate_polygon(polygon: &Entity) -> Vec<Entity> {
+    let mut result: Vec<Entity> = vec![];
+    let mut positions = polygon.vertices.clone();
```
Nico-Sanchez, line 61:
```suggestion
    let mut vertices_positions = polygon.vertices.clone();
```

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+            let previous_vertex_index = get_previous_cyclic_index(current_vertex_index, &positions);
+            let next_vertex_index = get_next_cyclic_index(current_vertex_index, &positions);
+
+            let previous_vertex = positions[previous_vertex_index];
+            let current_vertex = positions[current_vertex_index];
+            let next_vertex = positions[next_vertex_index];
+
+            if is_position_ear(current_vertex, previous_vertex, next_vertex, &positions) {
+                ear_found = true;
+                let triangle_vertex = vec![current_vertex, next_vertex, previous_vertex];
```
Nico-Sanchez, line 76: This a triangle, not a vertex.
```suggestion
                let ear_triangle = vec![current_vertex, next_vertex, previous_vertex];
```

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+    // This algorithm should always result n - 2 triangles where n is the amount of vertex
+    // in the polygon
+    if result.len() > polygon.vertices.len() - 2 {
+        panic!("Wrong triangulation result")
+    }
+
+    result
+}
+
+// Check if the three vertex are a valid ear, this means
```
Nico-Sanchez, line 105:
```suggestion
// Check if the three vertices are a valid ear, this means
```

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+// 1. The inner angle formed by the three vertex isn't greater that 180 degrees
```
Nico-Sanchez, line 106:
```suggestion
// 1. The inner angle formed by the three vertices aren't greater than 180 degrees
```

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
...
+// 2. There isn't any vertex inside the triangle area
+fn is_position_ear(
```
Nico-Sanchez, line 108: is position ear? I think an ear is a triangle (3 vertices). Please take a look at the names because they're not following the Math concepts.

Nico-Sanchez (changes requested):

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,197 @@
+use crate::map::{Entity, Position};
+
+/*
+Ear clipping triangulation algorithm
+
+The logic of the algorithm goes like this:
+1. Traverse the polygon looking for ears
```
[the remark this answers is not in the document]
agustinesco: The ear definition is already included in the documentation and the `is_position_ear` comments, should we include it here too? I think it's a lot of repetition

apps/arena/lib/physics.ex
agustinesco: yeah i think it's a better name, i'll add a "maybe_" since it doesn't always triangulate

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: You're right! we're checking convexity here

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: Yeah since we're using positions here, i got confused. I'll build the triangle before doing thing and use that instead of a position, so the name `is_triangle_ear` has sense

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: changed to "vertices" i think it has more sense in this context and it's shorter

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: done

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: done

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: done

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: i'll use the same naming convention as `triangle_has_point_inside` to name this variable where we describe the positions that form the vector and the direction "current_to_previous_vector" for example

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
agustinesco: I thought scalar was a good name but i looks like not, here we are just the only non-zero value resulting from doing a cross product between two-dimensional vectors, i'll just change it to cross_product_value to avoid missunderstandings

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,195 @@
...
+        let first_vector = Position::sub(&previous_vertex, &current_vertex);
+        let second_vector = Position::sub(&next_vertex, &current_vertex);
+
+        result = result && get_cross_product_value(&first_vector, &second_vector) >= 0.0;
+    }
+
+    result
+}
+
+fn get_previous_cyclic_index(index: usize, vector: &[Position]) -> usize {
```
Nico-Sanchez, line 181: What's this? What's a cyclic index?

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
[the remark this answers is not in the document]
Nico-Sanchez: I think it should be here.
Documentation is in other file.
is_position_ear is called from this function.
For the reasons I mentioned, I think it should be here. Either the ear definition or just a link to its documentation.

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
Nico-Sanchez: Answered via Slack. It's just we're getting previous vertex_index from a vector and current vertex index. In case the current is the first one, we get the last (works like a cyclic list).

Since the code doesn't ever talk about cyclic, and it's not explained, I think we should change the name of the function to something like "get_previous_vertex_index" or "get_previous_vertex".
I would use the 2nd with a comment above that we return the previous vertex by given vertex and vector.
The "cyclic" meaning just confuses, and the code is self-explanatory: there's an if that does that, it's easy to notice.

agustinesco: I think i prefer the `get_previous_vertex_index` naming, since `get_previous_vertex` leads me to think that we return a vertex, nice suggestion! will make the change

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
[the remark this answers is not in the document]
agustinesco: Understood, I expanded this section of documentation again

Nico-Sanchez (approved):

apps/arena/native/physics/src/collision_detection.rs
tvillegas98, line 1: Why don't we move this file to the folder collision_detection?

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    get_cross_product_value(&current_to_previous_vector, &current_to_next_vector) == 0.0
+}
+
+fn triangulate_polygon(polygon: &Entity) -> Vec<Entity> {
+    let mut result: Vec<Entity> = vec![];
+    let mut vertices = polygon.vertices.clone();
+    let mut ear_found: bool;
+
+    while vertices.len() > 3 {
```
tvillegas98, line 72: This while loop determines whether we can find a ear triangle or not by given some vertices right? Can we create an auxiliary function for this?

tvillegas98 (commented):
Nice work @agustinesco ! I left some comments that could be helpful for us

apps/arena/native/physics/src/collision_detection/ear_clipping.rs
tvillegas98: Thanks for the explanation!

apps/arena/native/physics/src/collision_detection.rs
agustinesco: I think it's due to how modules works in rust, we need a "collision_detection.rs" file to define modules under the "collision_detection" folder

https://doc.rust-lang.org/rust-by-example/mod/split.html

tvillegas98: Ohh I see! thanks for the link

tvillegas98 (approved):
Great work with this @agustinesco !

Nico-Sanchez (approved):

PR #631, opened 2024-05-17
[GH-595] Prestige ranking

### LINKED ISSUE #595: [FEATURE] Ranking system
https://www.notion.so/lambdaclass/Ranking-System-c04164a466b64e27ba9f9c8e81d755c6

When finished, check the corresponding bullet here -> https://github.com/lambdaclass/curse_of_mirra/issues/1102

AminArria:
## Motivation

Closes #595 

See [Ranking System](https://www.notion.so/lambdaclass/Ranking-System-c04164a466b64e27ba9f9c8e81d755c6) for an explanation on the values and how it should work

## Summary of changes

- [x] Add a table to track prestige for each character of a user
- [x] Created config for prestige
  - Contains the prestige rewards based on current prestige and match position
  - Contains the different prestige ranks
- [x] Remove thropies, users overall ranking will be tracked by adding up prestige. No need (at least right now) to have a separate currency/thing for the same purpose
- [x] When reporting match result add/subtract prestige to each user. When subtracting prestige it can't go lower than the minimum for that sub_rank

## How to test it?

To test you can run a game through the browser by going to [localhost:3000](http://localhost:3000/) and login or if you don't have the Google credentials you can create a user `GameBackend.Users.find_or_create_google_user_by_email("tester")` and then go to `http://localhost:3000/board/<YOUR_USER_ID>/muflus/player_name/quick_game`

### Create a google user

```elixir
GameBackend.Users.find_or_create_google_user_by_email("tester")
```

### Then after the match ends you can go to the console and see the results by doing

```elixir
google_user = GameBackend.Repo.get(GameBackend.Users.GoogleUser, google_user.id) |> GameBackend.Repo.preload([user: :units])
google_user.user.units
```

### If you don't feel like doing a match every single time, then you can run this in the console to directly insert a match result and run the functions

```elixir
results = [
  %{
    "user_id" => google_user.id,
    "character" => "muflus",
    "damage_done" => 0,
    "damage_taken" => 716,
    "deaths" => 1,
    "duration_ms" => 41051,
    "health_healed" => 0,
    "killed_by" => "h4ck",
    "killed_by_bot" => true,
    "kills" => 0,
    "position" => 5, ## Modify this to see different prestige rewards
    "result" => "loss"
  }
]
 
GameBackend.CurseOfMirra.Matches.create_arena_match_results(Ecto.UUID.generate(), results)
```

### And this can be automated a bit more in case you want to reach a specific prestige amount

```elixir
id = google_user.id
amount = <AMOUNT_YOU_WANT>
results = [
  %{
    ...
    "position" => 1, ## Modify to 1st position
    ...
  }
]

Enum.reduce_while(1..100000, %{}, fn _i, acc ->
  {:ok, %{{:update_prestige, ^id} => result}} = GameBackend.CurseOfMirra.Matches.create_arena_match_results(Ecto.UUID.generate(), results)
  case result.level >= amount do
    false -> {:cont, result}
    true -> {:halt, result}
  end
end)
```
## Checklist


apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -117,23 +104,42 @@ defmodule GameBackend.Matches do
...
+    match_prestige_reward(%{amount: 0}, position, rewards)
+  end
 
-  ## TODO: Properly pre-process `currencies_config` so the keys are integers and we don't need conversion
-  ##    https://github.com/lambdaclass/mirra_backend/issues/601
-  def get_amount_of_trophies_to_modify(current_trophies, position, currencies_config) when is_integer(position) do
-    get_amount_of_trophies_to_modify(current_trophies, to_string(position), currencies_config)
+  defp match_prestige_reward(prestige, position, rewards) do
+    Map.get(rewards, position)
+    |> Enum.find(fn %{min: minp, max: maxp} -> prestige.amount in minp..maxp end)
```
agustinesco, line 113: I'm a bit concerned about the complexity here, aren't we doing a lot of extra comparisons here? for example, if we're level 999 we'll do 999 comparisons in the worst-case scenario? if we do `prestige.amount > minp and prestige.amount <= maxp` we'll only do 4 comparisons in the worst-case scenario. Correct me if I'm wrong

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -117,23 +104,42 @@ defmodule GameBackend.Matches do
...
-      maximum > current_trophies
-    end)
-    |> Map.get(position)
+  defp calculate_rank_and_amount_changes(nil, reward, ranks) do
+    calculate_rank_and_amount_changes(%{amount: 0}, reward, ranks)
+  end
+
+  defp calculate_rank_and_amount_changes(prestige, reward, ranks) when reward >= 0 do
+    amount = prestige.amount + reward
+    new_rank = Enum.find(ranks, fn rank -> amount in rank.min..rank.max end)
```
agustinesco, line 123: same as above

config/runtime.exs
```diff
@@ -139,17 +139,52 @@ end
...
+  arena_prestige_ranks =
+    Enum.reduce(arena_prestige_ranks_json["ranks"], [], fn rank, acc ->
+      Enum.reduce(rank["sub_ranks"], acc, fn sub_rank, acc ->
+        entry = %{
+          rank: rank["rank"],
+          sub_rank: sub_rank["sub_rank"],
+          min: sub_rank["min"],
+          max: sub_rank["max"]
+        }
+
```
agustinesco, line 162: Just to be sure, all what this is doing is converting the map keys from string to atoms right?

agustinesco (changes requested):
The ranking is being correctly added when i finish a match! I'll approve after answering the questions about complexity

apps/game_backend/lib/game_backend/matches.ex
AminArria: TL;DR `in` for a range gets converted to that same expression you wrote

Ranges are not actually a list of numbers, they are a struct that represent a start, end, and step ([Representation](https://hexdocs.pm/elixir/Range.html#module-representation)), this means that is can have its own implementations of protocols, in particular `Enumerable`

`in` is a macro and gets translated to `Enum.member?/2` ([docs](https://hexdocs.pm/elixir/Kernel.html#in/2))

So, if you look at the code for [Range.member?/2](https://github.com/elixir-lang/elixir/blob/v1.16.3/lib/elixir/lib/range.ex#L545) you will see it is essentially the same check that you mention (plus something to account for the step)

config/runtime.exs
AminArria: Yes, but also a bit more. Its also making the entries more similar to what we would have if this was stored in the DB (every entry with `rank`) without actually duplicating things in the config

apps/game_backend/lib/game_backend/matches.ex
agustinesco: Ooohhh that's good to know! thanks for taking the time to write this down :)

agustinesco (dismissed):
LGTM!

apps/game_backend/priv/repo/migrations/20240517093111_add_prestige_system_tables.exs
```diff
@@ -0,0 +1,16 @@
...
+
+  def change do
+    create table(:character_prestiges) do
+      add :user_id, references(:users, on_delete: :delete_all), null: false
+      add :character, :string, null: false
+      add :rank, :string, null: false
+      add :sub_rank, :integer, null: false
+      add :amount, :integer, null: false
+      timestamps()
+    end
```
Nico-Sanchez, line 12: Let's be careful adding this. I think we should use the tables we already have.
In #637 I'm creating Units for Users.
Units have a 1-1 association with Character, and have attributes like rank.
I think we should use that if we'll share tables across games, but sharing some and creating our own seems like a mess.

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -45,34 +46,20 @@ defmodule GameBackend.Matches do
...
-            google_user.user.id,
-            "Trophies",
-            Utils.get_game_id(:curse_of_mirra),
-            amount
-          )
-        end
-      )
+    Enum.reduce(results, multi, fn result, transaction_acc ->
+      Multi.run(transaction_acc, {:update_prestige, result["user_id"]}, fn repo, %{get_google_users: google_users} ->
+        google_user = Enum.find(google_users, fn google_user -> google_user.id == result["user_id"] end)
```
Nico-Sanchez, line 54: What if the user is a guest (didn't link google account yet)?
I think we should start working on User and not on GoogleUser so it works fine for both.

config/runtime.exs
```diff
@@ -139,18 +139,53 @@ end
...
-
-  config :game_backend, :currencies_config, Jason.decode!(currency_config_json)
-
   {:ok, daily_rewards_json} =
     Application.app_dir(:game_backend, "priv/daily_rewards_rules.json")
     |> File.read()
 
   config :game_backend, :daily_rewards_config, Jason.decode!(daily_rewards_json)
 
+  arena_prestige_ranks_json =
```
Nico-Sanchez, line 148: Arena? I think it should be curse of mirra (we'll translate to champions some day 😂 )

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -117,23 +104,42 @@ defmodule GameBackend.Matches do
...
+  defp calculate_rank_and_amount_changes(prestige, reward, ranks) when reward >= 0 do
+    amount = prestige.amount + reward
+    new_rank = Enum.find(ranks, fn rank -> amount in rank.min..rank.max end)
+    %{rank: new_rank.rank, sub_rank: new_rank.sub_rank, amount: amount}
+  end
+
+  defp calculate_rank_and_amount_changes(prestige, reward, ranks) when reward < 0 do
+    current_rank = Enum.find(ranks, fn rank -> prestige.amount in rank.min..rank.max end)
+    amount = max(prestige.amount + reward, current_rank.min)
+    %{amount: amount}
```
Nico-Sanchez, line 130: I think all this logic should be under CurseOfMirra's path. It's business for CurseOfMirra only.

Nico-Sanchez (changes requested):
I left a few comments regarding our backend abstraction.

Also, is the following quote checked with @marco-paulucci or any PM?
> Remove thropies, users overall ranking will be tracked by adding up prestige. No need (at least right now) to have a separate currency/thing for the same purpose

config/runtime.exs
AminArria: And that's why we should use `arena` in the codebase, it will always be an arena. Regardless of what name it used outside

apps/game_backend/lib/game_backend/matches.ex
AminArria: > What if the user is a guest (didn't link google account yet)?

Is that an actual thing that will happen? Also, we need to fix the match results processing to handle this "bad"/missing users. But I will make a follow up ticket for it, that will touch to many things here and don't want to overcomplicate the PR

> I think we should start working on User and not on GoogleUser so it works fine for both.

Yep, this is also a mistake from the match result inception. I'll include it in the ticket to fix all this

Nico-Sanchez (changes requested):
Please update this PR with main branch and update this feature to match the codebase or viceversa accordingly.

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -104,44 +105,47 @@ defmodule GameBackend.Matches do
...
-  defp insert_or_update_prestige(repo, user_id, character, changes, nil) do
-    attrs = Map.merge(changes, %{user_id: user_id, character: character})
+  defp insert_or_update_prestige(repo, user_id, character, changes, _unit) do
+    ## Although this looks bad, since we are making a query per player per game, it is not that bad
+    ## because this only run on the first match of a character for a player, so it will be rare
+    %{id: character_id} = repo.get_by!(Character, name: character)
+    attrs = Map.merge(changes, %{user_id: user_id, character_id: character_id})
 
-    CharacterPrestige.insert_changeset(attrs)
+    Unit.moba_insert_changeset(attrs)
```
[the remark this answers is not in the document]
agustinesco: Yeah I think we should keep the curse_of_mirra convention for the time being (even tho we are champions now) and refactor the code in another issue so we don't go too outside of this prs scope

apps/game_backend/lib/game_backend/matches.ex
```diff
@@ -52,13 +53,13 @@ defmodule GameBackend.Matches do
...
-        reward = match_prestige_reward(prestige, result["position"], prestige_config[:rewards])
-        changes = calculate_rank_and_amount_changes(prestige, reward, prestige_config[:ranks])
-        insert_or_update_prestige(repo, google_user.user.id, result["character"], changes, prestige)
+        ## This is because currently characters name are in this format
+        ## later on if match results report character_id instead we can skip it
+        result_character = String.capitalize(result["character"])
+        unit = Enum.find(google_user.user.units, %{level: 0}, fn unit -> unit.character.name == result_character end)
+        reward = match_prestige_reward(unit, result["position"], prestige_config[:rewards])
+        changes = calculate_rank_and_amount_changes(unit, reward, prestige_config[:ranks])
+        insert_or_update_prestige(repo, google_user.user.id, result_character, changes, unit)
```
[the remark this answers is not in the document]
agustinesco: yeah we should match the character from the result with the character that some unit from the user has, if we don't find any that means the result is weird

Nico-Sanchez (changes requested):
Nice changes. Can you update the "how to test it" part so I can test it properly?

agustinesco:
done!

Nico-Sanchez (changes requested):
Really nice work @agustinesco! 🪨 

Let's create a new field "prestige" and use that new field instead of using "level".
Remember the prestige starts from zero, not from 1.
That way it's more accurate to our notion and business naming.

agustinesco:
> Really nice work @agustinesco! 🪨
> 
> Let's create a new field "prestige" and use that new field instead of using "level". Remember the prestige starts from zero, not from 1. That way it's more accurate to our notion and business naming.

Sounds good! done

Nico-Sanchez (approved):

Nico-Sanchez (approved):

tvillegas98 (approved):
Nice work both! 😄

PR #632, opened 2024-05-17
[GH-428]  Implement Buffs for Dungeon

lotuuu:
> [!NOTE]
> Closes https://github.com/lambdaclass/afk_gacha_game/issues/428

## Motivation
Introduces upgrades/buffs for users. The way this works is that we add:
- Unlocks for a user, which associate them to the upgrades they own.
- Upgrades, which contain the relevant data like upgrade cost, other upgrades requirements and unlocked buffs.
References:
[DungeonSpring (upgrades)](https://www.notion.so/lambdaclass/DungeonSpring-8c5a077e3c0449fd80737f56a0a471b2?pvs=4)

We want to handle these buffs *before* entering a simulation. We do NOT want to treat them as modifiers, since they are not removable by any means; instead we want them to affect the initial stats of units.

## Summary of changes

- Create new schemas for Unlocks (for user), Upgrades and UpgradeUnlocks (for upgrades that lock other upgrades)
- Add functions for purchasing and querying these upgrades/unlocks
- Implement buff application before battle start, both for regular stats and for the specific max_level debuff
- Create mock initial debuff & purchasable upgrades in seeds

## How to test it?

Only testable through console.

#1: You will need to create a user with `Champions.Users.register/1` and fight a regular level (from the main campaign).
In the response, take a look at the initial_state value, more specifically at the health of your units.

#2: Then play a level of the Dungeon campaign. You should see that the health they start with now is lower, as we're now applying the debuff for dungeon mode.

Now purchase the  `"HP Upgrade 1"` with `Champions.Users.purchase_dungeon_upgrade/2` (you'll need to give your user some Pearls with `GameBackend.Users.Currencies.add_currency_by_name_and_game/4`). Do #2 again and you should see that the initial health of your units grew a little bit.

Level up a unit a ridiculous amount, like get it to level 90 (easier by using a db management program like pgadmin or dbeaver) and do steps #1 and #2 again. You should see that for the main campaign level their initial health is much higher, while for the dungeon mode it's not too high, since aside from multiplying it by `0.1`, we're also limiting the level in the calculation to 10.

Useful commands
```
# For getting the next level in a specific super campaign (replace the id with the super campaign)
next_level = GameBackend.Campaigns.get_super_campaign_progress(user.id, "bc1944f9-64da-420a-be3e-b757f4b2dc26") |> elem(1) |> Map.get(:level_id)

# For purchasing an upgrade given its name
Champions.Users.purchase_dungeon_upgrade(user.id, GameBackend.Users.get_upgrade_by_name("HP Upgrade 1") |> elem(1) |> Map.get(:id))
```


ncontinanza (changes requested):

priv/repo/seeds.exs
```diff
@@ -298,6 +294,68 @@ _dungeon_settlement_levels =
...
+      name: "HP Upgrade 1",
+      description: "This upgrade increases the health of all units by 5%.",
+      group: 1,
+      cost: [
+        %{currency_id: pearls_currency.id, amount: 5}
+      ],
+      buffs: [
+        %{
+          modifiers: [
+            %{attribute: "health", magnitude: 1.05, operation: "Multiply"}
```
ncontinanza, line 329:
```suggestion
            %{attribute: "health", magnitude: 0.05, operation: "Multiply"}
```

ncontinanza (changes requested):

ncontinanza (dismissed):

apps/game_backend/lib/game_backend/users/upgrades/unlock_requirement.ex
agustinesco, line 1: We talked about this table and it represents a chain of upgrade dependencies, e.g. 
upgrade 2 --is locked behind --> upgrade 1 
So first you need to buy upgrade 1 to buy upgrade 2 

The thing is that it shared names with the table "unlocks" and it's a little bit misleading, maybe we could rename this table or the unlock table to make this clearer? also adding a comment into the "upgrade" table would help a lot

lotuuu: Will do!

apps/game_backend/lib/game_backend/units/skills/skill.ex
```diff
@@ -4,6 +4,7 @@ defmodule GameBackend.Units.Skills.Skill do
   use GameBackend.Schema
   import Ecto.Changeset
 
+  alias GameBackend.Units.Buffs.Buff
```
agustinesco, line 7: This alias is wrong right? this module doesn't exists. The right one would be
`GameBackend.Users.Upgrades.Buff`

apps/game_backend/lib/game_backend/users/upgrades/unlock_requirement.ex
agustinesco, line 1: The file name is missing

agustinesco (approved):
LGTM!

ncontinanza (approved):

PR #647, opened 2024-05-22
[GH-404] lowest and highest stat targeting strategies

### LINKED ISSUE #404: Increase items and power ups pickup area

ncontinanza:
## Motivation

We need to implement two new targeting strategies, that will target the unit with lowest or highest value of a stat, depending the case.

Closes https://github.com/lambdaclass/afk_gacha_game/issues/404

## Summary of changes

- Implement lowest stat strategy
- Implement highest stat strategy
- Add tests

## How to test it?

1. Start the client in the main branch, and the backend with this branch.
2. Pick one of your units and change its basic skill's targeting strategy to lowest. You should end up with something like the following:
```
"targeting_strategy": {
     "type": {"lowest": "health"},
     "target_allies": false
}
```
3. Run `Champions.Config.import_skill_config` in your iex console to apply the changes.
4. Run a battle in the client. To make things easier, you can only leave the unit you modified in your line up.
   4.1. Remember to change all `Logger.info()` to `IO.inspect()`, so that you can see the complete log in your terminal.
5. Check the logs, and verify that the unit you changed is always attacking the opponent's unit with the lowest health.
6. Repeat to test the case for `highest` strategy.

Also, you can check the tests that were added in this PR

## Checklist


apps/champions/test/battle_test.exs
```diff
@@ -1243,6 +1243,206 @@ defmodule Champions.Test.BattleTest do
       assert "team_2" ==
                Champions.Battle.Simulator.run_battle([self_damage_unit], [target_dummy], maximum_steps: maximum_steps).result
     end
+
+    test "Lowest Health" do
+      # This test pairs a unit with a basic skill that deals damage to the enemy with the lowest health, against two target dummies.
+      # The first target dummy has 10 health points and the second one has 9 health points. The dummy with lowest health can kill the opponent in one hit, so if the lowest health targeting strategy fails, the battle will end in a victory for the team_2.
```
lotuuu, line 1249: If the target dummy can obliterate the attacker with a skill then it's not a target dummy! let's rename it to something more descriptive

apps/game_backend/lib/game_backend/units/skills/mechanics/targeting_strategies/target_strategy.ex
```diff
@@ -18,14 +18,14 @@ defmodule GameBackend.Units.Skills.Mechanics.TargetingStrategies.Type do
 
   def cast(%{factions: factions}), do: {:ok, %{factions: factions}}
   def cast(%{classes: classes}), do: {:ok, %{classes: classes}}
-  def cast(%{lowest: attribute}), do: {:ok, %{lowest: attribute}}
-  def cast(%{highest: attribute}), do: {:ok, %{highest: attribute}}
+  def cast(%{lowest: attribute}), do: {:ok, %{"lowest" => attribute}}
```
lotuuu, line 21: Is this necessary?

lotuuu (changes requested):

apps/game_backend/lib/game_backend/units/skills/mechanics/targeting_strategies/target_strategy.ex
ncontinanza: I thought it was, but it isn't

lotuuu:
As we discussed, try looking into how hard it would be to randomize the order of units that are tied in `stat`, because right now there is a set precedence over units that are tied, making some units more vulnerable to attacks for no reason

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -1203,6 +1205,22 @@ defmodule Champions.Battle.Simulator do
     }
   end
 
+  defp sort_units_by_stat(units, stat, desc) do
+    Enum.sort(
+      units,
+      &cond do
+        calculate_unit_stat(&1, String.to_atom(stat)) > calculate_unit_stat(&2, String.to_atom(stat)) ->
```
lotuuu, line 1212: Let's capture calculate_unit_stat before so we run it twice instead of 4 times

lotuuu (changes requested):

lotuuu (approved):

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -621,6 +621,34 @@ defmodule Champions.Battle.Simulator do
     [caster.id]
   end
 
+  defp choose_targets_by_strategy(
+         caster,
+         %{type: %{"lowest" => stat}, target_allies: target_allies, count: count},
+         state
```
Nico-Sanchez, line 627: I have a few suggestions regarding these 2 new function signatures:
1. I wouldn't extract map keys that will only be used in the body and not for pattern-match use. In this case, I would just match the "type" which is the field that differs in these functions. (See https://hexdocs.pm/elixir/main/code-anti-patterns.html#complex-extractions-in-clauses).
2. I don't like some functions extracting "type" field as a value and others as a nested map key. I think every function should do the same, being just the type a value for type key my preferable option. If you need an additional parameter (stat) just add a new key to the map.

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -621,6 +621,34 @@ defmodule Champions.Battle.Simulator do
...
+       ) do
+    target_team =
+      Enum.filter(state.units, fn {_id, unit} -> unit.team == caster.team == target_allies end)
+
+    Enum.map(target_team, fn {_id, unit} -> unit end)
+    |> sort_units_by_stat(stat, true)
```
Nico-Sanchez, line 633: This "true" value seems random. I think we should send an atom (:desc or :asc) and do the 2 separate functions, to sort the units in the desired order.

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -621,6 +621,34 @@ defmodule Champions.Battle.Simulator do
...
+         %{type: %{"highest" => stat}, target_allies: target_allies, count: count},
+         state
+       ) do
+    target_team =
+      Enum.filter(state.units, fn {_id, unit} -> unit.team == caster.team == target_allies end)
+
+    Enum.map(target_team, fn {_id, unit} -> unit end)
+    |> sort_units_by_stat(stat, false)
+    |> Enum.take(count)
+    |> Enum.map(fn unit -> unit.id end)
```
Nico-Sanchez, line 649: This whole behavior is the same as above, it just changes the sort order. Can we wrap this up in a function, so each parent function calls it with the desired order to sort?

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -1069,20 +1097,29 @@ defmodule Champions.Battle.Simulator do
...
-            "random"
+          cond do
+            is_binary(targeting_strategy_type) && targeting_strategy_type in @implemented_targeting_strategies ->
+              targeting_strategy_type
+
+            hd(Map.keys(targeting_strategy_type)) in @implemented_targeting_strategies ->
+              targeting_strategy_type
+
+            true ->
+              "random"
```
Nico-Sanchez, line 1122: This is getting messy. However, I wrote a suggestion in a previous comment that should help to clean this.

Nico-Sanchez (changes requested):

apps/champions/lib/champions/battle/simulator.ex
ncontinanza: As discussed, we'll solve your second comment in a separate issue: https://github.com/lambdaclass/afk_gacha_game/issues/479

apps/champions/lib/champions/battle/simulator.ex
ncontinanza: As discussed, we'll solve your this comment in a separate issue: https://github.com/lambdaclass/afk_gacha_game/issues/479

Nico-Sanchez (approved):

lotuuu:
If we're going to change the design principle we're following, as per @Nico-Sanchez's great suggestion, we should go ahead and follow it in our other `choose_targets_by_strategy` while we're at it! We don't want to introduce heterogeneous code practices in our PRs if we can afford it

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -621,10 +621,40 @@ defmodule Champions.Battle.Simulator do
     [caster.id]
   end
 
-  defp choose_targets_by_strategy(caster, %{type: "all", target_allies: target_allies}, state),
+  defp choose_targets_by_strategy(
+         caster,
+         %{type: %{"lowest" => stat}} = targeting_strategy,
+         state
+       ) do
+    filter_and_choose_units_by_stat(
```
lotuuu, line 629:
```suggestion
  defp choose_targets_by_strategy(caster, %{type: %{"lowest" => stat}} = targeting_strategy, state) do
```

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -621,10 +621,40 @@ defmodule Champions.Battle.Simulator do
...
+      :desc
+    )
+  end
+
+  defp choose_targets_by_strategy(
+         caster,
+         %{type: %{"highest" => stat}} = targeting_strategy,
+         state
+       ) do
+    filter_and_choose_units_by_stat(
```
lotuuu, line 644:
```suggestion
  defp choose_targets_by_strategy(caster, %{type: %{"highest" => stat}} = targeting_strategy, state) do
```

lotuuu (changes requested):
some formatting shenanigans :P

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -636,6 +666,16 @@ defmodule Champions.Battle.Simulator do
     Enum.take(sorted_units, amount)
   end
 
+  defp filter_and_choose_units_by_stat(units, stat, count, caster, target_allies, order) do
```
lotuuu, line 669: I'd rename this to `choose_units_by_stat_and_team`, the "filter" word seems confusing

apps/champions/lib/champions/battle/simulator.ex
lotuuu: suggestion got suggested wrongly :P you will need to make this change manually

apps/champions/lib/champions/battle/simulator.ex
lotuuu: suggestion got suggested wrongly :P you will need to make this change manually

apps/champions/lib/champions/battle/simulator.ex
lotuuu: suggestion got suggested wrongly :P you will need to make this change manually

apps/champions/lib/champions/battle/simulator.ex
ncontinanza: lol I hate when that happens

lotuuu (changes requested):
Last one!

apps/champions/lib/champions/battle/simulator.ex
lotuuu: yeah it's also uneditable! sucks

apps/champions/lib/champions/battle/simulator.ex
ncontinanza: I like it!

lotuuu (approved):

Nico-Sanchez (approved):

PR #658, opened 2024-05-27
[GH-470] [Dungeon] Add tests for dungeon buffs/upgrades

lotuuu:
## Motivation

#632 added buffs but didn't implement tests since the PR was too long. This PR introduces those tests.
Closes https://github.com/lambdaclass/afk_gacha_game/issues/470

## Summary of changes

- Add integration tests for upgrade purchasing and application in battle through SocketTester
- Add Battle test for stat initial modifiers
- Add Battle test for max_level initial modifiers

## How to test it?

run the tests!
`mix test` 

also this messes with some protobuf messages so check in the client that nothing breaks :P (spoiler alert: nothing breaks)


apps/serialization/gateway.proto
```diff
@@ -347,6 +354,25 @@ syntax = "proto3";
...
+    string name = 1;
+    Upgrade upgrade = 2;
+  }
+
+  message Upgrade {
+    string name = 1;
+    string description = 2;
+    int32 group = 3;
+    repeated CurrencyCost cost = 4;
+    repeated Buff buffs = 9;
```
ncontinanza, line 367: Is there a reason for this jump, or could we assign number 5 to `buffs`?

apps/champions/test/battle_test.exs
```diff
@@ -1681,4 +1677,179 @@ defmodule Champions.Test.BattleTest do
...
     end
   end
+
+  describe "Initial Modifiers" do
+    test "Initial Modifiers modify stats correctly" do
+      # We will run a battle with a unit that has some modifications to its stats.
+      # Its initial state for "health" would have been 5 due to its character's stats, but it will be buffed up to 10.
+      # It will fight a unit that will deal 5 damage to it. Without the modifications, it would die.
+      # After surviving an attack, it will deal 100% ATK damage to the enemy, that also has 10 health.
+      # Because of the attack buff, which brings hist total ATK to 10, he will kill the enemy in one hit, and battle won't end on a timeout.
```
ncontinanza, line 1687:
```suggestion
      # Because of the attack buff, which brings his total ATK to 10, he will kill the enemy in one hit, and battle won't end on a timeout.
```

ncontinanza (changes requested):

apps/serialization/gateway.proto
lotuuu: Not really, probably comes from fields I added initially & removed afterwards

ncontinanza (approved):

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -388,6 +389,10 @@ defmodule GameBackend.Users do
         Currencies.substract_currencies(user_id, upgrade.cost)
       end)
       |> Transaction.run()
+      |> case do
+        {:ok, _} -> {:ok, get_user(user_id)}
+        {:error, _} -> {:error, :unknown_error}
```
agustinesco, line 394: I think this case is wrong right? multi error result is a cuadruple tuple {:error, key, value, changes so far}

agustinesco (approved):
LGTM!

ncontinanza (approved):
LGTM!

Issue #399, opened 2024-03-21
[BUG] Muflus ulti skill position calculation


Nico-Sanchez:
Muflus stays in the same place during the leap skill time and changes it to the new one when it ends. It should change its position little by little from origin to its end.

I'll attach some videos of this bugs replicated in the client:

https://github.com/lambdaclass/mirra_backend/assets/65268603/12dbd084-0727-4585-9a6a-2cdb251377bb

https://github.com/lambdaclass/mirra_backend/assets/65268603/ee3764de-81d1-4c18-8a89-6d741af2fbbe

Nico-Sanchez:
It's not a backend issue. It's a bug from the unity client.

Proofs:

https://github.com/lambdaclass/mirra_backend/assets/65268603/7a68c63a-342d-4a73-bbe9-60a36b880b5c


https://github.com/lambdaclass/mirra_backend/assets/65268603/139320de-d80f-4bf4-bf6e-5f4e3515561c

PR #662, opened 2024-05-28
[GH-399] Deal damage over time

ncontinanza:
## Motivation

We need to implement the damage over time executions for effects, following the guidelines found in [the design document](https://www.notion.so/lambdaclass/Executions-dd3daa21250848f99b4e6aa6bd2ec468?pvs=4)

Closes https://github.com/lambdaclass/afk_gacha_game/issues/399

## Summary of changes

- Add `executions_over_time` to effects
- Process the executions over time when a skill that has one of them is casted. 
- Add a test for this new behaviour

## How to test it?

Check the tests and play with the skills cooldowns, DoT duration and period to check that the behaviour is the expected one.

## Checklist


apps/champions/test/battle_test.exs
```diff
@@ -1681,4 +1681,68 @@ defmodule Champions.Test.BattleTest do
                Champions.Battle.Simulator.run_battle([speeder], [damager], maximum_steps: maximum_steps + 1).result
     end
   end
+
+  describe "Executions over time" do
+    test "DealDamage over time", %{target_dummy: target_dummy} do
```
lotuuu, line 1686:
```suggestion
    test "DealDamageOverTime", %{target_dummy: target_dummy} do
```

apps/champions/test/battle_test.exs
```diff
@@ -1681,4 +1681,68 @@ defmodule Champions.Test.BattleTest do
...
+      # We will create a battle between a damaging unit and a target dummy.
+      # The unit's basic skill will deal 4 points of damage to the target dummy over 3 steps, killing it in the third one. The cooldown is such that the skill can be used only once.
```
lotuuu, line 1688: it deals damage over dot_duration * dot_interval = 6 steps

apps/champions/test/battle_test.exs
```diff
@@ -1681,4 +1681,68 @@ defmodule Champions.Test.BattleTest do
...
+          ultimate_skill: TestUtils.build_skill(%{name: "DealDamageOverTime Empty Skill"}),
+          base_attack: 4
+        })
+        |> Characters.insert_character()
+
+      {:ok, unit} = TestUtils.build_unit(%{character_id: character.id}) |> Units.insert_unit()
+
+      {:ok, unit} = Units.get_unit(unit.id)
+
+      assert "team_1" ==
```
lotuuu, line 1744: let's check that if we have 1 less step, the battle ends in timeout

lotuuu (changes requested):
The overall implementation is in line with what I had imagined, love it! Left some small code design comments and general details

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -366,6 +365,12 @@ defmodule Champions.Battle.Simulator do
       |> put_in([:units, unit.id, :modifiers], new_modifiers)
       |> put_in([:units, unit.id, :tags], new_tags)
 
+    {new_unit, new_history} = process_executions_over_time(new_state.units[unit.id], new_history)
+
+    new_state =
+      new_state
+      |> put_in([:units, unit.id], new_unit)
```
lotuuu, line 380:
```suggestion
    new_state = put_in(new_state, [:units, unit.id], new_unit)
```

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -1071,8 +1071,8 @@ defmodule Champions.Battle.Simulator do
     {target, history}
   end
 
-  defp process_executions_over_time(unit, history) do
-    Enum.reduce(unit.executions_over_time, {unit, history}, fn execution_over_time, {unit_acc, history_acc} ->
+  defp process_executions_over_time(unit_initial_state, current_unit, history) do
+    Enum.reduce(unit_initial_state.executions_over_time, {current_unit, history}, fn execution_over_time, {unit_acc, history_acc} ->
```
lotuuu: We're using the initial unit's executions over time, but when we call `process_execution_over_time` we're not passing in the unit's initial state. That means the armor in the calculation gets affected by this steps' modifiers!

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -1130,68 +1131,53 @@ defmodule Champions.Battle.Simulator do
...
          history
        ) do
-    {new_target, new_history} =
-      apply_execution_over_time(
-        execution_over_time,
-        attack_ratio,
-        target,
-        history
-      )
+    case remaining_interval_steps do
```
lotuuu, line 1139: Let's just use an if here, we're not matching anything

apps/champions/test/battle_test.exs
```diff
@@ -2057,8 +2057,8 @@ defmodule Champions.Test.BattleTest do
   describe "Executions over time" do
     test "DealDamageOverTime", %{target_dummy: target_dummy} do
       # We will create a battle between a damaging unit and a target dummy.
-      # The unit's basic skill will deal 4 points of damage to the target dummy over 3 steps, killing it in the third one. The cooldown is such that the skill can be used only once.
-      # The battle should finish with a victory for team_1 after the third step, or in timeout if the steps are less than 3.
+      # The unit's basic skill will deal 4 points of damage to the target dummy over dot_duration * dot_interval = 6 steps, killing it in the third one. The cooldown is such that the skill can be used only once.
```
lotuuu, line 2060:
```
      # The unit's basic skill will deal 4 points of damage to the target dummy over 6 steps, killing it in the sixth one. The cooldown is such that the skill can be used only once.
```

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -1210,9 +1193,21 @@ defmodule Champions.Battle.Simulator do
...
       )
 
+    initial_interval_steps = get_interval_steps(execution_over_time.execution)
+
     new_target =
       target
       |> Map.put(:health, target.health - damage_after_defense)
+      |> update_in([:executions_over_time], fn current_executions ->
+        Enum.map(current_executions, fn exec ->
+          if exec.skill_id == execution_over_time.skill_id do
```
lotuuu, line 1203: Correct me if I'm wrong, but this will fail as the same execution over time could be applied to a unit twice (imagine a very short cooldown, or 2 same heroes in a fight. Also, two different executions that have the same `skill_id`). That would probably break this implementation: `remaining_duration` would decrease by `n` every step instead of 1, `n` being the amount of times `skill_id` is repeated among the target

lotuuu (changes requested):

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -1088,6 +1111,110 @@ defmodule Champions.Battle.Simulator do
...
+      )
+
+    initial_interval_steps = get_interval_steps(execution_over_time.execution)
+
+    new_target =
+      target
+      |> Map.put(:health, target.health - damage_after_defense)
+      |> update_in([:executions_over_time], fn current_executions ->
+        Enum.map(current_executions, fn exec ->
+          if exec == execution_over_time do
```
lotuuu, line 1206: I think there is a really, really edge-y edge case here where this fails still. Do you think we should add an issue to revisit this later?

lotuuu (changes requested):
Code is looking fine to me! However I'm getting an error when trying to play a level:
```
[error] Ranch listener Gateway.Endpoint.HTTP had connection process started with :cowboy_clear:start_link/4 at #PID<0.1514.0> exit with reason: {%Protocol.UndefinedError{protocol: Enumerable, value: nil, description: ""}, [{Enumerable, :impl_for!, 1, [file: ~c"lib/enum.ex", line: 1]}, {Enumerable, :reduce, 3, [file: ~c"lib/enum.ex", line: 166]}, {Enum, :reduce, 3, [file: ~c"lib/enum.ex", line: 4399]}, {Champions.Battle.Simulator, :apply_tags, 4, [file: ~c"lib/champions/battle/simulator.ex", line: 968]}, {Enum, :"-reduce/3-lists^foldl/2-0-", 3, [file: ~c"lib/enum.ex", line: 2528]}, {Champions.Battle.Simulator, :"-process_step_for_effects/1-fun-0-", 4, [file: ~c"lib/champions/battle/simulator.ex", line: 592]}, {Enum, :"-reduce/3-lists^foldl/2-0-", 3, [file: ~c"lib/enum.ex", line: 2528]}, {Champions.Battle.Simulator, :"-process_step_for_effects/1-fun-2-", 2, [file: ~c"lib/champions/battle/simulator.ex", line: 588]}, {Enum, :"-reduce/3-lists^foldl/2-0-", 3, [file: ~c"lib/enum.ex", line: 2528]}, {Champions.Battle.Simulator, :process_step_for_effects, 1, [file: ~c"lib/champions/battle/simulator.ex", line: 579]}, {Champions.Battle.Simulator, :"-simulate_battle/3-fun-0-", 3, [file: ~c"lib/champions/battle/simulator.ex", line: 193]}, {Enumerable.Range, :reduce, 5, [file: ~c"lib/range.ex", line: 538]}, {Enum, :reduce_while, 3, [file: ~c"lib/enum.ex", line: 2582]}, {Champions.Battle.Simulator, :simulate_battle, 3, [file: ~c"lib/champions/battle/simulator.ex", line: 184]}, {Champions.Battle, :run_battle, 3, [file: ~c"lib/champions/battle.ex", line: 61]}, {Ecto.Multi, :apply_operation, 5, [file: ~c"lib/ecto/multi.ex", line: 883]}, {Enum, :"-reduce/3-lists^foldl/2-0-", 3, [file: ~c"lib/enum.ex", line: 2528]}, {Ecto.Multi, :"-apply_operations/5-fun-1-", 5, [file: ~c"lib/ecto/multi.ex", line: 856]}, {Ecto.Adapters.SQL, :"-checkout_or_transaction/4-fun-0-", 3, [file: ~c"lib/ecto/adapters/sql.ex", line: 1358]}, {DBConnection, :run_transaction, 4, [file: ~c"lib/db_connection.ex", line: 1710]}]}
```


This is the battle:
Screenshot 2024-06-10 at 3 44 30 PM

apps/champions/lib/champions/battle/simulator.ex
ncontinanza: As discussed, the case in which two units of the same character apply the same EoT to a target would be handled by this comparison because the execution over time would have different values in the `caster` field.

lotuuu (dismissed):

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -268,6 +268,8 @@ defmodule Champions.Battle.Simulator do
   # Calculate the new state of the battle after a step passes for a unit.
   # Updates cooldowns, casts skills and reduces self-affecting modifier durations.
   defp process_step_for_unit(unit, current_state, initial_step_state, history) do
```
agustinesco, line 270: I think this function is growing a lot and it's starting to be hard to follow it, maybe we can refactor it following the same convention as `simulate_battle`

apps/champions/lib/champions/battle/simulator.ex
```diff
@@ -1088,6 +1111,110 @@ defmodule Champions.Battle.Simulator do
...
+  defp process_execution_over_time(
+         %{remaining_duration: -1} = execution_over_time,
+         target,
+         history,
+         _target_initial_state
+       ) do
+    # If the execution is over, we remove it from the target
+    new_target =
+      update_in(target, [:executions_over_time], fn current_executions ->
+        Enum.filter(current_executions, fn exec -> exec.skill_id != execution_over_time.skill_id end)
```
agustinesco, line 1130: What will happen if two characters with the same skill apply an effect to the same character in different times, isn't this going to cause that effect to be removed when the first one expires?

apps/champions/lib/champions/battle/simulator.ex
ncontinanza: Will do! I've created this issue so we don't lose track of it: https://github.com/lambdaclass/afk_gacha_game/issues/498

apps/champions/lib/champions/battle/simulator.ex
ncontinanza: Yes, that will turn into a problem. I changed it, same as I did in [fdafe26](https://github.com/lambdaclass/mirra_backend/pull/662/commits/fdafe26fc1dca556d7e9502c0a8cfd6e2396a69e)

agustinesco (approved):
LGTM!

lotuuu (approved):

PR #681, opened 2024-06-05
[GH-439] Currency Capping

lotuuu:
## Motivation

Adds caps to user currencies. They have been dynamic to meet with the general requirements of our backend: you can create a UserCurrencyCap for any user-currency combo. These caps can differ between users.

For AFK Autobattler we want to update the limit everytime with each Dungeon Settlement level up.

We could have gone for an implementation more similar to how we handle afk currencies, where the max supply is stored on each DungeonSettlementLevel and it's obtained by calling something like `get_currency_cap(user_id, "Supplies")` and pattern matching it and handling it like this. I chose to go for this approach because I wanted to keep the initial implementation of currency caps more abstract and usable across our games. We can and probably should go for that optimization later, but in order to introduce the concept I think this is the best we can do right now.

Closes https://github.com/lambdaclass/afk_gacha_game/issues/439

## Summary of changes

- Create user_currency_caps table and schema.
- Rename supply_limit to supply_cap.
- Add initial cap to user's Supplies based on initial DungeonSettlementLevel's cap.
- Modify user's Supplies on every DungeonSettlement level up.

## How to test it?

Give some "Supplies" to a user! You'll find that you're able to do so freely up until you reach 10 or more (initial `dungeon_settlement_level` setting). Afterwards, you won't be able to add anymore to them until you spend enough to cross the treshold.

### DIFF
--- a/apps/game_backend/lib/game_backend/users.ex
+++ b/apps/game_backend/lib/game_backend/users.ex
@@ -10,7 +10,6 @@ defmodule GameBackend.Users do
   """
 
   import Ecto.Query, warn: false
-  alias Gateway.Serialization.DungeonSettlementLevel
   alias Ecto.Multi
   alias GameBackend.CurseOfMirra.Users, as: CurseUsers
   alias GameBackend.Matches.ArenaMatchResult
--- a/apps/game_backend/lib/game_backend/users.ex
+++ b/apps/game_backend/lib/game_backend/users.ex
@@ -259,7 +259,7 @@ defmodule GameBackend.Users do
   Inserts a DungeonSettlementLevel into the database. If another one already exists with the same number, it updates it instead.
   """
   def upsert_dungeon_settlement_level(attrs) do
-    case get_dungeon_settlement_level(attrs.level) do
+    case get_dungeon_settlement_level_by_number(attrs.level) do
       nil -> insert_dungeon_settlement_level(attrs)
       dungeon_settlement_level -> update_dungeon_settlement_level(dungeon_settlement_level, attrs)
     end
--- a/apps/game_backend/lib/game_backend/users.ex
+++ b/apps/game_backend/lib/game_backend/users.ex
@@ -10,6 +10,7 @@ defmodule GameBackend.Users do
   """
 
   import Ecto.Query, warn: false
+  alias Gateway.Serialization.DungeonSettlementLevel
   alias Ecto.Multi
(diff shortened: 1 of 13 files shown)


apps/game_backend/lib/game_backend/users/currencies.ex
```diff
@@ -85,16 +83,33 @@ defmodule GameBackend.Users.Currencies do
...
+            |> max(0)
+
+          user_currency
+          |> UserCurrency.update_changeset(%{amount: new_amount})
+          |> Repo.update()
+
         nil ->
           # User has none of this currency, create it with given amount
-          insert_user_currency(%{user_id: user_id, currency_id: currency_id, amount: amount})
+          insert_user_currency(%{user_id: user_id, currency_id: currency_id, amount: max(amount, 0)})
```
agustinesco, line 112: Don't we need to check the cap here?

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -342,6 +349,15 @@ defmodule GameBackend.Users do
       |> Multi.run(:user_currency, fn _, _ ->
         Currencies.substract_currencies(user_id, level_up_costs)
       end)
+      |> Multi.run(:supply_cap, fn _, %{user: user} ->
+        dungeon_settlement_level = Repo.get(DungeonSettlementLevel, user.dungeon_settlement_level_id)
+
+        Currencies.update_user_currency_cap(
+          user.id,
+          {"Supplies", user.game_id},
```
agustinesco, line 357: Don't this need to cap every key of `afk_reward_rates`?

apps/game_backend/lib/game_backend/users/currencies.ex
lotuuu: We don't because we won't have currency caps of zero (at least it doesn't make any sense for now, we might want to revisit it in the future) so this will in the worst case be the first overflow of the cap, which we want to allow!

apps/game_backend/lib/game_backend/users.ex
agustinesco: nice!

apps/game_backend/lib/game_backend/users/currencies.ex
agustinesco: Sounds good!

agustinesco (approved):
LGTM!

ncontinanza (approved):
LGTM!

PR #677, opened 2024-06-04
[GH-463] fixes to levels with costs

ncontinanza:
## Motivation

Part of https://github.com/lambdaclass/afk_gacha_game/issues/463

## Summary of changes

- Preload the attempt_cost currency of a level before sending the campaign to the client.
- Fix the index of the slot of the dungeon levels units.
- Add max_units of a level to the proto message

## How to test it?

Start the client with this branch: https://github.com/lambdaclass/afk_gacha_game/pull/477
You should be able to spend up to 5 Supplies playing dungeon levels, and see the cost of each level in the line up scene.
You shouldn't be able to play a dungeon level if you exceed the units limit.

## Checklist
- [X] Tested the changes locally.
- [X] Reviewed the changes on GitHub, line by line.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #683, opened 2024-06-05
Fix failing dungeon test

lotuuu:
## Motivation

A failing test was introduced by #658. The reason why this escaped the CI is that the branch was not up to date with main when the passing CI was ran. To avoid this issue in the future, we could implement a restriction in the repo where only up-to-date with `main` branches can be merged. We have this setting enabled in the [AFK Gacha Game](https://github.com/lambdaclass/afk_gacha_game) repo.

In main, max_units restrictions had been introduced for levels, together with the seeds that set that field to 1 in the first level of the dungeon. This is what was missing on the wrong-succesful CI.

## Summary of changes

- Add initial supplies currencies (cherrypicks from #677)
- Unselects all units but one from the user before fighting first dungeon level

## How to test it?

Run `mix tests`

### DIFF
--- a/apps/champions/lib/champions/users.ex
+++ b/apps/champions/lib/champions/users.ex
@@ -145,7 +145,7 @@ defmodule Champions.Users do
     Currencies.add_currency(
       user.id,
       Currencies.get_currency_by_name_and_game!("Supplies", Utils.get_game_id(:champions_of_mirra)).id,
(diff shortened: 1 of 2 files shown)


agustinesco (approved):
LGTM!

ncontinanza (approved):

Issue #309, opened 2024-02-22
Add valtimer warp teleport


agustinesco:
The "dash" skill of valtimer will be a teleport to the destination instead of a dash

While the teleporting is occurring the player should not take damage

Implement the same for every dash and leap mechanic

lotuuu:
What should happen if the player takes damage while teleporting?

agustinesco:
> What should happen if the player takes damage while teleporting?

We decided that the player should not take damage while teleporting since you wont be able to see you in the animation, we talked about this with @marco-paulucci

PR #690, opened 2024-06-07
[GH-309] Implement Additive and Multiplicative modifiers

lotuuu:
## Motivation
Adds the ability to determine whether a multiplicative modifier stacks up multiplicatively or additively with others. If you're not aware of what this means, picture the following example:
You have a unit with an `attack = 1` and two multiplicative modifiers. One multiplies its' attack by 15, and the other one by 5. If both are additive, the final value is `1 * (15+5)= 20`. However, they are multiplicative, the final value is `1 * (15 * 5) = 75`.
If 15 was additive, 5 was multiplicative, and we added another additive of 25, then te equation would look like `1 * (15+ 25)* (5) 200`.
Here is the mathematical equation used:
image
Closes https://github.com/lambdaclass/afk_gacha_game/issues/309
## Summary of changes
- Add multiply_type to modifiers, defaulted to Additive in simulator.ex
- Implement additive and multiplicative calculation
- Add test
- Fix small issue in overrides where we were getting the first override applied instead of the last
## How to test it?
Play around with the values for additive and multiplicative modifiers! You can see the added test for an example.
Remember that you can add modifiers to a character's skills by modifying its entries in the `skills.json` file and running `Champions.Config.import_skills_config()`

### DIFF
--- a/apps/champions/lib/champions/battle/simulator.ex
+++ b/apps/champions/lib/champions/battle/simulator.ex
@@ -1253,7 +1253,16 @@ defmodule Champions.Battle.Simulator do
         end),
       delay: div(effect.initial_delay, @miliseconds_per_step),
       components: effect.components,
-      modifiers: Enum.map(effect.modifiers, &Map.put(&1, :skill_id, skill_id)),
+      modifiers:
+        Enum.map(
+          effect.modifiers,
+          fn modifier ->
+            modifier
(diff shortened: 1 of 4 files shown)


agustinesco (approved):
LGTM! i debugged the multiplier values and the seem to be correct!

PR #691, opened 2024-06-07
Limit amount of bounties sent to the client

agustinesco:
## Motivation

We're sending every bounty in the backend db when the client is expecting to receive only 3 
Also the client need the selecting stage timer and the start game timer

## Summary of changes

Limit amount of bounties sent to the client
Send needed timers

## How to test it?

Use this client pr -> https://github.com/lambdaclass/champions_of_mirra/pull/1791
1. run `GameBackend.CurseOfMirra.Config.import_quest_descriptions_config` in the backend terminal
2. wait for the bounty fetcher to retrieve bounties, 30 sec max
3. Start a match and you should see the 3 bounties that are created in the quest_descriptions.json file

## Checklist


Nico-Sanchez (changes requested):
Let's fix the "deal 300 damage" bounty, the value is set to 3000.
Also, the quests config is quite difficult to modify and very error prone, you have to know the field to track (user shouldn't know about this) and make sure you combine the 4 fields properly. I created an issue for this, please read it and complete it so we do it in another PR.

Nico-Sanchez (approved):

tvillegas98 (approved):
Looks good! We have to wait until the front PR is approved to merge this one

PR #701, opened 2024-06-12
[GH-699] Add environment default options in quick game

### LINKED ISSUE #699: Disable bots and zone by default when using quick game option

Nico-Sanchez:
## Motivation

Zone and bots can be toggled in client settings, but we need a faster way to test things without these kind of features.
Closes #699 

## Summary of changes

[Make zone and bots configurable and toggle them off in quick match](https://github.com/lambdaclass/mirra_backend/pull/701/commits/a7c7904c245aabff3f990b4cb28c7f3dc9725765)

## How to test it?

Play a quick match, both zone and bots should be disabled.
Play a regular match, both zone and bots are configurable in `apps/arena/priv/config.json`.

## Checklist

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -661,7 +661,7 @@ defmodule Arena.GameUpdater do
       |> Map.put(:zone, %{
         radius: config.map.radius,
         enabled: config.game.zone_enabled,
-        shrinking: config.game.zone_enabled,
+        shrinking: false,
         next_zone_change_timestamp:
           initial_timestamp + config.game.zone_shrink_start_ms + config.game.start_game_time_ms
       })
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -59,10 +59,6 @@ defmodule Arena.GameUpdater do
     game_state = new_game(game_id, clients ++ bot_clients, game_config)
     match_id = Ecto.UUID.generate()
 
-    unless game_config.game.bots_enabled do
-      Process.send_after(self(), :toggle_bots, 300)
-    end
-
     send(self(), :update_game)
     Process.send_after(self(), :selecting_bounty, game_config.game.bounty_pick_time_ms)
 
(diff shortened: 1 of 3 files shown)


apps/arena/lib/arena/game_updater.ex
```diff
@@ -653,8 +660,8 @@ defmodule Arena.GameUpdater do
       |> Map.put(:external_wall, Entities.new_external_wall(0, config.map.radius))
       |> Map.put(:zone, %{
         radius: config.map.radius,
-        enabled: false,
-        shrinking: false,
+        enabled: config.game.zone_enabled,
+        shrinking: config.game.zone_enabled,
```
agustinesco, line 664: I think the shriking should always be false, this isn't related to the zone being enabled or not, it's related to the zone shrinking

apps/arena/lib/arena/game_updater.ex
```diff
@@ -51,12 +51,18 @@ defmodule Arena.GameUpdater do
...
+  def init(%{clients: clients, bot_clients: bot_clients, game_params: game_params}) do
     game_id = self() |> :erlang.term_to_binary() |> Base58.encode()
     game_config = Configuration.get_game_config()
+    game_config = Map.put(game_config, :game, Map.merge(game_config.game, game_params))
+
     game_state = new_game(game_id, clients ++ bot_clients, game_config)
     match_id = Ecto.UUID.generate()
 
+    unless game_config.game.bots_enabled do
+      Process.send_after(self(), :toggle_bots, 300)
```
agustinesco, line 63: Why this 300? is this to give time to the bots to init? if that's the case i think this is too error prone, maybe we could think in another approach, maybe sending something in the bot init message when the match start to they wake up sleep.

apps/arena/lib/arena/game_updater.ex
Nico-Sanchez: You're right done.

apps/arena/lib/arena/game_updater.ex
Nico-Sanchez: Yeah, the 300 ms seems random. I moved it to the game_start event.
Moving it to the init bot ws message would be a longer solution since we don't read the config there. I believe it's fine and faster just do it in the game start, since we have the config there.

agustinesco (approved):
LGTM!

tvillegas98 (approved):
Builded and tested with a deploy in Brazil! Great work

PR #703, opened 2024-06-12
[GH-678] Fix bot authentication

### LINKED ISSUE #678: Fix authentication gap due to unauthenticated bots
#665 left an open issue for authentication and bots, currently we still support unauthenticated joining a match because bots don't have an authentication mechanism

To solve this we should have a way to generate a bot JWT that we can use for bots to connect, this JWT doesn't have to be unique per game nor per bot. We can have a sort of global JWT used by all bots and that we cache in the bot_manager app. Bonus point, if we reuse a `device_id` then we simply use the existing `/auth/refresh-token` endpoint for refreshing this token

AminArria:
## Motivation

Closes #678 

## Summary of changes

- Create endpoint `/auth/generate-bot-token` in gateway to generate tokens for bots to use
- Create a gen_server `TokenFetcher` in bot_manager to generate a JWT every 30 minutes (token expires ~2 hours)
- Remove hack in arena `GameSocketHandler` to force any new connection to have either the user query param (`gateway_jwt`) or the bot query params (`bot_token` and `bot_secret`)

## How to test it?

Play the game and if bots keep working then this is working 😄 

## Checklist


mFragaBA:
Closing since bots get merged into arena (check #1136) so all the authentication logic is now not needed for bots

PR #712, opened 2024-06-24
[GH-676] Make skills auto aim to crates too

### LINKED ISSUE #676: Include crates in auto-aim check.
Auto-aim should work with crates.

However, players should always be a priority if there is a player and a crate within autoaim range.

agustinesco:
## Motivation

Skills should auto aim to hit crates too, It should prioritize players over crates in the auto aim check

Closes #676

## Summary of changes

- Check for crates in the autoaim if no players are found
- This check to see if we were executing an auto aim had two issues: `nearest_entity_position_in_range != player.direction`
  - Sometimes you returned a position that can be not normalized, so this will never match since the player direction is normalized
  - If you normalized the `nearest_entity_position_in_range` you encountered precision errors
  - Now we'll check in rust if we'll auto aim or not and return it instead of comparing in elixir

## How to test it?

Start a match and try to hit a crate without aiming, if you have a player nearby it should prioritize it 

## Checklist

### DIFF
--- a/apps/arena/lib/arena/game/player.ex
+++ b/apps/arena/lib/arena/game/player.ex
@@ -3,6 +3,7 @@ defmodule Arena.Game.Player do
   Module for interacting with Player entity
   """
 
+  alias Arena.Game.Crate
   alias Arena.GameUpdater
   alias Arena.GameTracker
   alias Arena.Utils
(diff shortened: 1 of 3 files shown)


AminArria (approved):

apps/arena/lib/arena/game/player.ex
```diff
@@ -175,6 +176,13 @@ defmodule Arena.Game.Player do
         {auto_aim?, skill_direction} =
           skill_params.target
           |> Skill.maybe_auto_aim(skill, player, targetable_players(game_state.players))
+          |> case do
+            {false, _} ->
+              Skill.maybe_auto_aim(skill_params.target, skill, player, Crate.interactable_crates(game_state.crates))
```
tvillegas98, line 181: Why we need this case? 🤔 
I wonder if we can do something like `targetable_players(game_state.players) ++ Crate.interactable_crates(game_state.crates))`

agustinesco: The thing is that you want to prioritize player over crates, this case match if there's no player that you can auto aim, so you try to auto aim a crate

tvillegas98: Okay, lets iterate but I think we can find a way to improve this an order the entities by some kind of priority

tvillegas98 (approved):
Nice catch!

PR #728, opened 2024-06-27
[GH-727] Fix match register for old users

### LINKED ISSUE #727: Matches not being registered for google users
matches are not being registered when you play with a google user debug and fix

agustinesco:
## Motivation

Every time you played a match they wasn't being saved in the backend. This was due to the `create_arena_match_results` failing on the `give_prestige` step due to old users not having the unit to update. We have the following error in the backend:

``` 
** (exit) an exception was raised:
Jun 27 19:21:13 mirra-central-europe-testing entrypoint.sh[1388105]:     ** (CaseClauseError) no case clause matching: {:error, {:update_prestige, "a8875c87-7e09-48e3-9ab5-06e059a6a69e"}, :unit_not_found,
```

The behavior should be to create the unit if that user doesn't have it, this would be an issue in the future when we add new units to the db

Closes #727

## Summary of changes

- Create unit instead on failing when not present in the user when running `create_arena_match_results`

## How to test it?

- Log in as Guest
- Run this command `GameBackend.Repo.delete_all(GameBackend.Units.Unit)`
- Play a match and it should be registered

## Checklist


AminArria (dismissed):

apps/game_backend/lib/game_backend/curse_of_mirra/matches.ex
```diff
@@ -55,7 +56,15 @@ defmodule GameBackend.CurseOfMirra.Matches do
...
-            {:error, :unit_not_found}
+            unit_params =
+              Units.get_unit_default_values(result["character"])
+              |> Map.put(:user_id, user.id)
+
+            reward = match_prestige_reward(unit_params, result["position"], prestige_config[:rewards])
+            changes = calculate_rank_and_amount_changes(unit_params, reward, prestige_config[:ranks])
+
+            Map.merge(unit_params, changes)
+            |> Units.insert_unit()
```
Nico-Sanchez, line 67: I think this is bad for several reasons:
1. The function is called "give_prestige" and it inserts you a unit if you don't have it, which is way more than giving prestige (beyond responsibility).
2. The {:error, :unit_not_found} was intended. What happens if a user lose a unit due to some bug? We'll never notice since we're always inserting it here over and over.
3. Players shouldn't be allowed to play with a Unit they don't have. If I don't have some character unlocked and I manage to play with it somehow (bug), I'll be rewarded with it.

Nico-Sanchez (changes requested):
I don't find this fix accurate. I think we should fix this in another way.
I think we should insert/update/delete units to users accordingly, and also do the proper checks, instead of just patching it. It seems ok now, but it could lead to a worse bug in the future.
Finally, I would do the fix keeping in mind two things we mentioned in this PR:
> ... in the future when we add new units to the db.
It's important to make a fix that contemplates the behavior of new units, maybe inserting them only when they are unlocked, or everytime we add it (we need a boolean for unlocked/locked with this approach).
> Players shouldn't be allowed to play with a Unit they don't have.
I believe that's important, it will break more than the prestige in the future if we don't do that check. Players can select a character in the client, they should be picking the unit's character there.

apps/game_backend/lib/game_backend/curse_of_mirra/matches.ex
agustinesco: 1. Fair point, maybe moving this to a different multi operation is better
2. 3. We don't have an instance of "Log in" or "Character unlock" to assign this units right now, they're just broken. Maybe this may work as a patch until we develop those, i can create an issue and will be worked in the future when we have time.

agustinesco:
> I don't find this fix accurate. I think we should fix this in another way.
> I think we should insert/update/delete units to users accordingly, and also do the proper checks, instead of just patching it. > It seems ok now, but it could lead to a worse bug in the future.
> Finally, I would do the fix keeping in mind two things we mentioned in this PR:

As i mentioned in the pr we don't have any "unlock" mechanic yet to create this unit in the user, that may come in a near future but this is breaking the game now, maybe we can create every missing unit for the user that logs in through the routes  `create_guest_user` or `validate_token`  instead of creating them in the match result insert.
Every fix will end up in a "will be fixed in a near future when we have X" imo, so i wouldn't look for a permanent solution right now. We can talk about this

apps/game_backend/lib/game_backend/curse_of_mirra/matches.ex
Nico-Sanchez: Due to our need to fix this we should:
1. Move this whole "patch" to another function. Encapsulating this behavior will help us in the future to remove it.
2. Create an issue to reimplement this properly. Also, attach its link in the code.

Nico-Sanchez (approved):

AminArria (approved):
LGTM

PR #733, opened 2024-06-28
[Gh-732] Add missing bounties features

### LINKED ISSUE #732: Add missing bounty feature functionalities
- We should send a default bounty to the client so they can display which bounty is picked if the pick time runs out
- We should notify the client in game if the bounty was completed
- At the end of a match the client needs to know if the picked bounty was completed

agustinesco:
>[!warning]
> This pr should be merged alongside [This Client PR](https://github.com/lambdaclass/champions_of_mirra/pull/1941)

## Motivation

The bounty feature was not completed so we need to work on some points
- Bounty completion in game
- Default bounty communicated to the client
- Add a user information retrieve endpoint to display current currencies amounts

Closes #732

## Summary of changes

- Added a new check on the tick that validates if the user has completed his bounty, i have my doubts in the way i implemented this:
  - IMO the game_update should be the one in charge of this a player completing a bounty is something related to the game_state, we can discuss this point
- Added call method to game_tracker to retrieve the current result of a player
- Added default bounty to the player entity and used that to pick a bounty if the pick time runs out

### Default bounty
- Start a match and don't pick a bounty
- Play the match as usual 

### Selecting a bounty
- Start a match and pick a bounty
- Try to complete that bounty

if you want to debug what's being received in the client you can put this in the `GameServerConnectionManager` file in the client repo in line 193. 
```
                    foreach (Entity player in gameState.Players.Values.ToList())
                    {
                        if (player.Id == this.playerId){
                            Debug.Log(player.Player.BountyCompleted);
                        }
                    }
```

## Checklist


apps/arena/mix.exs
```diff
@@ -52,7 +52,8 @@ defmodule Arena.MixProject do
       {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
       {:plug_cowboy, "~> 2.5"},
       {:toxiproxy_ex, "~> 1.1.1"},
-      {:joken, "~> 2.6"}
+      {:joken, "~> 2.6"},
+      {:game_backend, in_umbrella: true}
```
Nico-Sanchez, line 56: Arena release doesn't have :game_backend app deployed. We're basing all our development towards that premise.
I believe we should add a way to develop this feature without this strong dependency. If we do this, the umbrella will lose most of its meaning.

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -274,29 +316,37 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+    false
+  end
+
+  def completed_quest?(quest, arena_match_results) do
     progress =
       arena_match_results
       |> filter_results_that_meet_quest_conditions(quest.conditions)
       |> Enum.reduce(0, fn arena_match_result, acc ->
-        type = String.to_atom(quest.objective["match_tracking_field"])
+        type = String.to_existing_atom(quest.objective["match_tracking_field"] || quest.objective.match_tracking_field)
```
Nico-Sanchez, line 328: Are you certain these atoms exist everytime at this point? Otherwise we could do a wiser approach by creating the atoms explicitly. For example: `def type_atom("foo"), do :foo, def type_atom("bar"), do :bar, ...`

apps/arena/lib/arena/entities.ex
```diff
@@ -55,7 +55,10 @@ defmodule Arena.Entities do
         visible_players: [],
         on_bush: false,
         bounties: [],
-        bounty_selected: false
+        bounty_selected: false,
+        selected_bounty: nil,
+        default_bounty: nil,
+        bounty_completed: false
```
Nico-Sanchez, line 61: Do we need all these booleans? Maybe we could merge some of them:
1. Do we need selected_bounty and default_bounty separately? Can't the default take its place then the user overrides it (select)?
2. Do we need a boolean for bounty_selected? Doesn't the nil in selected bounty suffix that?

Nico-Sanchez (changes requested):

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: Since they are a required field from the match_result table i think they're created yes, do you think this could crash at some point?

apps/arena/lib/arena/entities.ex
agustinesco: 1. I deleted the default bounty and now we'll do the random pick when the pick time runs out and the `selected_bounty ` is nil as you suggested
2. nope, this is related to the 1. point where we used this boolean to actually pick a default bounty, removed

apps/arena/mix.exs
agustinesco: I reverted this to a version where we have a `bounties.ex` module in arena with the quest check logic, maybe we're repeating code but it's the best way i found to have this check in arena too, maybe something better would be to have another app that's used as a library to be used in both gamebackend and arena, but i would iterate this, what do you think?

Nico-Sanchez: I agree. Let's merge but fix later. Create an issue please.

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
Nico-Sanchez: I think you misunderstood my comment. I'm not talking about the field in the database/schema. I'm talking about the Atom you're trying to cast to.

Let's say objective["field"] has the value "value". If you do String.to_existing_atom("value") and :value doesn't exists, it will crash.
I believe is way better to explicit the atom result so we:
1. Use it if it exists.
2. Define it in a controlled scope (only the ones we define will be created).

apps/arena/lib/arena/game/bounties.ex
```diff
@@ -0,0 +1,49 @@
...
+  def completed_bounty?(nil, _arena_match_results) do
+    false
+  end
+
+  def completed_bounty?(quest, arena_match_results) do
+    progress =
+      arena_match_results
+      |> filter_results_that_meet_quest_conditions(quest.conditions)
+      |> Enum.reduce(0, fn arena_match_result, acc ->
+        type = String.to_atom(quest.objective.match_tracking_field)
```
Nico-Sanchez, line 26: This is HIGHLY DISCOURAGED.
I already mentioned something regarding Atoms in another comment. Remember atoms ARE NOT garbage collected, we must not create atoms in an uncontrolled way like this one.

apps/arena/lib/arena/game/bounties.ex
```diff
@@ -0,0 +1,49 @@
...
+
+    comparator = parse_comparator(quest.objective.comparison)
+
+    comparator.(progress, quest.objective.value)
+  end
+
+  defp filter_results_that_meet_quest_conditions(arena_match_results, conditions) do
+    Enum.filter(arena_match_results, fn arena_match_result ->
+      Enum.all?(conditions, fn condition ->
+        type = String.to_atom(condition.match_tracking_field)
```
Nico-Sanchez, line 39: Same here.

Nico-Sanchez (changes requested):

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: Thanks for explaining! 
what you meant was to create these atoms in a controlled environment where we can define a limit and make sure they won't crash in any unexpected scenario, correct me if i'm wrong

apps/arena/mix.exs
agustinesco: Sure thing! issue created here -> https://github.com/lambdaclass/mirra_backend/issues/774

apps/arena/lib/arena/game/bounties.ex
agustinesco: Thanks for clarifying this! I took the approach you mentioned to create a function to pattern match the field and turn it into an atom if we want to, wrong fields will be taken as invalid and won't be taken into account when checking if a quest is completed

apps/arena/lib/arena/game/bounties.ex
agustinesco: Done!

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -292,14 +337,34 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+  defp arena_match_result_field_to_atom("character"), do: :character
+  defp arena_match_result_field_to_atom("match_id"), do: :match_id
+  defp arena_match_result_field_to_atom("user_id"), do: :user_id
+  defp arena_match_result_field_to_atom("position"), do: :position
+  defp arena_match_result_field_to_atom("damage_done"), do: :damage_done
+  defp arena_match_result_field_to_atom("damage_taken"), do: :damage_taken
+  defp arena_match_result_field_to_atom("health_healed"), do: :health_healed
+  defp arena_match_result_field_to_atom("killed_by_bot"), do: :killed_by_bot
+  defp arena_match_result_field_to_atom("duration_ms"), do: :duration_ms
+  defp arena_match_result_field_to_atom(_), do: nil
```
Nico-Sanchez, line 369: All these are defined here and in apps/arena/lib/arena/game/bounties.ex. Can you put them in some utils module or something?

Nico-Sanchez: Oh one is for arena and the other one for game_backend, nvm.

Nico-Sanchez (approved):

tvillegas98 (dismissed):
Seems to work well!

apps/arena/lib/arena/game/bounties.ex
```diff
@@ -0,0 +1,78 @@
+defmodule Arena.Game.Bounties do
+  @moduledoc """
+  Module to handle quest logic in arena application
+  TODO this module shouldn't exist, this a workaround until we have a shared application to
```
AminArria, line 4: Why not add it as a dependency library?

```elixir
{:game_backend, in_umbrella: true, runtime: false}
```

apps/arena/lib/arena/game_tracker.ex
```diff
@@ -144,6 +155,37 @@ defmodule Arena.GameTracker do
...
+    %{}
+  end
+
+  defp generate_player_result(match_data, player_id) do
+    duration = System.monotonic_time(:millisecond) - match_data.start_at
+
+    player_data =
+      Map.get(match_data.players, player_id)
+
+    %{
```
AminArria, line 168: Can we extract this and the one in `generate_results/2` so we don't need to modify the same map in 2 places?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -105,7 +106,14 @@ defmodule Arena.GameUpdater do
...
         state =
-          put_in(state, [:game_state, :players, player_id, :aditional_info, :bounties], bounties)
+          update_in(state, [:game_state, :players, player_id, :aditional_info], fn aditional_info ->
+            if Enum.empty?(bounties) do
+              aditional_info
+            else
+              aditional_info
+              |> Map.put(:bounties, bounties)
+            end
+          end)
```
AminArria, line 116: Can you explain a bit on why is this needed or what it is doing?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -1230,6 +1246,45 @@ defmodule Arena.GameUpdater do
...
+          path = "/curse/users/#{user_id}/quest/#{player.aditional_info.selected_bounty.id}/complete_bounty"
+          gateway_url = Application.get_env(:arena, :gateway_url)
+
+          Finch.build(:get, "#{gateway_url}#{path}")
+          |> Finch.request(Arena.Finch)
+        end)
+
+        update_in(game_state, [:players, player_id, :aditional_info], fn aditional_info ->
+          Map.put(aditional_info, :bounty_completed, true)
+        end)
```
AminArria, line 1277: Why not?

```elixir
put_in(game_state, [:players, player_id, :aditional_info, :bounty_completed], true)
```

apps/arena/lib/arena/game_updater.ex
```diff
@@ -1230,6 +1246,45 @@ defmodule Arena.GameUpdater do
...
+    actual_players =
+      state.clients
+      |> Enum.map(fn client_id ->
+        player_id =
+          get_in(game_state, [:client_to_player_map, client_id])
+
+        player = get_in(game_state, [:players, player_id])
+
+        {player_id, player}
+      end)
```
AminArria, line 1259: I don't get why you need, isn't this exactly the same as `Map.to_list(game_state.players)`? And if its only to use it in `Enum.reduce/3` then you should be able to just past and it will give you the same

apps/gateway/lib/gateway/controllers/curse_of_mirra/user_controller.ex
```diff
@@ -10,6 +10,18 @@ defmodule Gateway.Controllers.CurseOfMirra.UserController do
 
   action_fallback Gateway.Controllers.FallbackController
 
+  def show(conn, %{"id" => user_id}) do
+    game_id = Utils.get_game_id(:curse_of_mirra)
+
+    with {:get_user, {:ok, user}} <- {:get_user, Users.get_user(user_id)},
+         {:curse_user, ^game_id} <- {:curse_user, user.game_id} do
```
AminArria, line 17: If we are doing this check then just make a function to fetch by `user_id` and `game_id`, relying on this check is not good

apps/gateway/lib/gateway/controllers/curse_of_mirra/user_controller.ex
```diff
@@ -10,6 +10,18 @@ defmodule Gateway.Controllers.CurseOfMirra.UserController do
...
+      send_resp(conn, 200, Jason.encode!(user))
+    else
+      {:get_user, _} -> send_resp(conn, 404, "User not found")
+      {:curse_user, _} -> send_resp(conn, 400, "User from another game")
```
AminArria, line 21: Don't do this, use the `FallbackController` and handle things there

References:
- https://hexdocs.pm/elixir/code-anti-patterns.html#complex-else-clauses-in-with
- https://hexdocs.pm/phoenix/json_and_apis.html#action-fallback

AminArria (changes requested):

apps/arena/lib/arena/game_updater.ex
agustinesco: Whoops, this was something i needed to pick the `default_bounty` since if the bounties list is empty the `Enum.random/1` method would crash

apps/arena/lib/arena/game/bounties.ex
agustinesco: We talked about this with @Nico-Sanchez , in the end, we'll go for the option to create a new application to contain the champion's logic and both GameBackend and Arena will take the bounties logic from there.
But not in this pr, we wan't to have it in the short term

apps/arena/lib/arena/game_tracker.ex
agustinesco: sure thing!

apps/arena/lib/arena/game_updater.ex
agustinesco: nope, it's not the same as doing game_state.players since here we only got the real players and not bot players, since we don't care about the bots getting currencies. It's something similar to what game_tracker is doing since the `clients` field from the `game_state` only has humans. 
I'll rename `actual_players` to `human_players` to avoid confusion

apps/gateway/lib/gateway/controllers/curse_of_mirra/user_controller.ex
agustinesco: Done! I created a `get_user_by_id_and_game_id` function, also i noticed that we're always doing two queries when retrieving users, one to get it and other one to preload some fields, i created this function in a way we only run one query

apps/gateway/lib/gateway/controllers/curse_of_mirra/user_controller.ex
agustinesco: Done! thanks for the references

apps/arena/lib/arena/game_updater.ex
agustinesco: Yeah! that's much better

AminArria (approved):

Nico-Sanchez (approved):

PR #742, opened 2024-07-02
[GH-723] Add Consumable items Configuration

### LINKED ISSUE #723: [FEATURE] Configurator WebApp: Add Items/Consumables
We have to migrate the data from the key `items` in `game_config.json` to a new schema. In `GameBackend` we already have an `Item` schema but it doesn't represent the logic of the items we use in Champions of Mirra.

This schema could be named like `consumable_items` or just `consumables`.

tvillegas98:
## Motivation

Migrate items from the `config.json` file to a relational database. 

Closes #723 

## Summary of changes

- Although we call them `items` these items are more likely to be `consumable items` since they only exists during a match. So, I've added the `ConsumableItem` MVC 
- Added a LiveView form for consumable items as a basis for further implementations
- The new config for items has been integrated into the game, you'll see some changes in arena's logic.
- We will only send to the client items that are active
DISCLAIMER: We still have pending to add the relationship between items and effects, but that's going to be addressed in another issue #747 . At the moment it is just a map displayed in the view

## How to test it?

Go to `localhost:4100/consumable_items` modify or add current items (be careful, our client could crash because we have to add the needed assets for a new item) and then play a match and check that everything works!.

## Checklist


agustinesco (changes requested):
There're a few things to fix:
1. when i try to delete an item i got an error:
image
2. If you try to edit an item and click on submit the mechanics field triggers an error and you can change anything, even tho you didn't changed the mechanics field

https://github.com/user-attachments/assets/f826d9c6-e536-4d48-ac33-f8c67b5934d1

3. When you try to create an item you got an error related to the number 2. but this time the changes you do on the mechanics field are lost and replaced with "null":

https://github.com/user-attachments/assets/0a1095ef-5c35-474c-818c-dbc1640fe551

5. In the characters view the whole row works as a "show" button, but in this view we have an explicit show button, i think we should keep consistency and do the same as the characters index, what do you think?
6. I can add effects to the items but there's no way to actually delete an effect from an item
7.  we should change the title to display the item name instead of the id
image

8. For some reason the "back to consumable_items" button doesn't work

https://github.com/user-attachments/assets/381c761e-118d-410b-ba4b-2878dd4ad8ed

tvillegas98:
> If you try to edit an item and click on submit the mechanics field triggers an error and you can change anything, even tho you didn't changed the mechanics field

Great! that's a nice catch, the reason behind is that when submitting the form, I'm not encoding the json string into a proper map

tvillegas98:
> When you try to create an item you got an error related to the number 2. but this time the changes you do on the mechanics field are lost and replaced with "null"

That was really weird, fixed in [4b4d66a](https://github.com/lambdaclass/mirra_backend/pull/742/commits/4b4d66a34975dfd231bf5996d0bd79aeb475b158)

tvillegas98:
> In the characters view the whole row works as a "show" button, but in this view we have an explicit show button, i think we should keep consistency and do the same as the characters index, what do you think?

I rather remove the row_click from characters since it conflicts with modal buttons

tvillegas98:
> we should change the title to display the item name instead of the id
Done in [0ea8d3f](https://github.com/lambdaclass/mirra_backend/pull/742/commits/0ea8d3f08f7ef6e707550b6639278c300d64a40d)

tvillegas98:
> For some reason the "back to consumable_items" button doesn't work

Fixed in [7f5bcca]
(https://github.com/lambdaclass/mirra_backend/pull/742/commits/7f5bcca42ab11c77aed5710e35131843873d9b96)

Really weird to be honest, I had to check this issue thread to solve it https://github.com/phoenixframework/phoenix_live_view/issues/2832

agustinesco (changes requested):
The edit for the items have the id issue too:
image

When i try to create a new item and delete the first braket of the mechanic field i encounter an error, maybe we should fix that

agustinesco (approved):
LGTM! great job

apps/arena/lib/arena/game/player.ex
```diff
@@ -315,15 +315,14 @@ defmodule Arena.Game.Player do
...
-  def use_item(player, game_state, game_config) do
+  def use_item(player, game_state) do
     case player.aditional_info.inventory do
       nil ->
         game_state
 
       item ->
         game_state =
-          Enum.reduce(item.effects, game_state, fn effect_name, game_state_acc ->
-            effect = Enum.find(game_config.effects, fn %{name: name} -> name == effect_name end)
```
Nico-Sanchez, line 326: Why don't we keep this until we do the effects feature?
We could just store the effect name string, so we avoid this transition using the weird and hard-to-read map of effects.

apps/configurator/lib/configurator_web/controllers/consumable_item_html/show.html.heex
```diff
@@ -0,0 +1,28 @@
+<.header>
+  Consumable item <%= @consumable_item.name %>
+  <:subtitle>This is a consumable_item record from your database.</:subtitle>
+  <:actions>
+    <.link href={~p"/consumable_items/#{@consumable_item}/edit"}>
+      <.button>Edit consumable_item</.button>
```
Nico-Sanchez, line 6:
```suggestion
      <.button>Edit Consumable Item</.button>
```

apps/configurator/lib/configurator_web/controllers/consumable_item_html/new.html.heex
```diff
@@ -0,0 +1,6 @@
+<.header>
+  New Consumable item
+  <:subtitle>Use this form to manage consumable_item records in your database.</:subtitle>
```
Nico-Sanchez, line 3:
```suggestion
  <:subtitle>Use this form to manage Consumable Item records in your database.</:subtitle>
```

apps/configurator/lib/configurator_web/controllers/consumable_item_html/edit.html.heex
```diff
@@ -0,0 +1,6 @@
+<.header>
+  Edit Consumable item <%= @consumable_item.name %>
+  <:subtitle>Use this form to manage consumable_item records in your database.</:subtitle>
```
Nico-Sanchez, line 3:
```suggestion
  <:subtitle>Use this form to manage Consumable Item records in your database.</:subtitle>
```

apps/configurator/lib/configurator_web/controllers/consumable_item_html/show.html.heex
Nico-Sanchez, line 1: This form doesn't have any navigation button, you can't exit this view unless you change the url explicitly

apps/game_backend/lib/game_backend/items/consumable_item.ex
```diff
@@ -0,0 +1,29 @@
...
+  alias GameBackend.Effects.ConfigurationEffect
+
+  @derive {Jason.Encoder, only: [:name, :radius, :effects, :mechanics]}
+  schema "consumable_items" do
+    field(:active, :boolean, default: false)
+    field(:name, :string)
+    field(:radius, :float)
+    field(:mechanics, {:map, :map})
+
+    has_many(:effects, ConfigurationEffect, on_replace: :delete)
```
Nico-Sanchez, line 17: Why is this called ConfigurationEffect and not just Effect? Is this a temporary fix?
I think we could just use a list of strings (effect names) then create the effects table with the mechanics relation all at once, since the skills PR is already adding the Mechanic schema.

Nico-Sanchez (changes requested):
Left some comments. I think we can just use the effect names as strings then add the mechanics later (there's a PR adding that DB schema and relationships).
In doing so, we avoid going a step back from our current implementation. The effects map is almost unreadable, hence difficult to edit or know what I'm doing. I find it easier to read and edit in our current JSON. We should iterate this without affecting our current performance, UX or whatsoever.

apps/configurator/lib/configurator_web/controllers/consumable_item_html/show.html.heex
tvillegas98: I think I've remove it by mistake 😨

apps/configurator/lib/configurator_web/controllers/consumable_item_html/new.html.heex
tvillegas98: Done

apps/configurator/lib/configurator_web/controllers/consumable_item_html/edit.html.heex
tvillegas98: Done

apps/configurator/lib/configurator_web/controllers/consumable_item_html/show.html.heex
tvillegas98: Done

apps/arena/lib/arena/game/player.ex
tvillegas98: You meant the mechanics map right?

apps/game_backend/lib/game_backend/items/consumable_item.ex
tvillegas98: Main reason is that already exists a schema named `effect.ex` but we can add this effect schema in another PR

tvillegas98:
> I think we can just use the effect names as strings then add the mechanics later (there's a PR adding that DB schema and relationships).
In doing so, we avoid going a step back from our current implementation. The effects map is almost unreadable, hence difficult to edit or know what I'm doing. I find it easier to read and edit in our current JSON. We should iterate this without affecting our current performance, UX or whatsoever.

Sure thing. I have removed effects logic in [0d77b92](https://github.com/lambdaclass/mirra_backend/pull/742/commits/0d77b928b3e0cc04bff0a004874ccd034b699219)
We already have to add the effects to the items here #747 so we can address effects and mechanics logic there

apps/arena/lib/arena/game/player.ex
Nico-Sanchez: Yeah

apps/game_backend/lib/game_backend/items/consumable_item.ex
Nico-Sanchez: Yep. Also remember, we can name the schema Effect, but the table could be called "configurator_effects" to differentiate them in the DB.

Nico-Sanchez:
> > I think we can just use the effect names as strings then add the mechanics later (there's a PR adding that DB schema and relationships).
> > In doing so, we avoid going a step back from our current implementation. The effects map is almost unreadable, hence difficult to edit or know what I'm doing. I find it easier to read and edit in our current JSON. We should iterate this without affecting our current performance, UX or whatsoever.
> 
> Sure thing. I have removed effects logic in [0d77b92](https://github.com/lambdaclass/mirra_backend/pull/742/commits/0d77b928b3e0cc04bff0a004874ccd034b699219) We already have an issue to add the effects to the items here #747 so we can address effects and mechanics logic there

Nice work! I know you had to delete a lot of code, bit I think it looks much more clean now and we can get a better implementation using this as base.

Nico-Sanchez (dismissed):

apps/arena/lib/arena/configuration.ex
```diff
@@ -51,6 +54,16 @@ defmodule Arena.Configuration do
...
   end
 
+  def get_consumable_items_configuration() do
+    gateway_url = Application.get_env(:arena, :gateway_url)
+
+    {:ok, payload} =
+      Finch.build(:get, "#{gateway_url}/curse/configuration/consumable_items", [{"content-type", "application/json"}])
+      |> Finch.request(Arena.Finch)
+
+    Jason.decode!(payload.body, [{:keys, :atoms}])
```
SantiagoPittella, line 64: Howdy!
I was looking at the PR and saw that this line is creating atoms dynamically (`{:keys, :atoms}` is basically a call to `String.to_atom`), which can led to a runtime crash (more info [here](https://curiosum.com/blog/elixir-anti-patterns#dynamic-atom-creation)). It is (almost) always recommended to use `String.to_existing_atom` and only allowing whitelisted atoms only.
In this case, you can change the decoding options to `[{:keys, :atoms!}]` which uses `String.to_existing_atom` under the hood, as [explained here](https://hexdocs.pm/jason/Jason.html#decode/2-options).

Feel free to reach me out.

SantiagoPittella (changes requested):

apps/arena/lib/arena/configuration.ex
tvillegas98: Thanks for the heads up! done in [043d63c](https://github.com/lambdaclass/mirra_backend/pull/742/commits/043d63cb653fab4190d038182b108ef9564cf5ca)

SantiagoPittella (dismissed):

agustinesco (changes requested):
When i try to edit any field from an item i get an error in the effect field

agustinesco (approved):
LGTM! tested playing with every item and they all worked as expected! also i was able to edit them and it worked!

Nico-Sanchez (approved):

PR #749, opened 2024-07-04
Add pool activation delay

agustinesco:
>[!Warning]
> Merge alongside this [Client pr](https://github.com/lambdaclass/champions_of_mirra/pull/1900)

## Motivation

Pools were delay damage before the cliend could display the visual representation of them, we should add a little delay before the pool start performing his mechanic to the players above it 

## Summary of changes

- Add new `status` field to the pool entity 
- Trigger a pool activation message after some delay 
- Do not add pool effects to players over when the pool isn't activated

## How to test it?

Play as either valtimer or h4ck and use you ultimate, the pool should start doing mechanics after the animation runs in the client

## Checklist

### DIFF
--- a/apps/serialization/messages.proto
+++ b/apps/serialization/messages.proto
@@ -246,6 +246,7 @@ enum  PowerUpstatus {
 
 message Pool {
   uint64 owner_id = 1;
+  PoolStatus status = 2;
 }
 
 message Bush {
@@ -278,6 +279,11 @@ enum TrapStatus{
   PREPARED = 1;
   TRIGGERED = 2;
   USED = 3;
+}
+
+enum PoolStatus{
+  ACTIVATING = 0;
+  ACTIVATED = 1;
 }
 
 /*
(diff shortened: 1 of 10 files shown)


apps/serialization/messages.proto
```diff
@@ -281,6 +282,11 @@ enum TrapStatus{
   USED = 3;
 }
 
+enum PoolStatus{
+  ACTIVATING = 0;
+  ACTIVATED = 1;
```
Nico-Sanchez, line 287: I think the following are better names, what do you think?
```suggestion
  WAITING = 0;
  READY = 1;
```

agustinesco: completely agreed

agustinesco: done!

manucamejo (dismissed):

Nico-Sanchez (dismissed):

Nico-Sanchez (approved):

manucamejo (approved):

Issue #753, opened 2024-07-05
Reimplement bot spawning into the game 


agustinesco:
We removed the fill of matches with bots and we need to add it again, also we need to add an environmental variable to turn it off

PR #754, opened 2024-07-05
Remove bot spawning

agustinesco:
## Motivation

We need to remove the feature where players are added to matches when there are not enough real players

This will be reimplemented when we fix https://github.com/lambdaclass/mirra_backend/issues/753

## Summary of changes

Remove everything related to bot spawning

## How to test it?

Start a match and you should be alone

## Checklist


AminArria:
Why are we removing this? 

Personally this is super useful for local development so I would rather have a toggle for it so in prod its disabled

agustinesco:
Will be fixed in #755

PR #760, opened 2024-07-06
Refresh effect time instead of losing item

agustinesco:
## Motivation

Currently whenever you use an item and you already have that item effect applied you lose that item effect, that's wrong

The correct behavior should be to refresh the item effect duration to his original duration, notice that we don't add up durations, take the next scenario as example:
1. You use an item that lasts 10 seconds
2. After 5 seconds you use that item again
3. The remaining duration of the effect should be 10 again 

## Summary of changes

Change effect application behavior when an item has the `one_time_application` flag in true

## How to test it?

Start a match and use an item twice, the total duration should depend in the second item usage

## Checklist

### DIFF
--- a/apps/arena/lib/arena/game/effect.ex
+++ b/apps/arena/lib/arena/game/effect.ex
@@ -24,57 +24,16 @@ defmodule Arena.Game.Effect do
   end
 
   def put_effect_to_entity(game_state, entity, owner_id, start_action_removal_in_ms, effect) do
-    last_id = game_state.last_id + 1
-
     same_applied_effect =
       Enum.find(entity.aditional_info.effects, fn entity_effect ->
         entity_effect.owner_id == owner_id and entity_effect.name == effect.name
       end)
 
-    now = System.monotonic_time(:millisecond)
-
-    updated_entity =
-      if same_applied_effect && effect.one_time_application do
-        updated_effects =
-          Enum.map(entity.aditional_info.effects, fn effect ->
-            if effect.id == same_applied_effect.id and same_applied_effect.expires_at do
-              Map.put(effect, :expires_at, now + same_applied_effect.duration_ms)
-            else
-              effect
-            end
-          end)
-
-        put_in(
-          entity,
-          [:aditional_info, :effects],
-          updated_effects
-        )
-      else
-        action_removal_at = now + start_action_removal_in_ms
-
-        expires_at =
-          case effect[:duration_ms] do
-            nil -> nil
-            duration_ms -> now + duration_ms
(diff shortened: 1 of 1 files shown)


tvillegas98 (commented):
Seems to work very well! I've a question, why don't we just update the expires_at of the effect instead of creating a new one? just wondering

agustinesco:
I couldn't think in a better updating function in that time 😅  i just did what you suggested and updates the effect instead of deleting it!

tvillegas98 (dismissed):
Great work @agustinesco !

apps/arena/lib/arena/game/effect.ex
```diff
@@ -291,4 +305,15 @@ defmodule Arena.Game.Effect do
...
+  defp refresh_effect_expires_at(effect, start_time) do
+    new_expires_at =
+      if effect.expires_at do
+        start_time + effect.duration_ms
+      else
+        nil
+      end
+
+    Map.put(effect, :expires_at, new_expires_at)
+  end
```
Nico-Sanchez, line 318: This function is called inside an if. Why can't you just use `and`?

apps/arena/lib/arena/game/effect.ex
```diff
@@ -26,16 +26,30 @@ defmodule Arena.Game.Effect do
...
+            else
+              effect
+            end
+          end)
+
+        put_in(
+          entity,
+          [:aditional_info, :effects],
+          updated_effects
+        )
```
Nico-Sanchez, line 51: If you use `and` this whole block could be the private function.

Nico-Sanchez (changes requested):
Currently the item effects stack. Is that the correct behavior?
For example, you have a giant item effect then you use a purple mirror, they stack up.

agustinesco:
> For example, you have a giant item effect then you use a purple mirror, they stack up.

I think that's the correct behavior indeed, we can double check with the team

apps/arena/lib/arena/game/effect.ex
agustinesco: yeah i think both if path can be replaced with functions, in fact i just discovered a bug, we were always increasing the game_state.last_id even when we were refreshing the expire_time. I just created two private functions for each path

apps/arena/lib/arena/game/effect.ex
agustinesco: Completely agreed

Nico-Sanchez (approved):

AminArria (approved):
LGTM

PR #764, opened 2024-07-10
Update code for loadtests

Nico-Sanchez:
## Motivation

Loadtests weren't working with current code, several changes made in arena broke the loadtests app.

## Summary of changes

- [Add new OVERRIDE_JWT env var to avoid validation for loadtest's bots](https://github.com/lambdaclass/mirra_backend/commit/54cf5e2aa9eb23a7139796059bf6ef925ef205d2)
- [Update loadtest ws message handler due to changes in messages](https://github.com/lambdaclass/mirra_backend/commit/f858df2f775415c8f07286113d1bcf52127ceaa2)

## How to test it?

Run some loadtests and see if it works. Remember to set the new env var.
You can follow loadtest's README to do so.

## Checklist


tvillegas98 (changes requested):
It doesn't seem to work straightforwardly yet. Here is the list of things I have to do to use it.
- There's a function named `ArenaLoadTest.Utils.get_server_ip/1) that might not be longer needed (because the env var `TARGET_SERVER` now uses a public URL. This caused the load test to crash because the pattern matching doesn't catch this
- In the load test server, I had to set the env var OVERRRIDE_JWT to true (I understand that this is going to be solved after we merged this PR)

tvillegas98:
As a heads up too, we might have to update the current `load_testing.md` documentation

Nico-Sanchez:
> As a heads up too, we might have to update the current `load_testing.md` documentation

Hmm I believe that was the old loadtest docs, we moved it when we split the repos.
The docs I maintain is the README. We could delete it or leave it for further improvements (it has useful info like the tooling section).

Nico-Sanchez:
> It doesn't seem to work straightforwardly yet. Here is the list of things I have to do to use it.
> 
> 1. There's a function named `ArenaLoadTest.Utils.get_server_ip/1)` that might not be longer needed (because the env var `TARGET_SERVER` now uses a public URL. This caused the load test to crash because the pattern matching doesn't catch this
> 2. In the load test server, I had to set the env var `OVERRRIDE_JWT` to true (I understand that this is going to be solved after we merged this PR)
> 
> I think we only need to solve 1)

Done both!

tvillegas98:
> > As a heads up too, we might have to update the current `load_testing.md` documentation
> 
> Hmm I believe that was the old loadtest docs, we moved it when we split the repos. The docs I maintain is the README. We could delete it or leave it for further improvements (it has useful info like the tooling section).

Thanks for the clarification! I thought that the information would be in that markdown. Maybe we can merge both markdowns 😄

tvillegas98 (approved):
Great work fixing load test! I only had to run the spawn players function to execute it!


https://github.com/user-attachments/assets/b9c3cf1d-dbd8-4aca-b602-a4dd54aafeb9

agustinesco (approved):
LGTM! tested deploying both servers and the load tests runned like a charm

PR #768, opened 2024-07-11
[GH-767] Fix items and power ups spawning outside of the map

### LINKED ISSUE #767: Fix power ups and items spawning outside of the map boundaries

agustinesco:
## Motivation

When we spawned an item or a power up they could spawn a little bit outside of the map boundaries so you can't pick them up.
This was due to our random position generation solving the position only if it's outside of the map, even if the majority of the area is outside of it 

Closes #767

## Summary of changes

- Always resolver collision when generating random positions instead of checking map inclussion

## How to test it?

Start a match and kill some players close to the map boundaries, every power up should spawn in a pickupeable position, the same should happen to items

## Checklist

### DIFF
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -1552,24 +1552,15 @@ defmodule Arena.GameUpdater do
       name: "Circle 1"
     }
 
-    entities_to_collide_with =
+    static_obstacles =
       Map.filter(obstacles, fn {_obstacle_id, obstacle} -> obstacle.aditional_info.type == "static" end)
-      |> Map.merge(%{external_wall.id => external_wall})
 
-    external_wall_id = external_wall.id
-
-    case Physics.check_collisions(circle, entities_to_collide_with) do
-      [^external_wall_id | []] ->
-        circle.position
-
-      _ ->
-        Physics.get_closest_available_position(
-          circle.position,
-          circle,
-          external_wall,
-          entities_to_collide_with
-        )
-    end
+    Physics.get_closest_available_position(
+      circle.position,
+      circle,
+      external_wall,
+      static_obstacles
+    )
   end
 
   defp update_visible_players(%{players: players, bushes: bushes} = game_state, game_config) do
--- a/apps/arena/lib/arena/game_updater.ex
+++ b/apps/arena/lib/arena/game_updater.ex
@@ -1553,7 +1553,7 @@ defmodule Arena.GameUpdater do
     }
 
     static_obstacles =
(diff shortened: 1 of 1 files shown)


apps/arena/lib/arena/game_updater.ex
```diff
@@ -1552,24 +1552,15 @@ defmodule Arena.GameUpdater do
       name: "Circle 1"
     }
 
-    entities_to_collide_with =
+    static_obstacles =
```
Nico-Sanchez, line 1555: Shouldn't we add active dynamic obstacles here?

apps/arena/lib/arena/game_updater.ex
```diff
@@ -1552,24 +1552,15 @@ defmodule Arena.GameUpdater do
...
       Map.filter(obstacles, fn {_obstacle_id, obstacle} -> obstacle.aditional_info.type == "static" end)
-      |> Map.merge(%{external_wall.id => external_wall})
-
-    external_wall_id = external_wall.id
-
-    case Physics.check_collisions(circle, entities_to_collide_with) do
-      [^external_wall_id | []] ->
```
Nico-Sanchez, line 1562: Why was this check here? Are we losing an edge case handler by removing this?

Nico-Sanchez (changes requested):
Liftable entities can spawn inside dynamic obstacles still, I believe we have to include those.

apps/arena/lib/arena/game_updater.ex
agustinesco: This was something used by the recursive implementation and caused trouble in both errors:
- Infinite loop
- Items spawning outside of the map

apps/arena/lib/arena/game_updater.ex
agustinesco: This was some creative freedom from our part, there could be some risk in taking power ups in spikes about to go up, if you think this looks more like a bug than a feature i could change it

Nico-Sanchez: Maybe I wasn't clear enough. I think we should add the dynamic obstacles that are ACTIVE (you can't pass through) so the liftable entity can't go through as well.
It's ok to ignore it if the obstacle is off (underground).

agustinesco: oooh okey, fair enough

Nico-Sanchez: You forgot to change the variable name

agustinesco: Whoops, fixed

Nico-Sanchez (approved):

tvillegas98 (approved):
Seems to work well!

PR #770, opened 2024-07-12
Fix spikes throwing players outside map

agustinesco:
> [!warning]
> Merge alongside [This client pr](https://github.com/lambdaclass/champions_of_mirra/pull/1953)

## Motivation

When you stand in the spikes close to the map border and the spikes goes up you get thrown outside of the map, this is due to the `invalid_axis` method not calculating if an axis is valid correctly. This error is due to the line with polygon collision detection not working if the line is completely inside the polygon

## Summary of changes

- Fixed browser client to show the whole map 
- Fixed `invalid_axis` to correctly check if a player can be moved in an specific axis
- Fixed spikes vertices list not being sorted in clockwise

## How to test it?

Start a match and get inside the spikes when they're down, you should move to a valid position when they go up, this means no going outside of the map or inside another obstacle

## Checklist


AminArria (approved):
LGTM

apps/arena/native/physics/src/collision_detection/sat.rs
```diff
@@ -77,16 +67,24 @@ pub(crate) fn intersect_circle_polygon(
...
+            axis,
+            circle,
+            polygon,
+            &current_vertex,
+            &next_vertex,
+            obstacles,
+            external_wall,
+        ) {
+            continue;
+        }
```
Nico-Sanchez, line 80: Why is this moved? Is the order of the code important?

apps/arena/native/physics/src/collision_detection/sat.rs
```diff
@@ -306,19 +304,37 @@ fn invalid_axis(
...
+    let mut current_vertex_point_collisions =
+        current_vertex_point.collides_with(&external_wall_vector);
+
+    let next_vertex_moved = Position::add(next_vertex, &Position::mult(&axis, circle.radius * 2.0));
+    let mut next_vertex_point = Entity::new_point(polygon.id, next_vertex_moved);
+    let mut next_vertex_point_collisions = next_vertex_point.collides_with(&external_wall_vector);
+
+    if current_vertex_point_collisions.is_empty() && next_vertex_point_collisions.is_empty() {
+        return true;
+    }
```
Nico-Sanchez, line 323: Can you add comments explaining the reason behind all these operations?
It's easy to notice what are you doing code-wise (multiplying a position by player radius, checking a collision, if a list is empty, etc.) but it's pretty hard to know the reason behind each one.

apps/game_client/assets/js/hooks/board_game.js
```diff
@@ -52,6 +52,17 @@ export const BoardGame = function () {
 
     document.getElementById("board_container").appendChild(app.view);
 
+    let zoneCircle = new Graphics();
```
Nico-Sanchez, line 55: What's Graphics?

Nico-Sanchez (changes requested):

apps/arena/native/physics/src/collision_detection/sat.rs
agustinesco: This was something i tried in the first iteration of this fix, since the issue was something else these lines should be reverted, i'll leave the `Position::mult` change since we're doing that for everything in out code base

apps/arena/native/physics/src/collision_detection/sat.rs
agustinesco: Sure thing! i also corrected some typos and updated the whole method comment since it was outdated

apps/arena/priv/config.json
```diff
@@ -80,6 +80,10 @@
         "shape": "polygon",
         "type": "dynamic",
         "vertices": [
+          {
+            "x": 3350.0,
+            "y": 4300.0
+          },
```
Nico-Sanchez, line 86: Why is this moved here?

apps/arena/priv/config.json
```diff
@@ -112,6 +112,10 @@
               "polygon_hit": {
                 "damage": 10,
                 "vertices": [
+                  {
+                    "x": 3350.0,
+                    "y": 4300.0
+                  },
```
Nico-Sanchez, line 118: Same here.

apps/game_client/lib/game_client_web/live/pages/board/show.ex
```diff
@@ -14,10 +14,11 @@ defmodule GameClientWeb.BoardLive.Show do
...
       GameClient.ClientSocketHandler.start_link(self(), gateway_jwt, player_id, game_id)
 
-    mocked_board_width = 2000
-    mocked_board_height = 2000
-    backend_board_size = 10_000
+    mocked_board_width = 3000
+    mocked_board_height = 3000
+    backend_board_size = 15_000
     back_size_to_front_ratio = backend_board_size / mocked_board_width
+    map_radius = round(5520 / back_size_to_front_ratio)
```
Nico-Sanchez, line 21: Shouldn't this hardcoded 5520 be read from the JSON config instead?
What about the other values? Why those?

apps/game_client/lib/game_client_web/live/pages/board/show.ex
```diff
@@ -127,8 +129,8 @@ defmodule GameClientWeb.BoardLive.Show do
       category: entity.category,
       shape: entity.shape,
       name: entity.name,
-      x: entity.position.x / 5 + 1000,
-      y: entity.position.y / 5 + 1000,
+      x: entity.position.x / 5 + 1500,
+      y: entity.position.y / 5 + 1500,
```
Nico-Sanchez, line 133: Why 1500?

Nico-Sanchez (changes requested):
Nice work. Looks much better now.

Left some comments regarding some magic numbers in the client. I believe we can replace some of them with values in our config. Furthermore, we could make configurable the ones that aren't.

apps/game_client/lib/game_client_web/live/pages/board/show.ex
agustinesco: the other 3 values besides `map_radius` are "configurations" for the game client size, this 5520 can be taken from the config file since  it's the radius from the backend

apps/arena/priv/config.json
agustinesco: Yeah this change isn't needed at all, restored

apps/arena/priv/config.json
agustinesco: same

apps/game_client/lib/game_client_web/live/pages/board/show.ex
agustinesco: This 1500 is a scale from the mocked board size, since it went up from 10000 to 15000 we needed to change this too, I'll change this number for a division by 1000 since that's what we're doing basically

apps/game_client/lib/game_client_web/live/pages/board/show.ex
agustinesco: done!

Nico-Sanchez:
It's working nice. Good job.

However, now the algorithm in our client and here are different, resolving the collision differently.
Let's add a client PR before merging this.

https://github.com/user-attachments/assets/d91fa3f2-ddc8-42e1-b3f4-0212af8df7b2

Nico-Sanchez:
Check this out
https://github.com/lambdaclass/champions_of_mirra/pull/1953#issuecomment-2251155710

Nico-Sanchez (dismissed):

tvillegas98 (dismissed):
Great work you did @agustinesco !

apps/arena/native/physics/src/collision_detection/sat.rs
```diff
@@ -43,12 +43,13 @@ pub(crate) fn intersect_circle_polygon(
         let current_line = Position::sub(&current_vertex, &next_vertex);
         // the axis will be the perpendicular line drawn from the current line of the polygon
         axis = Position {
-            x: -current_line.y,
-            y: current_line.x,
+            x: current_line.y,
```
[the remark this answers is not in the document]
tvillegas98: Great! thanks for the explanation!

tvillegas98 (approved):
Great work!

Nico-Sanchez (approved):

PR #783, opened 2024-07-17
[GH-782] username changes

### LINKED ISSUE #782: [FEATURE] Remove username unique constraint and add name length validation

agustinesco:
## Motivation

The client needed some changes in the users.username behavior :
- usename should be able to be repeated 
- we need the same length validation that the client has

Closes #782

## Summary of changes

- Drop unique index game_id, username in the users table
- Add length validation to the user changeset

## How to test it?

- Create 2 users in the repo backend using:
```
{:ok, user1} = GameBackend.Users.create_guest_user
{:ok, user2} = GameBackend.Users.create_guest_user
```

- Then try to create an invalid name:
  - Short name:
    - `curl -X PUT -d username="a"  http://localhost:4001/users/user1.id`
  - Long name:
    - `curl -X PUT -d username="agustinesco"  http://localhost:4001/users/user1.id`
  - Repeated name
     - `curl -X PUT -d username="agustin"  http://localhost:4001/users/user1.id`
     - `curl -X PUT -d username="agustin"  http://localhost:4001/users/user2.id`

Testing champions users:
- Creating user
  - `{:ok, champions_user} = Champions.Users.register("something")`
- Try to change the name to anything
  - `curl -X PUT -d username="ag"  http://localhost:4001/users/champions_user.id`
It should change without any restriction

## Checklist


apps/game_backend/lib/game_backend/users/user.ex
```diff
@@ -64,10 +67,10 @@ defmodule GameBackend.Users.User do
       :profile_picture,
       :google_user_id
     ])
-    |> unique_constraint([:game_id, :username])
     |> cast_assoc(:unlocks)
     |> assoc_constraint(:google_user)
     |> validate_required([:game_id, :username])
+    |> validate_length(:username, min: @min_chars, max: @max_chars)
```
AminArria, line 73: More a philosophical question, but do we see any value of having `@min_chars` vs just putting `3`?  (When we are talking one single usage, multiple for sure makes more sense)

AminArria (approved):
LGTM, on a more side note we probably should rename the field (or have a different one) called `display_name`

apps/game_backend/lib/game_backend/users/user.ex
agustinesco: hmmm yeah, it doesn't add anything. I'll just delete them

AminArria (dismissed):

apps/game_backend/lib/game_backend/users/user.ex
```diff
@@ -64,10 +64,10 @@ defmodule GameBackend.Users.User do
       :profile_picture,
       :google_user_id
     ])
-    |> unique_constraint([:game_id, :username])
     |> cast_assoc(:unlocks)
     |> assoc_constraint(:google_user)
     |> validate_required([:game_id, :username])
+    |> validate_length(:username, min: 3, max: 9)
```
Nico-Sanchez, line 70: Minimum should be 2.
```suggestion
    |> validate_length(:username, min: 2, max: 9)
```

apps/gateway/test/champions_test.exs
```diff
@@ -54,11 +54,6 @@ defmodule Gateway.Test.Champions do
 
       assert user.username == username
 
-      # CreateUser with the same username fails
-      :ok = SocketTester.create_user(socket_tester, username)
-      fetch_last_message(socket_tester)
-      assert_receive %WebSocketResponse{response_type: {:error, %Error{reason: "username_taken"}}}
```
Nico-Sanchez, line 60: You could create the user and test the opposite instead of just deleting this.

apps/gateway/test/champions_test.exs
```diff
@@ -888,7 +883,7 @@ defmodule Gateway.Test.Champions do
     end
 
     test "leveling up the Dungeon Settlements increments the afk rewards", %{socket_tester: socket_tester} do
-      {:ok, user} = Users.register("DungeonSettlementAFKRewardsUser")
+      {:ok, user} = Users.register("TestUser")
```
Nico-Sanchez, line 886: Why are you changing this names? I believe this was on purpose to differentiate users by their test purpose.

Nico-Sanchez (changes requested):
Nice work! Just a few comments.

apps/gateway/test/champions_test.exs
agustinesco: The username has a max_length restriction of 9, all these names have a length longer than 9. For example `DungeonSettlementAFKRewardsUser` has a length of too long

apps/gateway/test/champions_test.exs
agustinesco: Done! i also discovered that the afk_gacha_team created the method `get_user_by_username`, i decided to remove that method since the username is not an id thus is not unique causing that method to explode. I checked both repos of afk_gacha_game and that method is not used so it should be sure to delete

apps/serialization/gateway.proto
```diff
@@ -3,7 +3,6 @@ syntax = "proto3";
   message WebSocketRequest {
     oneof request_type {
       GetUser get_user = 1;
-      GetUserByUsername get_user_by_username = 2;
```
tkz00, line 6: Why isn't the index of PB updated?

agustinesco: We should never change the indexing number of the .proto to keep consistency with previous/future versions of the file, that's the same reason why you always add a new field with the latest index instead of moving them.

apps/gateway/test/champions_test.exs
Nico-Sanchez: Is it ok to put this restriction for every user? Can't we do a custom validation just for "game_id=com_game_id"? This way other clients (games) can have different rules.
It's not mandatory, but we could abstract this rule beforehand.

apps/gateway/test/champions_test.exs
Nico-Sanchez: Yeah, nice catch. If Gacha game needs that kind of function, it should use `get_user_by_game_id_and_username`

agustinesco: yeah! i think if needed it'll be created when the times come

apps/gateway/test/champions_test.exs
agustinesco: Hmmm yeah i can see that, since we'll share user schema with different games each one will have a different name ruling, will do!

agustinesco: done!

Nico-Sanchez (approved):

AminArria (approved):

PR #789, opened 2024-07-19
[GH-394] Add mana mechanic for ultimates

### LINKED ISSUE #394: Implement mana mechanic for ultimate skills
https://www.notion.so/lambdaclass/Tasks-tracking-09e6d0d73cf440ffa64304a27199afe1?p=6d4d819e15b7423d975c24955d02b7f2&pm=s

AminArria:
## Motivation

Closes #394 

## Summary of changes

- Add fields to `characters` and `users` for mana

## How to test it?

## Checklist


apps/arena/lib/arena/entities.ex
```diff
@@ -36,6 +36,13 @@ defmodule Arena.Entities do
         stamina_interval: character.stamina_interval,
         cooldown_multiplier: 1,
         recharging_stamina: false,
+        max_mana: character.base_mana,
+        mana: 50,
```
agustinesco, line 40: I think this should be 0 instead of 50 right? is this the current amount of mana?

AminArria: Yes, this is the current mana and it starts at 50

apps/game_backend/lib/game_backend/units/skills/skill.ex
agustinesco, line 1: Can we add validation in the changeset here regarding the cooldown_mechanism? e.g. if you select the mana mechanism you need to add the mana_cost for example

apps/game_backend/lib/game_backend/units/characters/character.ex
agustinesco, line 1: Same as above, lets add a validation regarding the mana_recovery_strategy. I feel this change is very error prone

apps/arena/lib/arena/game/player.ex
```diff
@@ -124,14 +132,34 @@ defmodule Arena.Game.Player do
     put_in(player, [:aditional_info, :stamina_interval], stamina_interval)
   end
 
+  def recover_mana(player) do
+    now = System.monotonic_time(:millisecond)
+    time_since_last = now - player.aditional_info.mana_recovery_time_last_at
+
+    if time_since_last >= player.aditional_info.mana_recovery_time_interval_ms do
```
agustinesco, line 139: Won't this value be nil in case we choose a damage recovery strategy? but the game is not crashing. Can you explain this pls?

apps/arena/lib/arena/entities.ex
agustinesco: nice

agustinesco (changes requested):

apps/arena/lib/arena/game/player.ex
AminArria: Got lucky on this one (a.k.a not planned), I should make an explicit check on the type of mana recovery (time or damage based)

As to why it works, well its due to [term ordering](https://hexdocs.pm/elixir/Kernel.html#module-term-ordering), in essence any 2 terms in Elixir can be compared and there is a specific ordering if compering different terms
```
number < atom < reference < function < port < pid < tuple < map < list < bitstring
```

In this particular case we are comparing a number (`now`) and an atom (`nil`), so it will always return false

agustinesco (dismissed):
LGTM!

priv/repo/seeds.exs
agustinesco, line 1: maybe we want to restore this until the unity client implement a mana visualization

agustinesco (approved):

Nico-Sanchez (approved):

PR #795, opened 2024-07-23
Leaderboard feature missing information

agustinesco:
## Motivation

The client needed the prestige in the user show endpoint to be able to display it in the main screen
Also there were a lot of users without any unit, causing the leaderboard list to display a lot of users without any rank sorted in any order, replaced the left join with join to remove those users

## Summary of changes
- Add prestige virtual field to user schema
- Calculate prestige and send it in the show endpoint for users
- Change left join with join with units for users in leaderboard listing

### Leaderboard listing

Create an user:
iex: `{:ok, user} = GameBackend.Users.create_guest_user()`

List leaderboard:
terminal: `curl -X GET http://localhost:4001/curse/users/leaderboard`

Delete units:
iex: `GameBackend.Repo.delete_all(GameBackend.Units.Unit)`

List leaderboard:
terminal: `curl -X GET http://localhost:4001/curse/users/leaderboard`

It should appear empty

### User show

Create another user:
iex: `{:ok, user} = GameBackend.Users.create_guest_user()`

get user:
terminal: `curl -X GET http://localhost:4001/curse/users/user.id`

## Checklist


apps/game_backend/lib/game_backend/users.ex
```diff
@@ -77,7 +77,11 @@ defmodule GameBackend.Users do
     q =
       from(u in User,
         where: u.id == ^id and u.game_id == ^game_id,
-        preload: [units: [:character, :items], currencies: :currency]
+        left_join: unit in Unit,
```
AminArria, line 80: Why `left_join` and not `join`?

agustinesco: you're right! this was the first iteration of this query and then it got changed to `join` since there were some users without units. Changed!

AminArria (approved):

Nico-Sanchez (approved):

PR #797, opened 2024-07-24
[GH-796] Quests missing features

### LINKED ISSUE #796: [FEATURE] Add quest missing features and weekly quest
- We need to send the client the following information from the quests associated with the user when retrieving it from the backend
  - Description
  - Objective
  - Conditions
  - Progress
  - Goal
- We need to implement weekly quests , they will behave the same as daily quests but they will take matches that happend in the whole week to accumulate progress
- Implement and endpoint to create and claim quests, remove the quest completion when the match ends

agustinesco:
## Motivation

We need to rework how we communicate and complete daily quests, also add weekly quests:

- We need to send the state and information of the quest that a user has in an endpoint, we'll the user show endpoint for this purpose
- implement *weekly quests* they are the same as daily but you have a whole week to complete them instead of just one day
- Currently, we're completing the daily quest when a match is finished, this is wrong. The daily quests should be claimed through a button in the client. 
- We need an endpoint to be able to generate quest with a request

Closes #796

## Summary of changes
The main objective of this pr is to be able to fully use the quest feature without accessing the elixir terminal, and also implement the weekly quests feature

- Removed quest completion on match registration and added endpoint to complete/claim quests
- Added endpoint to generate user quests for a user
- Added quests to the user when retrieving him through the show endpoint
- Added weekly quest type

## How to test it?

- Enter the game as a guest in Unity and copy the user_id:
Captura de pantalla 2024-07-24 a la(s) 5 10 15 p  m

- Run get through the terminal to see that the user doesn't own any quest
`curl -X GET  http://localhost:4001/curse/users/user_id`
- Give daily quests to the user
`curl -X POST -d type=daily -d amount=3 http://localhost:4001/curse/users/user_id/quest/add_quests`
- Give weekly quests to the user
`curl -X POST -d type=weekly -d amount=3 http://localhost:4001/curse/users/user_id/quest/add_quests`
- Get to see if he has any quest
`curl -X GET  http://localhost:4001/curse/users/user_id`
- Play a match in unity and lose 
- if you run the get endpoint again you should see that the quest has some progress depending on your actions in the match, there's a quest that you need to lose a match to complete, copy that quest "id" field and run the following command:
`curl -X GET  http://localhost:4001/curse/users/user_id/quest/quest_id/complete_quest`
You should get a `quest completed` message the first time and an error message the second time. The same should happen to the rest of the unfinished quests

## Checklist


apps/game_backend/lib/game_backend/quests/quest.ex
```diff
@@ -19,6 +20,10 @@ defmodule GameBackend.Quests.Quest do
     timestamps()
   end
 
+  @types ["daily", "bounty", "weekly"]
+
+  def types(), do: @types
```
AminArria, line 25: Modify `field(:type)` to be `field(:type, Ecto.Enum,  values: [:daily, :bounty, :weekly])`

Then if you need the possible values you can do `Ecto.Enum.values(GameBackend.Quests.Quest, :type)`

Bonus, you can remove `validate_inclusion/3` below

apps/gateway/lib/gateway/controllers/curse_of_mirra/quest_controller.ex
```diff
@@ -44,4 +46,19 @@ defmodule Gateway.Controllers.CurseOfMirra.QuestController do
         |> send_resp(400, "Error completing bounty")
     end
   end
+
+  def add_quests(conn, %{"user_id" => user_id, "type" => type, "amount" => amount}) do
```
AminArria, line 50: Why do we have/need an endpoint to add quests? Who is to use this?

apps/gateway/lib/gateway/router.ex
```diff
@@ -34,8 +34,13 @@ defmodule Gateway.Router do
       put "/currency", CurrencyController, :modify_currency
       get "/claim_daily_reward", UserController, :claim_daily_reward
       get "/get_daily_reward_status", UserController, :get_daily_reward_status
-      get "/quest/:quest_id/reroll_daily_quest", QuestController, :reroll_daily_quest
-      get "/quest/:quest_id/complete_bounty", QuestController, :complete_bounty
+
+      scope "/quest" do
+        put "/add_quests", QuestController, :add_quests
```
AminArria, line 39: The semantics of this endpoints are kinda weird, a PUT under `/quests` you would expect it to be modifying a quest, but it is actually modifying a user

apps/gateway/lib/gateway/controllers/curse_of_mirra/quest_controller.ex
```diff
@@ -44,4 +46,19 @@ defmodule Gateway.Controllers.CurseOfMirra.QuestController do
...
+    with {:ok, _changes} <- Quests.add_quests_to_user_id_by_type(user_id, type, String.to_integer(amount)) do
+      conn
+      |> send_resp(200, "Quests added")
+    end
+  end
+
+  def complete_quest(conn, %{"user_id" => user_id, "quest_id" => quest_id}) do
```
AminArria, line 57: Why does the client need to ask us to complete a quest? Shouldn't we (backend) be doing this automatically?

AminArria (changes requested):

apps/game_backend/lib/game_backend/quests/quest.ex
agustinesco: Nice suggestion! done

apps/gateway/lib/gateway/controllers/curse_of_mirra/quest_controller.ex
agustinesco: Now the client will be the one completing question by taping a "claim" button, that's why we need this endpoint

apps/gateway/lib/gateway/controllers/curse_of_mirra/quest_controller.ex
agustinesco: It's still to be defined but we need to find a way to generate user quests when a user log in, so i though about generating quests through an endpoint to have something working. Maybe we could move it to the auth function or something like that but i think this could be developed when we finish this feature 
in https://github.com/lambdaclass/mirra_backend/issues/800

apps/gateway/lib/gateway/router.ex
agustinesco: Maybe replacing this with a post instead of a put may do the trick

AminArria (approved):

agustinesco:
This pr leaves the daily quest in an unfinished state and we want feature complete prs, the quest behavior will be fully finished in this pr https://github.com/lambdaclass/mirra_backend/pull/815

PR #799, opened 2024-07-24
Multiple matchmaking queues

AminArria:
## Motivation

Base to have multiple matchmaking queues

## Summary of changes

- Create `Matchmaking` module to give the module to use for matchmaking (in essence the queue). In a near future we can define the matchmaking behaviour in it to simplify the logic in the others 
- Create 3 different matchmaking queues
  - `BattleRoyalMode` for the regular matchmaking we currently have
  - `QuickGameMode` for matchmaking quick games, only bot matches for testing purposes
  - `PairMode` more an example model, but could be the one we use for 2v2v2... (if we decide on that mode). For now a copy&paste of `BattleRoyalMode`

Important to note that the changes are backwards compatible, so we don't need a PR on the client to be merged at the time. We will use a fallback to go through `BattleRoyalMode` or `QuickGameMode`. After a PR on client side we will be able to remove this fallbacks and the `quick_game_handler.ex` (finally! Only one matchmaking socket implementation)

## How to test it?

Try playing regular and quick games, nothing should change

Try playing some games in Unity and some in game_client, but this time in game_client chose `Play Pair` and notice that you end up in a different match than Unity players playing regular matchmaking

## Checklist

### DIFF
--- a/apps/arena/lib/arena/matchmaking.ex
+++ b/apps/arena/lib/arena/matchmaking.ex
@@ -3,7 +3,7 @@ defmodule Arena.Matchmaking do
   Module that handles matchmaking queues
   """
 
-  def get_queue("battle-royal"), do: Arena.Matchmaking.GameLauncher
+  def get_queue("battle-royale"), do: Arena.Matchmaking.GameLauncher
   def get_queue("pair"), do: Arena.Matchmaking.PairMode
   def get_queue("quick-game"), do: Arena.Matchmaking.QuickGameMode
   def get_queue(:undefined), do: Arena.Matchmaking.GameLauncher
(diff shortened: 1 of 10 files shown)


agustinesco (approved):
LGTM! tested with an android build and it works amazing on main! nice job

apps/arena/lib/arena/matchmaking.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Arena.Matchmaking do
+  @moduledoc """
+  Module that handles matchmaking queues
+  """
+
+  def get_queue("battle-royale"), do: Arena.Matchmaking.GameLauncher
+  def get_queue("pair"), do: Arena.Matchmaking.PairMode
+  def get_queue("quick-game"), do: Arena.Matchmaking.QuickGameMode
+  def get_queue(:undefined), do: Arena.Matchmaking.GameLauncher
```
tvillegas98, line 9: Instead of pattern matching against undefined, we can just ignore the parameter? just wondering, not mandatory

tvillegas98 (commented):
The game works well! Only have a question, Why do we need to wait for the client to remove the quick game handler?

PR #805, opened 2024-07-25
"Bigger Map"

marco-paulucci:
- Player count reduced  to 8 to give more space in the map
    - In the future the idea is to 6
    - Spawnpoints re-adjusted
    - For the time being, spawnpoints 9 and 10 are located in the center.
- Power-ups
    - Nerfed effect, since there will be more of them.
    - Doubled their amount in the map
    - Reduced crate HP by 20%
- Reduced player speed by 10% (through configurator)


tvillegas98:
Based on what I talked to @manucamejo I've lowered the players amount to 7. 
Also you'll had to change the initial positions in the configurator `localhost:4100/map_configurations`

PR #809, opened 2024-07-26
Add support for team mode in arena

AminArria:
## Motivation

In preparation for team modes we need to have support for them

## Summary of changes

- Add a `team` attribute to players and a corresponding `owner_team` to projectiles and pools
- When executing a mechanic, projectile damaging, or pool giving an effect filter out players from the same team
- Modify matchmaking to assign a team to each player (and use a proper map for params instead of tuples)

Making friendly fire a toggle setting is left as a follow up cause it would be a way bigger change to be able to pass the information to the necessary places at the moment

## How to test it?

**Regular arena (a.k.a battle royale)**
- Play as any other day and everything should work as usual

**Team arena**
To test this you have 3 options
- **Option 1:** Use game client and select different game modes
- **Option 2**: change the link in you clients, if not merged, in this PR https://github.com/lambdaclass/champions_of_mirra/pull/1987 you can go to `GameNavigationManager.cs:22` and modify `battle-royal` to `pair`, that will make `Play` button use the pair mode (2v2v2...)
- **Option 3:** modify `game_updater.ex:830` with `team: rem(last_id, 2)`, this will essentially make teams of pairs

## Checklist


apps/arena/lib/arena/entities.ex
```diff
@@ -6,21 +6,34 @@ defmodule Arena.Entities do
...
+          team: integer(),
+          player_name: String.t(),
+          position: %{x: float, y: float},
+          direction: %{x: float, y: float},
+          character_name: String.t(),
+          config: map,
+          now: integer()
+        }
+
+  @spec new_player(new_player_params()) :: map()
```
agustinesco, line 20: niiiice

agustinesco (dismissed):
LGTM!

agustinesco:
We should also create a pr in the client that adds the field in the message.proto file

agustinesco (changes requested):
Testing this pr i noticed that with the introduction of the bushes we may forgot to add this logic to the team logic, i think a player should always see the players in his team. Let's ask about this

agustinesco (approved):
Talked with the team and the bushes issues should be fixed in next iterations, this looks good! great job!

agustinesco (dismissed):
LGTM! since this pr is compatible with the current version of main in the client repo there's no need to update the version of the backend

apps/arena/lib/arena/matchmaking/quick_game_mode.ex
```diff
@@ -95,18 +99,20 @@ defmodule Arena.Matchmaking.QuickGameMode do
...
-        clients: clients,
-        bot_clients: bot_clients,
-        game_params: game_params
-      })
+    ## For Battle Royale (quick game mode) there are no teams so we assing each player to a different team
+    {teams, _} =
+      Enum.reduce(clients ++ bot_clients, {[], 1}, fn client, {team_acc, team_id} ->
+        client = Map.put(client, :team, team_id)
+        {[client | team_acc], team_id + 1}
+      end)
```
Nico-Sanchez, line 107: Can we put this team assignment in a separate function? We could create functions for this and we could pattern-match the game mode to assign the teams accordingly.
It will be easier to maintain and also this code block is repeated.

PS: there's a typo and its assign not assing.
```suggestion
    ## For Battle Royale (quick game mode) there are no teams so we assign each player to a different team
    {teams, _} =
      Enum.reduce(clients ++ bot_clients, {[], 1}, fn client, {team_acc, team_id} ->
        client = Map.put(client, :team, team_id)
        {[client | team_acc], team_id + 1}
      end)
```

Nico-Sanchez (changes requested):

apps/arena/lib/arena/matchmaking/quick_game_mode.ex
AminArria: Done, went a bit more and cleaned up `QuickGameMode` duplicated logic by calling directly `GameLauncher` function (essentially they are supposed to be almost the same, so didn't see any harm on it)

PR #792, opened 2024-07-22
Use own ping implementation

Nico-Sanchez:
> [!Warning]
> Merge this pr alongside [This client pr](https://github.com/lambdaclass/champions_of_mirra/pull/1975)

## Motivation

Ping was working really weird. It was using cowboy websocket's native implementation which is hard to customize and see what's really going on.
This is part of #794 

## Summary of changes

- [implement new ping and pong messages in game websocket](https://github.com/lambdaclass/mirra_backend/pull/792/commits/d5d201806ba97209634015b06aa61456da545042)
- [Delete unused code](https://github.com/lambdaclass/mirra_backend/pull/792/commits/317f3bd7f032457bc4f36e0f8394bc401a9e74fe)
- [Rename now variable to timestamp_now in proto](https://github.com/lambdaclass/mirra_backend/pull/792/commits/b4c7b523a0e5237c42a421d54c6c0a71bacff0a4)
- [Send ping every 500 ms instead of 30](https://github.com/lambdaclass/mirra_backend/pull/792/commits/b1446e2a1d1be7a7db76d98123372b0ae0b4d260)

## How to test it?

Play a game using client and this branches, you'll see the ping being updated every second precisely.

## Checklist
- [x] Tested the changes locally.
- [x] Reviewed the changes on GitHub, line by line.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #811, opened 2024-07-29
Fix ping message for game client

agustinesco:
## Motivation

After the merge of https://github.com/lambdaclass/mirra_backend/pull/792 the game client broke since we were expecting another kind of message in the :ping message, we'll simply ignore it since we don't use the ping info in the game client

## Summary of changes

Add ping and ping_update message to the noop_message list

## How to test it?

Go to [This link](http://localhost:3000/) and play a match

## Checklist

### DIFF
--- a/apps/game_client/lib/game_client_web/live/pages/board/show.ex
+++ b/apps/game_client/lib/game_client_web/live/pages/board/show.ex
@@ -102,11 +102,7 @@ defmodule GameClientWeb.BoardLive.Show do
     {:noreply, assign(socket, game_status: :finished, winner_id: finished_event.winner.id)}
   end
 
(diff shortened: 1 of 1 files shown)


tvillegas98 (approved):
Nice catch with this! 
Although it let us still playing with the client, don't you think it is worth to display the ping in browser's game client?

agustinesco:
> Although it let us still playing with the client, don't you think it is worth to display the ping in browser's game client?
> maybe in the future we can address it

Hmmm yeah i think it may be worth it but we're not deploying the game client to any server, and also you can't select a server to play in with the game client. Maybe we could include it when we develop the game client further

Nico-Sanchez (approved):

PR #815, opened 2024-07-29
[GH-800] Add active feature to quests

### LINKED ISSUE #796: [FEATURE] Add quest missing features and weekly quest

### LINKED ISSUE #800: [REFACTOR] Implement activated and deactivated quests
We need to change the scope in which the daily quests count matches to accumulate progress:

When we generate a users quest we need to create 7 quests:
- 3 activated quests that can be rerolled for free
- 3 deactivated quests. when you claim one of the previous quests one of these gets activated
- 1 meta quest that can be claimed when you complete 6 quests in that day

Each reroll costs 10 gold 

It's very important to start counting matches for progress when the quest activates, for example:
if I activate quest A at 14:00 pm and played a match at 11:00 am that match won't count towards quest completion of A

agustinesco:
>[!Warning]
> Merge with [gh-1971-add-quest-feature](https://github.com/lambdaclass/champions_of_mirra/pull/2003)
## Motivation

We need to rework how we communicate and complete daily quests, also add weekly quests:
- We need to send the state and information of the quest that a user has in an endpoint, we'll the user show endpoint for this purpose
- implement weekly quests they are the same as daily but you have a whole week to complete them instead of just one day
- Currently, we're completing the daily quest when a match is finished, this is wrong. The daily quests should be claimed through a button in the client.
- We need to generate quests when either:
  -  We create an user
  - A user logs in and it has passed a day since the quests were generated. We'll check if the day have changed, not counting time

The specs for daily quests now goes like:
- Now a user will begin the day with 6 quests:
  - 3 deactivated and 3 activated
- We'll only count progress for matches that happened on the same day of that quest creation and that happened AFTER the quest was activated
- When you claim a quest you'll activate one of the inactive quests and activate it

Closes #800 
Closes #796

## Summary of changes

The main objective of this pr is to be able to fully use the quest feature without accessing the elixir terminal, and also implement the weekly quests feature

- Removed quest completion on match registration and added endpoint to complete/claim quests
- Added quests to the user when retrieving him through the show endpoint
- Added weekly quest type
- Add `activated_at` for `user_quest`
- Add `last_daily_quest_generation` to users
- Add inactive quest activation on quest completion
- Add 3 new quests to meet the 6 quest minimum

## How to test it?

- Enter the game as a guest in Unity and copy the user_id:
Captura de pantalla 2024-07-24 a la(s) 5 10 15 p  m

- Get to see if he has any quest
`curl -X GET  http://localhost:4001/curse/users/user_id`

- Play a match in unity and lose 

- if you run the get endpoint again you should see that the quest has some progress depending on your actions in the match, there's a quest that you need to lose a match to complete, copy that quest "id" field and run the following command:
`curl -X GET  http://localhost:4001/curse/users/user_id/quest/quest_id/complete_quest`

- Get the user and check that the quest is finished and a new quest turn actives, this means that it has a value in `activated_at`
`curl -X GET  http://localhost:4001/curse/users/user_id`

## Checklist


apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -109,8 +109,13 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
   def get_user_quest(user_quest_id) do
     q = from(uq in UserQuest, preload: [:quest], where: uq.id == ^user_quest_id)
+    user_quest = Repo.one(q)
 
-    Repo.one(q)
+    if user_quest do
+      {:ok, user_quest}
+    else
+      {:error, :not_found}
+    end
```
Nico-Sanchez, line 118: Use is_nil and maybe a case statement. Also the user_quest variable isn't needed.

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -157,52 +162,45 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+
+        Repo.transaction(multi)
+    end
   end
 
   def reroll_daily_quest(daily_quest_id) do
     reroll_configurations = Application.get_env(:game_backend, :quest_reroll_config)
 
-    daily_quest =
+    {:ok, daily_quest} =
```
Nico-Sanchez, line 203: This will break if :error is returned, are we ok with that?

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -298,6 +296,79 @@ defmodule GameBackend.CurseOfMirra.Quests do
     |> Repo.transaction()
   end
 
+  def get_user_quest_progress(%UserQuest{quest: %Quest{} = quest} = user_quest, arena_match_results) do
+    arena_match_results =
+      case quest.type do
+        type when type in ["daily", :daily] ->
+          Enum.filter(arena_match_results, fn arena_match_result ->
```
Nico-Sanchez, line 303:
```suggestion
     if quest.type in ["daily", :daily] do
        Enum.filter(arena_match_results, fn arena_match_result ->
```

seems more simple. Also, why do we need to handle both string and atom cases?

And why just the dailies? What about weekly or bounty ones?

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -298,6 +296,79 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+        nil ->
+          acc
+
+        progress ->
+          acc + progress
+      end
+    end)
+  end
+
+  def try_to_complete_quest_for_user(user, user_quest) do
```
Nico-Sanchez, line 329:
```suggestion
  def complete_user_quest(user, user_quest) do
```

is a better name, there's no try. If it fails it will return error.

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
```diff
@@ -298,6 +296,79 @@ defmodule GameBackend.CurseOfMirra.Quests do
...
+    |> Multi.run(:maybe_activate_quest, fn _, _ ->
+      Enum.find(user.user_quests, fn user_quest -> is_nil(user_quest.activated_at) end)
+      |> case do
+        nil ->
+          {:ok, :no_quests_left}
+
+        user_quest ->
+          user_quest
+          |> UserQuest.changeset(%{activated_at: NaiveDateTime.utc_now()})
+          |> GameBackend.Repo.update()
```
Nico-Sanchez, line 355: Do we want to do this everytime? Shouldn't we isolate this behavior for weekly ones only?
Are bounty and daily ones under this "activate" behavior too?
What happens if I complete a bounty. Will that activate a weekly?

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -74,60 +70,104 @@ defmodule GameBackend.Users do
...
         group_by: u.id,
-        select: %{u | prestige: sum(unit.prestige)}
+        select: %{u | prestige: sum(unit.prestige), quest_refresh_at: ^quest_refresh_at}
       )
+      |> quests_preloads()
+      |> arena_match_results_preloads()
+
+    user =
+      Repo.one(q)
+      |> add_quest_progress_and_goal()
```
Nico-Sanchez, line 93: This will get time consuming. Is this a decision made to send the whole user information (quests, etc) at once in a single endpoint that calls this query?

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -74,60 +70,104 @@ defmodule GameBackend.Users do
...
 
-    if user = Repo.one(q), do: {:ok, user}, else: {:error, :not_found}
+    if user, do: {:ok, user}, else: {:error, :not_found}
   end
 
   @doc """
   Get a list of GoogleUser based on their id with the necessary preloads
```
Nico-Sanchez, line 99:
```suggestion
  Get a list of User based on their id with the necessary preloads
```

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -533,4 +573,132 @@ defmodule GameBackend.Users do
...
+      CurseUsers.create_user_params()
+      |> register_user()
+    end)
+    |> Multi.run(:generate_quests, fn
+      _, %{insert_user: user} ->
+        quests_results = generate_daily_quests_for_user(user)
+
+        any_failed_quest? = Enum.find(quests_results, fn {result, _quest} -> result == :error end)
+
+        case any_failed_quest? do
```
Nico-Sanchez, line 606: I think a failing quest is more like when you don't complete a quest. Here you want to check if a DB operation failed.
```suggestion
        error? = Enum.find(quests_results, fn {result, _quest} -> result == :error end)

        case error? do
```

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -533,4 +573,132 @@ defmodule GameBackend.Users do
...
+            {:ok, :quests_generated}
+        end
+    end)
+    |> Multi.run(:user, fn _, %{insert_user: user} ->
+      get_user_by_id_and_game_id(user.id, curse_id)
+    end)
+    |> Repo.transaction()
+  end
+
+  def get_user_and_maybe_insert_daily_quests(user_id) do
```
Nico-Sanchez, line 620: Either call the function "get_curse_user..." or receive the game_id as parameter.

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -533,4 +573,132 @@ defmodule GameBackend.Users do
...
+      _, %{get_user: user} ->
+        today = Date.utc_today()
+
+        should_generate_quests? =
+          user.last_daily_quest_generation_at &&
+            Date.compare(NaiveDateTime.to_date(user.last_daily_quest_generation_at), today) == :lt
+
+        if should_generate_quests? do
+          quests_results = generate_daily_quests_for_user(user)
+          any_failed_quest = Enum.find(quests_results, fn {result, _quest} -> result == :error end)
```
Nico-Sanchez, line 637: Same here regarding the failed quest naming.

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -533,4 +573,132 @@ defmodule GameBackend.Users do
...
+            quest_id: quest_params.id,
+            status: "available"
+          }
+
+          changeset = UserQuest.changeset(%UserQuest{}, attrs)
+
+          Repo.insert(changeset)
+      end)
+
+    active_quests ++ inactive_quests
```
Nico-Sanchez, line 702: Is there any reason to return every transaction result even if some of their insertions failed?
If we only want them all to succeed we should implement a multi then return either {:ok, list_of_quests} or {:error, _changesets}

apps/game_backend/lib/game_backend/users/user.ex
```diff
@@ -47,6 +47,10 @@ defmodule GameBackend.Users.User do
     has_many(:currency_caps, UserCurrencyCap)
 
     timestamps()
+
+    # Virtual fields for client display
+    field(:prestige, :integer, virtual: true)
+    field(:quest_refresh_at, :any, virtual: true)
```
Nico-Sanchez, line 53: Why are this virtual? Don't we want to store them?

apps/game_backend/lib/game_backend/quests/user_quest.ex
```diff
@@ -9,12 +9,18 @@ defmodule GameBackend.Quests.UserQuest do
...
   schema "user_quests" do
     field(:completed_at, :utc_datetime)
     field(:status, :string)
+    field(:activated_at, :utc_datetime)
+
     belongs_to(:quest, Quest)
     belongs_to(:user, User)
 
+    field(:progress, :integer, virtual: true)
+    field(:goal, :integer, virtual: true)
```
Nico-Sanchez, line 22: Why are these virtual?

apps/gateway/lib/gateway/controllers/auth_controller.ex
```diff
@@ -36,7 +36,7 @@ defmodule Gateway.Controllers.AuthController do
     with {:ok, claims} <- TokenManager.verify(gateway_jwt),
          hashed_client_id = :crypto.hash(:sha256, client_id),
          {:ok, ^hashed_client_id} <- Base.url_decode64(claims["dev"]),
-         {:ok, user} <- Users.get_user(claims["sub"]) do
+         {:ok, %{user: user}} <- Users.get_user_and_maybe_insert_daily_quests(claims["sub"]) do
```
Nico-Sanchez, line 39: Can't we do this separately? I think it would add more clarity and wouldn't replace former get_user behavior.
Just another with clause to "maybe_insert_daily_quests".

apps/gateway/lib/gateway/controllers/curse_of_mirra/user_controller.ex
```diff
@@ -38,7 +38,7 @@ defmodule Gateway.Controllers.CurseOfMirra.UserController do
   end
 
   def create_guest_user(conn, %{"client_id" => client_id}) do
-    with {:ok, user} <- Users.create_guest_user() do
+    with {:ok, %{user: user}} <- Users.insert_user_and_insert_daily_quests() do
```
Nico-Sanchez, line 41: Same here. Can't we get the user then insert the daily quests?

Nico-Sanchez (changes requested):

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: I used a case statement with a nil case

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: 1. I don't actually remember why we handle both cases, will see if it breaks under any circumstance
2. Because two reasons
   a. Users don't get to complete bounties manually, that is completed inside a match and we have a proper check in arena where we don't need to filter the match result since it only checks the one in the match
   b. This method will be used only inside the `get_user_by_id_and_game_id` method that has the `arena_match_results_preloads` which only preloads matches that happened in that week so for "weeklies" you don't need that filter. It was the easiest solution i found to not perform a query for each user quest that we want to check

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: completly agreed

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: 1. This is for the daily quests
2. Since we don't have a specification on how the weekly quests will behave I'll just add a type filter so we don't mess up the different quests

apps/game_backend/lib/game_backend/users.ex
agustinesco: yup this will send the quest information in the same endpoint. It's debatable to add it in the same endpoint or do a new specific endpoint for the quests. What do you think?

apps/game_backend/lib/game_backend/users/user.ex
agustinesco: Actually, you don't want to store this field. This is just to send the refresh date (a simple utc.now) to the client so all stay consistent. Maybe for some reason the server and the client have a different hour and that will cause inconsistency, the field that you wanna persist is `last_daily_quest_generation_at`

apps/game_backend/lib/game_backend/quests/user_quest.ex
agustinesco: You don't want to save these fields in the db, you always want to calculate the progress of the quest.

apps/game_backend/lib/game_backend/users.ex
agustinesco: I'll go with the "curse" approach

apps/game_backend/lib/game_backend/users.ex
agustinesco: I'll change the failed with an error

apps/game_backend/lib/game_backend/users.ex
agustinesco: agreed!

apps/gateway/lib/gateway/controllers/auth_controller.ex
agustinesco: Sure thing, i'll split the quest generation and user retrieval in this module

apps/gateway/lib/gateway/controllers/curse_of_mirra/user_controller.ex
agustinesco: In this case i don't think that a good idea, you want to either generate a user with everything that the game needs to work or to not generate anything. Since we'll be creating buggy users if we split the user creation from everything else

apps/game_backend/lib/game_backend/users.ex
agustinesco: removed since i changes the quest generation output

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: hmmmm i'd rather not let this break here, i'll move the daily quest retrieval to the controller just as anything else

apps/game_backend/lib/game_backend/users.ex
Nico-Sanchez: Discussed in a call. We will leave it as it is until this is a priority.

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
Nico-Sanchez: 2. Alright.
What about 1? Do we need both checks? Can we fix that somehow to use string or atom but not both?
Finally, did you see my "use if not case" suggestion? I believe is cleaner.

Nico-Sanchez (changes requested):

apps/game_backend/lib/game_backend/curse_of_mirra/quests.ex
agustinesco: 1. Yeah the string case wasn't necessary at all, since this function is used in two places
   a. to get the quest progress and send it to the client to display
   b. to check if the user has completed the quest to be claimed, and both cases uses atoms 

I think this was a trace from a previews implementation so i just removed it

Nico-Sanchez (dismissed):

Nico-Sanchez:
Tested with client branch in unity and retrieved the progress using `curl` in the terminal ✅

apps/game_backend/lib/game_backend/users.ex
```diff
@@ -127,48 +136,80 @@ defmodule GameBackend.Users do
...
 
   ## Examples
 
-      iex> get_users_with_todays_daily_quests(["51646f3a-d9e9-4ce6-8341-c90b8cad3bdf"])
-      [%GoogleUser{}]
+      iex> get_users_with_quests_and_results(["51646f3a-d9e9-4ce6-8341-c90b8cad3bdf"])
+      [%User{}]
   """
-  def get_users_with_todays_daily_quests(ids, repo \\ Repo) do
+  def get_users_with_quests_and_results(ids, repo \\ Repo) do
```
tvillegas98, line 152:
```suggestion
  def list_users_with_quests_and_results(ids, repo \\ Repo) do
```

agustinesco: Done!

Nico-Sanchez (approved):

tvillegas98 (approved):

PR #817, opened 2024-07-30
[GH-816] Add polymorph item

### LINKED ISSUE #816: [FEATURE] add polymorph item
Create a new item that applies a new effect. 
This effect will be removed when either:
- 9 seconds passed
- The player performed an action

This item will only apply an effect without any action to the gameplay, since this will be a visual item that change the player visuals with a bug

agustinesco:
## Motivation

We need an item that only applies an effect in the backend, with no gameplay change whatsoever, this is for the game client to apply a visual change in the character models

Closes #816

## Summary of changes

- Added new effect to config.json
- Added new effect to `consumable_item` form
- Added debug mode to game_client, this is a new button on top of the game client map

## How to test it?

For easy testing you can comment every other `consumable_item` in the `seeds.ex` file and start a fresh db, the go to the game client and pick up an use an item, to do so use the "L" key

## Checklist

### DIFF
--- a/apps/arena/priv/config.json
+++ b/apps/arena/priv/config.json
@@ -506,12 +506,7 @@
       "duration_ms": 9000,
       "remove_on_action": true,
       "one_time_application": true,
-      "effect_mechanics": {
-        "polymorph": {
-          "execute_multiple_times": true,
-          "effect_delay_ms": 0
-        }
-      }
+      "effect_mechanics": {}
     }
   ],
   "crates": [
--- a/apps/arena/priv/config.json
+++ b/apps/arena/priv/config.json
@@ -500,6 +500,18 @@
           "execute_multiple_times": true
         }
       }
+    },
+    {
+      "name": "polymorph_effect",
+      "duration_ms": 9000,
+      "remove_on_action": true,
+      "one_time_application": true,
+      "effect_mechanics": {
+        "polymorph": {
+          "execute_multiple_times": true,
+          "effect_delay_ms": 0
+        }
+      }
     }
   ],
   "crates": [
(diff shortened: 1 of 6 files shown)


apps/arena/priv/config.json
```diff
@@ -500,6 +500,18 @@
...
         }
       }
+    },
+    {
+      "name": "polymorph_effect",
+      "duration_ms": 9000,
+      "remove_on_action": true,
+      "one_time_application": true,
+      "effect_mechanics": {
+        "polymorph": {
```
AminArria, line 510: Why do you need this? Wouldn't an empty list be the same?

agustinesco: We don't need it you're right, i was replicating what the invisible effect have but this isn't necessary, deleted

priv/repo/seeds.exs
```diff
@@ -689,6 +689,17 @@ giant_fruit_params = %{
 {:ok, giant_fruit} =
   GameBackend.Items.create_consumable_item(giant_fruit_params)
 
+polymorph_params = %{
+  active: true,
```
tvillegas98, line 693: We have to set this to false, otherwise the game will crash since the client needs an image to display the items in the UI 

Screenshot 2024-08-05 at 17 59 21

apps/game_client/assets/js/hooks/board_game.js
```diff
@@ -263,5 +287,14 @@ export const BoardGame = function () {
         }
       }
       entity.boardObject.tint = color;
+    }),
+    (this.updateEntityText = function (entity, backEntity) {
```
tvillegas98, line 291: Thanks for this!

tvillegas98 (changes requested):
Thanks for the new additions to the browser client!! Left a comment regarding the new item

priv/repo/seeds.exs
agustinesco: Whoops, done!

tvillegas98 (approved):

AminArria (approved):

PR #824, opened 2024-07-31
[GH-594] Add otix

### LINKED ISSUE #594: [FEATURE] Implement Otix
https://www.notion.so/lambdaclass/Otix-7ec9118c585c46aa96a42a04b4c66e5f

- [ ] Basic
- [ ] Ulti
- [ ] Dash

agustinesco:
## Motivation

Add otix into the game, the skills are specified in this notion https://www.notion.so/lambdaclass/Otix-7ec9118c585c46aa96a42a04b4c66e5f

basic: A circular hit with range
dash: A regular dash like the other characters
Ultimate: A pool without any range that deal damage and slow down enemies

Closes #594 

## Summary of changes

- Add `position_hit` mechanic
- This pr also fixes an pre-existent error with the pool effects removal where if you stand in a pool when it expires the effect is not cleaned

## How to test it?

Go to the game client and choose otix as player, play as regular

You can also use this branch from the client to play in unity `add-otix`

## Checklist


apps/arena/native/physics/src/lib.rs
```diff
@@ -232,6 +248,30 @@ fn direction_from_positions(position_a: Position, position_b: Position) -> Direc
     }
 }
 
+#[rustler::nif()]
+fn get_relative_positions_from_entity_with_angles(
```
tvillegas98, line 252: Is this function being used?

agustinesco: Nope, this was from a previous implementation for the ultimate, i just removed it

tvillegas98 (dismissed):

apps/arena/lib/arena/game_updater.ex
```diff
@@ -1848,6 +1867,16 @@ defmodule Arena.GameUpdater do
...
 
+  defp remove_pool_effects_from_entities(game_state, pool, entities) do
+    Enum.reduce(entities, game_state, fn {entity_id, _entity}, acc ->
+      if entity_id in pool.collides_with and pool.aditional_info.status == :READY do
+        Effect.remove_owner_effects(acc, entity_id, pool.id)
+      else
+        acc
+      end
+    end)
+  end
```
Nico-Sanchez, line 1878: If you're changing the game state only for a pool that is :READY, we should use pattern-match to only modify the game state in that case and just return the state otherwise.
Here, we are traversing the entities map just to return the same game state.

apps/arena/lib/arena/game_updater.ex
```diff
@@ -1848,6 +1867,16 @@ defmodule Arena.GameUpdater do
     game_state
   end
 
+  defp remove_pool_effects_from_entities(game_state, pool, entities) do
+    Enum.reduce(entities, game_state, fn {entity_id, _entity}, acc ->
+      if entity_id in pool.collides_with and pool.aditional_info.status == :READY do
```
Nico-Sanchez, line 1872: We're always checking if entity_id is in pool.collides_with. I think it's cheaper to just take the collided entities, apply the modifications, then merge (update) the entities map.

apps/arena/lib/arena/game/skill.ex
```diff
@@ -258,6 +180,51 @@ defmodule Arena.Game.Skill do
...
+      y: entity.position.y + direction.y
+    }
+
+    direction =
+      Utils.normalize(%{
+        x: target_position.x - entity.position.x,
+        y: target_position.y - entity.position.y
+      })
+
+    duration_ms = Physics.calculate_duration(entity.position, target_position, simple_shoot.speed, simple_shoot.range)
```
Nico-Sanchez, line 205: What are the cases where the target_position is outside the range? Shouldn't we handle or restrict that before doing the calculation in the physics module? I think we're overfitting our physics to handle backend's unhandled edge cases, but maybe I'm wrong.

apps/arena/native/physics/src/lib.rs
```diff
@@ -135,9 +135,25 @@ fn get_direction_from_positions(position_a: Position, position_b: Position) -> D
...
+        direction.normalize();
+
+        let new_target = Position::add(&position_a, &Position::mult(&direction, max_range));
+
+        let len = distance_between_positions(position_a, new_target);
+
+        (len / speed) as u64
+    } else {
+        (len / speed) as u64
+    }
```
Nico-Sanchez, line 156: This is what I mentioned in Skill.ex file.

apps/game_client/lib/game_client_web/controllers/page_html/home.html.heex
```diff
@@ -1,6 +1,6 @@
 <%= if assigns[:current_user_id] do %>
   <.form :let={f} for={%{}} action={~p"/"}>
-    <.input field={f[:character]} name="character" label="Select a Character" type="select" options={["muflus", "h4ck", "uma", "valtimer", "kenzu"]} value=""/>
+    <.input field={f[:character]} name="character" label="Select a Character" type="select" options={["muflus", "h4ck", "uma", "valtimer", "kenzu", "otix"]} value=""/>
```
Nico-Sanchez, line 3: Please add otix in loadtest character list too 😄

Nico-Sanchez (changes requested):

apps/arena/lib/arena/game/skill.ex
agustinesco: Now projectiles can pick a destination, maybe the client cheats and the destination has more range than the possible one provided by the mechanic.
Also this happens when the auto-aim have more range than the projectile range

apps/arena/native/physics/src/lib.rs
agustinesco: Answered above

apps/game_client/lib/game_client_web/controllers/page_html/home.html.heex
agustinesco: Whoops, done

apps/arena/lib/arena/game_updater.ex
agustinesco: This was more complicated than it should, i reworked it

apps/arena/lib/arena/game_updater.ex
agustinesco: Done with the refactor

apps/arena/native/physics/src/lib.rs
Nico-Sanchez: I'm not sure if this is the best place, we could handle it in Elixir before but we can address this later.

Nico-Sanchez (approved):

Nico-Sanchez (approved):
It's working fine. We should playtest this before merging though.

apps/arena/priv/config.json
```diff
@@ -179,6 +179,24 @@
...
+      "one_time_application": true,
+      "allow_multiple_effects": false,
+      "effect_mechanics": {
+        "damage": {
+          "damage": 13,
+          "effect_delay_ms": 400,
+          "execute_multiple_times": true
+        },
+        "speed_boost": {
+          "modifier": -0.60,
```
tvillegas98, line 195: I think we can lower this later, sometimes between other effects and using the ultimate I got stuck and I couldn't move. Causing the movement to feel weird

tvillegas98 (dismissed):

Nico-Sanchez (approved):

tvillegas98 (approved):

PR #847, opened 2024-08-08
Fix over defense bug

agustinesco:
## Motivation

When you play as kenzu and used both your ultimate and a gaint fruit you defense turned into 1.25 wich causes the damage to heal which broke the damage logic

## Summary of changes

Capped defense 

## How to test it?

Play as kenzu and use your ultimate and a giant fruit and then receive damage. That shouldn't break

## Checklist


tvillegas98 (approved):

tkz00 (approved):
Greattt! Solves the issue!

PR #850, opened 2024-08-12
[GH-851] Add square map to seeds

### LINKED ISSUE #851: [FEATURE] Add merliot map config
Create a map configuration for the map in this client branch `merliot-map`

agustinesco:
## Motivation

We are creating a new map and we need to create the seeds for it, also since we're picking a map at random we need a way to deactivate this new map since it stills in development

closes #851

## Summary of changes

- Added merliot map to seeds
- Added `active` field to map configs
- Filter `active` maps when creating a game

## How to test it?

Run the seeds and in the game configurator edit the `Merliot` map configuration `active` field to true, if you play in the game client you should sometimes encounter a square map

Test with this client branch and see if it works too: `merliot-map`

## Checklist

### DIFF
--- a/apps/arena/lib/arena/configuration.ex
+++ b/apps/arena/lib/arena/configuration.ex
@@ -29,7 +29,11 @@ defmodule Arena.Configuration do
 
     Jason.decode!(payload.body, [{:keys, :atoms}])
     |> Map.update!(:map, fn maps ->
-      map = Enum.random(maps)
+      map =
+        maps
+        |> Enum.filter(fn map -> map.active end)
+        |> Enum.random()
+
       parse_map_config(map)
     end)
     |> Map.update!(:characters, fn characters ->
--- a/apps/configurator/lib/configurator_web/controllers/map_configuration_html/map_configuration_form.html.heex
+++ b/apps/configurator/lib/configurator_web/controllers/map_configuration_html/map_configuration_form.html.heex
@@ -2,6 +2,7 @@
   <.error :if={@changeset.action}>
     Oops, something went wrong! Please check the errors below.
   </.error>
+  <.input field={f[:active]} type="checkbox" label="Active" />
   <.input
     field={f[:version_id]}
     type="select"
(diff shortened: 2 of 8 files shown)


apps/arena/lib/arena/configuration.ex
```diff
@@ -29,7 +29,11 @@ defmodule Arena.Configuration do
 
     Jason.decode!(payload.body, [{:keys, :atoms}])
     |> Map.update!(:map, fn maps ->
-      map = Enum.random(maps)
+      map =
```
tvillegas98, line 32: This seems ok for now, It would be better to send the maps already filtered but I'm not sure what are we going to do with the maps in the client, so, I think this is ok

apps/configurator/lib/configurator_web/controllers/map_configuration_html/map_configuration_form.html.heex
```diff
@@ -9,6 +9,7 @@
     label="Version"
   /> <.input field={f[:name]} type="text" label="Name" />
   <.input field={f[:radius]} type="number" label="Radius" step="any" />
+  <.input field={f[:active]} type="select" options={[true, false]} label="Active" />
```
tvillegas98, line 12: This could be changed for a checkbox and we can move it at the top of the form

tvillegas98 (changes requested):
You did a nice approach to the square map! left a comment and a change for the map's form

apps/configurator/lib/configurator_web/controllers/map_configuration_html/map_configuration_form.html.heex
agustinesco: Sounds good!

tvillegas98 (approved):
Nice work!

apps/game_backend/priv/repo/migrations/20240812135550_add_active_to_map_config.exs
```diff
@@ -0,0 +1,9 @@
+defmodule GameBackend.Repo.Migrations.AddActiveToMapConfig do
+  use Ecto.Migration
+
+  def change do
+    alter table(:map_configurations) do
+      add :active, :boolean, default: :false
+    end
```
AminArria, line 7: Merging this migration as it is and with the filtering for `active == true` will cause the game to break until there is at least one map configuration with `active == true`

AminArria (changes requested):

AminArria (approved):

Issue #852, opened 2024-08-12
[REFACTOR] Migrate missing config.json fields to configurator app


agustinesco:
- [x] Power ups -> https://github.com/lambdaclass/mirra_backend/pull/853
- [x] Effects -> https://github.com/lambdaclass/mirra_backend/pull/856
- [x] Crates -> https://github.com/lambdaclass/mirra_backend/pull/854
- [x] Traps -> https://github.com/lambdaclass/mirra_backend/pull/860

manucamejo:
done

PR #854, opened 2024-08-12
[GH-852] Configurator: Migrate crates config

agustinesco:
## Motivation

Keep migrating field from config.json to configurator app

Part of https://github.com/lambdaclass/mirra_backend/issues/852

## Summary of changes

- Added crates list to map_configuration schema
- Added edit and update for these fields in configurator app
- Used map config instead of fields from json in game updater to spawn crates

## How to test it?

Start a match and everything should keep working as before

## Checklist


apps/arena/lib/arena/configuration.ex
```diff
@@ -184,6 +185,15 @@ defmodule Arena.Configuration do
     }
   end
 
+  defp parse_crate(crate) do
```
tvillegas98, line 188: It seems parse_pool, parse_bush, and parse crate do the same thing, we can have just one function for them 🤔 . What do you think?

tvillegas98 (changes requested):
Nice work overall! a left a comment about an opportunity to refactor

apps/arena/lib/arena/configuration.ex
agustinesco: Sounds good! 🪨

tvillegas98 (dismissed):
Nice work!

AminArria (approved):

AminArria (approved):

AminArria (approved):

tvillegas98 (approved):

PR #864, opened 2024-08-15
Client server autodetection

agustinesco:
> [!Warning]
> This pr can be merged without the client one

## Motivation

We want to list the servers in the backend list in the game client, we need to filter them and change the encoding of the endpoint response
This feature will also include the refactor of the way the client generate the servers buttons. Now it'll retrieve the list of every active server from the db and list it dinamically instead of hardcode them. We'll list every server even if they are for testing enviroment so we can choose them at will

## How to test it?

- Play in [this client branch](https://github.com/lambdaclass/champions_of_mirra/pull/2035) and everything should work

## Checklist


apps/game_backend/lib/game_backend/configuration.ex
```diff
@@ -213,6 +213,24 @@ defmodule GameBackend.Configuration do
...
+      [%ArenaServer{status: "active"}, ...]
+
+  """
+  def list_active_arena_servers do
+    q =
+      from(as in ArenaServer,
+        where: as.status == ^"active"
+      )
+
+    Repo.all(q)
```
Nico-Sanchez, line 231: Do we use this for unity production app? Shouldn't we filter by `environment: "production"`?

agustinesco: We don't want that behavior, at least not for the time being. We'll use the `enviroment` field for the ping autodetection only but we want to list the rest of the servers

Nico-Sanchez (approved):

AminArria (approved):

PR #868, opened 2024-08-16
State diffs

AminArria:
## Motivation

In preparation for sending state diffs rather than entire state

## Summary of changes

- Add function`diff/2` to calculate diff of 2 terms
- Commented out code to send diff state
- Adapted broadcast code to allow for missing fields

## How to test it?

Code is commented out so nothing will change, but feel free to uncomment the call to `diff/2` and run some games to make sure it does not crash

## Checklist


AminArria:
Included in #888

PR #869, opened 2024-08-16
New map layout v0.2

tkz00:
- Changes crates positions to fit better with the new map layout.
- Adds new changes to the seeds, even though this is now managed by the configurator, when we use the backend locally, it still uses the seeds, so this change is needed for local development.

To test this play with the `main` branch from the client pointing to this backend branch locally and check that the bushes, lakes, rocks and spikes colliders work as expected.


apps/configurator/priv/config.json
agustinesco, line 1: the initial positions are already in the configurator si this isn't doing anything 😬

agustinesco (approved):
LGTM!

apps/configurator/priv/config.json
tvillegas98: Also, we can remove this file, it is no longer being used

agustinesco (approved):

tvillegas98 (approved):

PR #876, opened 2024-08-21
Change game ticking to start when it changes to RUNNING state

AminArria:
## Motivation

This is required for sending state diffs, otherwise when players are joining they already are behind/don't know the original state the diff is referring to

## Summary of changes

- Change when we start calling `update_game`

## How to test it?

Play games and nothing should have changed

## Checklist


AminArria:
This breaks something in the clients (or how the flow is managed) and results in clients able to move as soon as it loads rather than after the countdown

AminArria:
Not going to happen for now

PR #886, opened 2024-08-28
Fixes for players visibility on bushes 

agustinesco:
## Motivation

We need to always show players with an item effect active or if they are executing an skill

## Summary of changes

Add item effect duration to players entity to identify items duration

## How to test it?

Using another client use an item and go inside a bush, you should be able to see that client as ling as it has the item effect

## Checklist

### DIFF
--- a/apps/arena/lib/arena/entities.ex
+++ b/apps/arena/lib/arena/entities.ex
@@ -65,7 +65,8 @@ defmodule Arena.Entities do
         bounties: [],
         selected_bounty: nil,
         bounty_completed: false,
-        current_basic_animation: 0
+        current_basic_animation: 0,
+        item_effect_duration: now
       },
       collides_with: []
     }
--- a/apps/arena/lib/arena/entities.ex
+++ b/apps/arena/lib/arena/entities.ex
@@ -66,7 +66,7 @@ defmodule Arena.Entities do
         selected_bounty: nil,
         bounty_completed: false,
         current_basic_animation: 0,
-        item_effect_duration: now
+        item_effects_expires_at: now
       },
       collides_with: []
     }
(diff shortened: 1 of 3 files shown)


apps/arena/lib/arena/game/player.ex
```diff
@@ -575,4 +587,23 @@ defmodule Arena.Game.Player do
...
+    duration =
+      System.monotonic_time(:millisecond) + duration_ms
+
+    game_state
+    |> update_in([:players, player.id, :aditional_info, :item_effect_duration], fn item_duration ->
+      if item_duration < duration do
+        duration
+      else
+        item_duration
+      end
```
Nico-Sanchez, line 602:
```suggestion
      max(item_duration, duration)
```

apps/arena/lib/arena/game/player.ex
```diff
@@ -415,6 +417,16 @@ defmodule Arena.Game.Player do
     |> Effect.apply_stat_effects()
   end
 
+  def player_executing_skill(player) do
```
Nico-Sanchez, line 420:
```suggestion
  def player_executing_skill?(player) do
```

apps/arena/lib/arena/entities.ex
```diff
@@ -65,7 +65,8 @@ defmodule Arena.Entities do
         bounties: [],
         selected_bounty: nil,
         bounty_completed: false,
-        current_basic_animation: 0
+        current_basic_animation: 0,
+        item_effect_duration: now
```
Nico-Sanchez, line 69: Duration: now? What does this represent? Feels wrong to set a timestamp to a "duration".
Maybe rename this to `item_expires_at` if that's what it represents.

Nico-Sanchez (changes requested):

apps/arena/lib/arena/game/player.ex
agustinesco: done!

apps/arena/lib/arena/game/player.ex
agustinesco: done

apps/arena/lib/arena/entities.ex
agustinesco: it represent the time where the last item effect expires, i replaced it with `item_effect_expires_at`

Nico-Sanchez (approved):

manucamejo (approved):

Issue #494, opened 2024-04-18
Send diffs instead of the whole state in game updates 


Nico-Sanchez:
Currently we do modifications of the game state every tick then broadcasting the whole game state.
I think we could broadcast just the modifications without breaking our updates pipe (see code at the end):
- We can do it like a "fold", piping not only the game state but also a modifications map.
- As a first iteration we can start doing the modifications in the modifications map and in the game state, and conserve the broadcast of the whole state.
- As long as we're developing the modifications in each function, we could start changing the broadcast message.
- At the end of these iterations, we'll only broadcast the modifications map.

Updates pipe:
```elixir
game_state =
      game_state
      |> Map.put(:ticks_to_move, ticks_to_move)
      |> reduce_players_cooldowns(time_diff)
      |> move_players()
      |> update_projectiles_status()
      |> move_projectiles()
      |> resolve_players_collisions_with_power_ups()
      |> resolve_players_collisions_with_items()
      |> resolve_projectiles_collisions_with_players()
      |> apply_zone_damage_to_players(state.game_config.game)
      |> explode_projectiles()
      |> handle_pools(state.game_config)
      |> Skill.apply_effect_mechanic()
      |> Map.put(:server_timestamp, now)
```

PR #888, opened 2024-08-28
State deltas for game updates of players and projectiles

AminArria:
> [!warning] 
> Needs to be merged with https://github.com/lambdaclass/champions_of_mirra/pull/2084

## Motivation

Part of #494

## Summary of changes

- Update protobuf
  - Mark fields that we are going to delta as `optional`, this allows to check if there are set or not. Otherwise default value for the type is implied (e.g. 0 for integers)
  - Modify `vertices` to support deltas
  - Add `UNDEFINED` enums to use as `null` value
- Created `diff/2` to receive any 2 terms and output the resulting delta, check the code for a more thorough explanation and edge cases
- Modify creation of update message to handle all cases where something could be `nil` due to the result of diffing
- Send deltas for `obstacles`, `bushes`, and `crates`. This 3 things represented more than 80% of the delta reduction that we could expect to achieve
- `game_client` **is broken due to this change**. Fixing it should be a follow up to this

## How to test it?

Test with this client branch https://github.com/lambdaclass/champions_of_mirra/pull/2084

To test play as usual

We should be doing some load tests and profiling in apps to compare against not doing deltas

## Checklist


apps/serialization/messages.proto
```diff
@@ -171,17 +175,17 @@ enum GameStatus {
...
-  string name = 4;
+  optional uint64 id = 1;
+  optional string category = 2;
+  optional string shape = 3;
+  optional string name = 4;
   Position position = 5;
-  float radius = 6;
-  repeated Position vertices = 7;
+  optional float radius = 6;
+  optional ListPositionPB vertices = 7;
```
AminArria, line 184: This is an interesting change that is needed for diffing lists under a specific scenario, otherwise its a near impossible task. So the scenarios
- If `new_list` is different than `old_list` (meaning not an exact match) the resulting delta is `new_list` 
- If `new_list` is exactly the same as `old_list` the resulting delta is `nil` (no delta)

To allow for that second scenario we would have needed a sort of `optional repeated` field in protobuf, but this is not supported. So the second best thing is this solution where we wrap the list in another message. This wrapper becomes `optional` and the inner only have the `repeated` field

apps/serialization/messages.proto
```diff
@@ -223,27 +227,28 @@ message Effect {
...
   ProjectileStatus status = 3;
-  string skill_key = 4;
+  optional string skill_key = 4;
 }
 
 enum ProjectileStatus {
-  ACTIVE = 0;
-  EXPLODED = 1;
-  CONSUMED = 2;
+  PROJECTILE_STATUS_UNDEFINED = 0;
```
AminArria, line 241: This new enum value `PROJECTILE_STATUS_UNDEFINED` and the similar one in `CrateStatus` is a [protobuf best practice](https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum) cause it been the 0 value will serve as the default for the enum. You are not supposed to actually have business logic tied to it, for example I only use it in the client as a null check

Additionally you probably noticed that its called `PROJECTILE_STATUS_UNDEFINED` rather than `UNDEFINED` that's cause all enum names share a namespace and the protobuf C# compiler didn't like it

Nico-Sanchez (approved):

agustinesco (approved):
LGTM!

PR #889, opened 2024-08-29
Add randomness to the spikes behaviour

manucamejo:
## Motivation

All the spikes were moving together, we needed to add some randomness.

## Summary of changes

The next status depends on the value set on the ConfiguratorApp and a random value also depending on that value.

## How to test it?

See if the spikes are spawning and hiding separately.

## Checklist

### DIFF
--- a/apps/arena/lib/arena/game/obstacle.ex
+++ b/apps/arena/lib/arena/game/obstacle.ex
@@ -85,7 +85,7 @@ defmodule Arena.Game.Obstacle do
     random_ms =
       Enum.random(
         Integer.floor_div(-time_until_transition_ms, 2)..Integer.floor_div(
-          next_status_params.time_until_transition_ms,
+          time_until_transition_ms,
           2
         )
       )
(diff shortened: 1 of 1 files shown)


agustinesco (dismissed):
LGTM!

apps/arena/lib/arena/game/obstacle.ex
```diff
@@ -80,14 +80,24 @@ defmodule Arena.Game.Obstacle do
 
     now = DateTime.utc_now() |> DateTime.to_unix(:millisecond)
 
+    time_until_transition_ms = next_status_params.time_until_transition_ms
+
+    random_ms =
+      Enum.random(
+        Integer.floor_div(-time_until_transition_ms, 2)..Integer.floor_div(
+          next_status_params.time_until_transition_ms,
```
Nico-Sanchez, line 88: Isn't this just...
```suggestion
          time_until_transition_ms,
```

Nico-Sanchez (changes requested):
Doing this we lose the constant time between spike transitions, which we learn through a game to make decisions.
Also, the 'randomness range list' could be just a coefficient in the configurator (randomness coefficient).

We can merge it as it is, but I want to know if we took all this into account and to create the respective issues if what I said brings something new to the table.

manucamejo:
Nico you are right, let's merge it and create the issue please.

agustinesco (approved):

PR #908, opened 2024-09-09
Configurator versioning

Nico-Sanchez:
## Motivation

*Explain why not how.*
Closes #issue_number

## Summary of changes

*List your changes here.*

## How to test it?

## Checklist


apps/game_backend/lib/game_backend/configuration.ex
```diff
@@ -541,4 +543,50 @@ defmodule GameBackend.Configuration do
...
     Repo.one(from(g in GameConfiguration, order_by: [desc: g.inserted_at], limit: 1))
   end
+
+  @doc """
+  Returns a 16 base encoded string representing the hashed value of the map configurations.
+  ## Examples
+      iex> get_configuration_hash_version()
+      "A6F0CC6917D195AB8A03129ACBE1FA48364845B8"
+  """
+  def get_configuration_hash_version do
```
AminArria, line 553: Why not `:erlang.term_to_binary/1` the map configuration (or whatever subfields we actually want)?

apps/game_backend/lib/game_backend/configuration.ex
```diff
@@ -541,4 +543,50 @@ defmodule GameBackend.Configuration do
...
+    get_current_version()
+    |> Map.get(:map_configurations)
+    |> Enum.flat_map(fn map_config ->
+      Map.take(map_config, [:obstacles, :initial_positions, :bushes, :radius]) |> Map.values()
+    end)
+    |> List.flatten()
+    |> Enum.map(fn config -> sum_shape_coordinates(config) end)
+    |> Enum.reduce(fn coordinates_last, coordinates_current -> Decimal.add(coordinates_last, coordinates_current) end)
+    |> Decimal.to_string()
+    |> (fn coordinates -> :crypto.hash(:sha, coordinates) |> Base.encode16() end).()
```
AminArria, line 563: Why base16?

apps/game_backend/lib/game_backend/configuration.ex
Nico-Sanchez: Because the query order of everything changes the hash everytime, and we only want to change the hash when the desired values change.
We can discuss a better approach in Slack.

apps/game_backend/lib/game_backend/configuration.ex
Nico-Sanchez: Just selected one at random. Any suggestions?

AminArria: Nope, was just curious. Also, don't think is worth discussing for this use case

tvillegas98 (dismissed):

manucamejo (dismissed):

manucamejo:
This is not working properly because this PR is trying to access to a module in another app.
It works locally, but it doesn't in the servers.

apps/arena/lib/arena_web/controllers/health_controller.ex
```diff
@@ -8,8 +8,11 @@ defmodule ArenaWeb.HealthController do
   end
 
   def version(conn, _params) do
+    arena_version = Application.spec(:arena, :vsn) |> to_string()
+    configurator_version = GameBackend.Configuration.get_configuration_hash_version()
```
manucamejo, line 12: This is not correct

manucamejo (changes requested):
This is not working properly because this PR is trying to access to a module in another app.
It works locally, but it doesn't in the servers.

apps/arena/lib/arena_web/controllers/health_controller.ex
Nico-Sanchez: Be careful, this controller is duplicated in both arena and gateway apps.
We should distinct them somehow or see if both are needed.

PR #948, opened 2024-09-11
Check if killer is alive before giving powerup

### LINKED ISSUE #943: [BUG] Match registry is not working as expected
trophies are not being updated when you finish a match, this is due to the match registry failing to create an `arena_match_result` 

We should check what's causing the match result insert to fail and check if that fix the trophies

AminArria:
## Motivation

Closes #943 

## Summary of changes

Check that the player killer is alive itself before giving a power up. Not performing the check was essentially reviving players cause they get healed by the power up and this revival messed up the match position counter (resulted in negative positions) and thus match results could not be saved

## How to test it?

## Checklist


Nico-Sanchez (approved):

agustinesco (approved):
lets goooooooooooooo

PR #957, opened 2024-09-13
Add 404 error to user not found

agustinesco:
## Motivation

We needed to identify when a log in request failed due to an unexistant user

## Summary of changes

Added 404 error lo user log in 

## How to test it?

try to log in with an inexistant user id and you should see a 404 error

## Checklist

### DIFF
--- a/apps/gateway/lib/gateway/controllers/fallback_controller.ex
+++ b/apps/gateway/lib/gateway/controllers/fallback_controller.ex
@@ -64,6 +64,10 @@ defmodule Gateway.Controllers.FallbackController do
     send_resp(conn, 400, Jason.encode!(%{"error" => "the user doesn't have that quest"}))
   end
 
+  def call(conn, {:error, _failed_operation, :not_found, _changes_so_far}) do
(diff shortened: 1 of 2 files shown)


apps/gateway/lib/gateway/controllers/fallback_controller.ex
```diff
@@ -64,6 +64,10 @@ defmodule Gateway.Controllers.FallbackController do
     send_resp(conn, 400, Jason.encode!(%{"error" => "the user doesn't have that quest"}))
   end
 
+  def call(conn, {:error, _failed_operation, :not_found, _changes_so_far}) do
+    send_resp(conn, 404, Jason.encode!(%{"error" => "not found"}))
+  end
```
Nico-Sanchez, line 69: This is ok for the specific case when user is not found, but the with clause has many other operations. Are all of them properly handled?
They were being caught by a wild card in the with-else statement.

agustinesco: fair point, i added a default case

manucamejo (approved):

Nico-Sanchez (approved):

PR #1061, opened 2025-01-29
Link game modes table with matchmaking

tvillegas98:
## Motivation

In #1032 we have added a new table to store parameters that configures game modes and settings for the matchmaking, in this PR we're linking that table to our current matchmakings

## Summary of changes

- [feat: changes on the game mode configurations table and views](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/91efe2e586502d59db9c421abf365579cdbd3d0a)
- [feat: create an endpoint to retrieve the game mode configuration](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/f4edba485989389c5a8efdade13c312124a5450d)
- [chore: move every migration into one file](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/cbe5adc0805f17bd2438908b41268cd18a94ecc4)
- [chore: add seeds for game modes](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/331baec94bb569eda91f14f5514cd89656684ae4)
- [feat: add new fields and logic for game mode form](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/a5e5633a1338d3c0271e8c51d9cf149f693b70c2)
- [code: handle the failed requests when the application starts](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/d4bc7b4610c1de90deb806e034000f6b08a1ae4c)
- [feat: fix pair and battle not using the game mode configuration](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/9bf6df07f705ef82ee5524bfe7dffdd14b4f0d8f)
- [feat: changes on the initial positions and map](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/e84883aea8d39a2e025434b8897c4ed962b3e3f8)
- [feat: use the team size parameter](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/2779636f726298ad564ee59ffa2800788a80c9b6)

## How to test it?

I’ve added a name and a type for each game mode. This will allow us to create different game modes with unique configurations and, in the future, render them on the front end.

Additionally, if you’d like, you can adjust the parameters in the configurator under the Game Modes section.

I’ve also introduced some new parameters:

- Team Size: Specifies the size of each team.

You most be able to tweak those parameters and also these as well:
- match_duration_ms
- team_size
- respawn_time_ms

And you'll have to wait until the changes impact into the game (the params are refreshed every 5 seconds)

## Checklist


apps/arena/lib/arena/matchmaking/pair_mode.ex
```diff
@@ -53,23 +53,54 @@ defmodule Arena.Matchmaking.PairMode do
...
   def handle_info(:start_game, state) do
-    {game_clients, remaining_clients} = Enum.split(state.clients, Application.get_env(:arena, :players_needed_in_match))
-    create_game_for_clients(game_clients)
+    {game_clients, remaining_clients} = Enum.split(state.clients, state.game_mode_configuration.amount_of_players)
+    create_game_for_clients(game_clients, state.game_mode_configuration)
 
     {:noreply, %{state | clients: remaining_clients}}
   end
 
+  def handle_info(:update_params, state) do
```
tvillegas98, line 90: This is needed to update the params every certain amount of time. That way we will avoid to query the backend every millisecond

Nico-Sanchez (changes requested):
As we discussed at the office. The field :amount_of_players in map_mode_params table was intended and discussed to be that way.
Let's re-arrange this PR to add the feature without that change.
This will lead to different matchmakings per map and mode (many to many), and that's what we should aim for.

tvillegas98:
> As we discussed at the office. The field :amount_of_players in map_mode_params table was intended and discussed to be that way.
Let's re-arrange this PR to add the feature without that change.
This will lead to different matchmakings per map and mode (many to many), and that's what we should aim for.

Done in 
- [feat: restore the amount of players to the map mode params](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/cade3eead9b18af480a031b07c3fd02089487978)
- [feat: use the map amount params in the matchmakings](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/21e28f8e020c7821713f7ddd9d511b6d79786333)
- [feat: use the map selected from the matchmaking](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/d79aecda7a11bd1425c33d4d4f9f96ebcf12aa8a)
- [chore: restore changes in game modes form](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/3d6ff8e9f5817aa7611d8af42b2af2588e892dcb)

apps/arena/lib/arena/configuration.ex
```diff
@@ -7,14 +7,41 @@ defmodule Arena.Configuration do
...
 
-    get_current_game_configuration()
+    get_current_game_configuration(map_selected)
     |> Map.put(:client_config, client_config)
   end
 
-  defp get_current_game_configuration do
+  def get_game_mode_configuration(name, type) do
+    gateway_url = Application.get_env(:arena, :gateway_url)
+    query_params = URI.encode_query(%{"name" => name, "type" => type})
```
Nico-Sanchez, line 19: This kind of query is very error-prone. Type should be enough.
If we need something to differentiate between teams and solo, maybe we could create another field in the game mode schema as an enum for team mode types (:solo, :duo, :trio, etc).

apps/arena/lib/arena/matchmaking/deathmatch_mode.ex
```diff
@@ -60,19 +58,43 @@ defmodule Arena.Matchmaking.DeathmatchMode do
...
+      if Map.has_key?(state, :game_mode_configuration) do
+        state
+      else
+        case Arena.Configuration.get_game_mode_configuration("deathmatch", "deathmatch") do
+          {:error, _} ->
+            state
+
+          {:ok, game_mode_configuration} ->
+            # This is needed because we might not want to send a request every 300 seconds to the game backend
+            Process.send_after(self(), :update_params, 5000)
```
Nico-Sanchez, line 71: This function is called every 300 milliseconds.
If we fall here every 300 ms, we are queueing a message for 5 seconds from here, but every 300 ms, so we're just delaying the problem.
If we want to just update the params every 5 seconds, we should take this out of this function.

apps/arena/lib/arena/matchmaking/deathmatch_mode.ex
```diff
@@ -60,19 +58,43 @@ defmodule Arena.Matchmaking.DeathmatchMode do
...
   def handle_info(:start_game, state) do
-    {game_clients, remaining_clients} = Enum.split(state.clients, Application.get_env(:arena, :players_needed_in_match))
-    create_game_for_clients(game_clients)
+    {game_clients, remaining_clients} =
+      Enum.split(state.clients, state.current_map.amount_of_players)
+
+    create_game_for_clients(game_clients, state.game_mode_configuration, state.current_map)
 
-    {:noreply, %{state | clients: remaining_clients}}
+    map = Enum.random(state.game_mode_configuration.map_mode_params)
```
Nico-Sanchez, line 95: Why are we doing this? Couldn't we fall in an error where we use a different current_map to create_game_for_clients and then override a different one in the state.current_map?

apps/arena/lib/arena/matchmaking/pair_mode.ex
```diff
@@ -53,21 +53,58 @@ defmodule Arena.Matchmaking.PairMode do
...
+          {:ok, game_mode_configuration} ->
+            # This is needed because we might not want to send a request every 300 seconds to the game backend
+            map = Enum.random(game_mode_configuration.map_mode_params)
+
+            Process.send_after(self(), :update_params, 5000)
+
+            Map.put(state, :game_mode_configuration, game_mode_configuration)
+            |> Map.put(:current_map, map)
+        end
+      end
```
Nico-Sanchez, line 74: We could create a function for this block of code that we use in every matchmaking.

Nico-Sanchez (changes requested):
The PR works good in its scope.
However, I think we are being redundant or error-prone in the game mode table columns and its queries.
Let's refactor those to avoid redundancy and potential errors by defining the specific and minimum params to retrieve whatever mode and map (we're randomizing the maps for now) the client asks for.

apps/arena/lib/arena/configuration.ex
```diff
@@ -23,10 +50,11 @@ defmodule Arena.Configuration do
 
     Jason.decode!(payload.body, [{:keys, :atoms}])
     |> Map.update!(:map, fn maps ->
+      random_map = Enum.random(maps)
+
       map =
         maps
-        |> Enum.filter(fn map -> map.active end)
-        |> Enum.random()
+        |> Enum.find(random_map, fn map -> map.name == map_selected end)
```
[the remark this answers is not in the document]
tvillegas98: Random map is a default value for `Enum.find` https://hexdocs.pm/elixir/1.12/Enum.html#find/3
Weird definition to be honest

apps/arena/lib/arena/matchmaking/deathmatch_mode.ex
tvillegas98: done in [chore: send the update params event in the init of each matchmaking](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/b3df822b0785c0bb230428488125c6915a79964d)

apps/arena/lib/arena/matchmaking/pair_mode.ex
tvillegas98: done in [refactor: modularize function in matchmaking module](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/39f35cae42fb70e3b8acb37627e6c15a9a97ae1c)

tvillegas98:
> team_enabled is misleading. It tells that you can play as teams, but it doesn't tell this restriction.
If we want a boolean that makes the game mode for teams only, it should be "is_team?" or something like that.
> Furthermore and more important, we have a separate field for team and not team positions, but with this boolean we only use one of them, leaving the other one blank.
We should either use one of those positions checking if the client asked to join :team, :solo OR just use one position field and check the new boolean. But not both behavior because is redundant.


As we talked, we will:
- Remove the `:team_enabled` field
- Merge `team_initial_positions` and `solo_initial_positions` into one field called `initial_positions`
- Query the game modes by type and team size

apps/arena/lib/arena/matchmaking.ex
```diff
@@ -7,4 +7,22 @@ defmodule Arena.Matchmaking do
...
+  def get_matchmaking_configuration(matchmaking_state, team_size, game_mode_type) do
+    if Map.has_key?(matchmaking_state, :game_mode_configuration) do
+      matchmaking_state
+    else
+      case Arena.Configuration.get_game_mode_configuration(team_size, game_mode_type) do
+        {:error, _} ->
+          matchmaking_state
+
+        {:ok, game_mode_configuration} ->
+          # This is needed because we might not want to send a request every 300 seconds to the game backend
```
Nico-Sanchez, line 20: You forgot to delete this comment right?

tvillegas98: yes!

Nico-Sanchez (changes requested):
Just one more thing and we're ready to go.
Can we only select "active" maps in the configurator for the game modes?
Otherwise, we could link an inactive map to them which could lead to an error.
Doing so we can merge this and not link Araban by mistake until it's ready to play.

tvillegas98:
>Just one more thing and we're ready to go.
Can we only select "active" maps in the configurator for the game modes?
Otherwise, we could link an inactive map to them which could lead to an error.
Doing so we can merge this and not link Araban by mistake until it's ready to play.

Makes a lot of sense, done in [feat: only display active maps for map mode params](https://github.com/lambdaclass/mirra_backend/pull/1061/commits/6c078d41fd101cdcbf29c97b797ebf27e9ebc56d)

Nico-Sanchez (dismissed):

manucamejo (changes requested):
I was testing the branch, but I couldn't see Araban, are you filtering the INACTIVE maps?
image

tvillegas98:
> I was testing the branch, but I couldn't see Araban, are you filtering the INACTIVE maps? image

whoops. Fixed in 
feat: filter maps only in the game mode form

manucamejo (approved):

Nico-Sanchez (approved):

PR #1136, opened 2025-03-25
1131 merge bots inside arena (remove ws communication)

### LINKED ISSUE #1131: [REFACTOR] Merge bot behavior inside Arena's app
## 🚧 Refactor

we want to avoid the websockets' handling and be able to see internal game state from bots.

Nico-Sanchez:
## Motivation

We want to reduce the workload of enconding and decoding messages for bots.
Also, we want to make things easier for bot features, so they have to be able to interact with the backend internal state from there.
Closes #1131 

## Summary of changes

Besides the merge of the bots behavior inside arena, I did a few more things here:
- Don't allow player actions (move/attack) if they're dead.
- Move Collision Grid creation and management to a separate GenServer, allowing bots to fetch the grid on demand.
- Update Game Client characters list.

Here's a breakdown of the most important commit changes:
- [Add new Bot Supervisor and GenServers](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/e556f0d35bdff5fe96e98e52c160d074ce6f88c0)
- [Use new Bots GenServer instances in every game mode](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/70f8a1c5af272bbd9db13df812f1b312b6ef2d22)
- [Start BotSupervisor with Arena app and make bots work in games](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/1aadef49c1b91c8b72fbb3b060e94cc3a4ea3e47)
- [Update character list in game client](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/83a982de01038627493553eab236ddc71564edcd)
- [Complete entities and also add new pathfinder behavior](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/2c538303acab7ae77a1b951184000a6833e67844)
- [Use PubSub for bots updates instead of sending messages individually](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/95a5986ac18cd5316a2bd859025763b5abf77e38)
- [Delete everything related to old bots behavior in BotManager app](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/703482955ff505b940adf0f38f76183862eedcd3)
- [Add new GenServer to manage bot collision grids](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/88b66408dfdb26ae7f6abaa50e7cd6d4e1ac4158)
- [Refactor game state to complete entities once then encode when needed…](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/9c02117d685ec700a9eff2d01f3ad3a315730c4a)
- [Fix and refactor toggle_bots message in game updater](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/47e0a1174cd41c8300b6079803ed92fc80b34436)
- [Update bot state constant params once instead of every tick](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/dbbdccf4b85356be50c148dd9309d84cae86e09e)
- [Allow move or attack actions when player (or bot) is alive only](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/6b5dfd76338a90a69e803e99488f7a39fe0594b6)
- [Minor fixes to PathfindingGrid GenServer](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/a22097e914c78033faf151fdb876b1148408e5d5)
- [Start pathfinding grid genserver only when env-var is enabled](https://github.com/lambdaclass/mirra_backend/pull/1136/commits/3dd8f671c6c5f261bcba2ed7e55c6e3b2e805418)

## How to test it?

Everything should work as usual, but you won't see the bots websockets spawning in the logs.
Backend with PATHFINDING_TEST should work too.

## Checklist


apps/arena/lib/arena/bots/pathfinding_grid.ex
```diff
@@ -0,0 +1,101 @@
...
+        collision_grid =
+          case AStarNative.build_collision_grid(map_obstacles) do
+            {:ok, collision_grid} ->
+              collision_grid
+
+            {:error, reason} ->
+              Logger.error("Grid construction failed with reason: #{inspect(reason)}")
+
+              %{}
+          end
```
mFragaBA, line 69: can we return nil on the error case? I don't see too much sense in having to handle an empty map besides the binary and nil cases

apps/arena/lib/arena/bots/bot.ex
```diff
@@ -0,0 +1,146 @@
...
+  use GenServer
+  alias Phoenix.PubSub
+  alias Arena.Bots.PathfindingGrid
+  alias BotManager.BotStateMachine
+  alias BotManager.BotStateMachineChecker
+  require Logger
+  @action_delay_ms 30
+
+  def start_link(%{bot_id: bot_id, game_id: _game_id} = params) do
+    GenServer.start_link(__MODULE__, params, name: generate_bot_name(bot_id))
```
mFragaBA, line 15: Why are we using `Registry`? I noticed we're only using it here, but we never reference the bot by its name

apps/arena/lib/arena/bots/pathfinding_grid.ex
```diff
@@ -0,0 +1,101 @@
...
+    {:noreply, state}
+  end
+
+  def handle_call({:get_map_collision_grid, map_name}, _from, state) do
+    grid =
+      Enum.find(state.maps, fn map -> map.name == map_name end)
+      |> Map.get(:grid)
+
+    {:reply, grid, state}
+  end
```
mFragaBA, line 94: do we need the 2 ways of handling `get_map_collision_grid`? The function seems rather simple since it doesn't generate the collision grid itself so I would say we can just keep the `handle_call`, no?

mFragaBA: on a second thought, I think we might need the `handle_cast` because `update_config` can stall the message queue (so `get_map_collision_grid` request takes at least that amount of time to be replied) for a bit and we don't want to wait for that if we're the bot. Is that why? In that case we can just remove the `handle_call`

mFragaBA (changes requested):
Awesome work! I played locally and didn't see nothing too weird with the bot behavior (with and without pathfinding).

I left some minor observations but overall awesome work. I like how you split up the changes so we don't do absolutely everything in a single PR. There were a lot of changes but they were rather simple to understand because of that.

apps/arena/lib/arena/bots/pathfinding_grid.ex
Nico-Sanchez: It's just I forgot to remove the call, I did it that way first. Done.

apps/arena/lib/arena/bots/bot.ex
Nico-Sanchez: Since each bot is a different GenServer instance, I just wanted to name them uniquely: https://hexdocs.pm/elixir/1.12/GenServer.html#module-name-registration

apps/arena/lib/arena/game_updater.ex
```diff
@@ -206,12 +206,8 @@ defmodule Arena.GameUpdater do
...
 
   def handle_cast(:toggle_bots, state) do
-    encoded_msg =
-      GameEvent.encode(%GameEvent{
-        event: {:toggle_bots, %ToggleBots{}}
-      })
-
-    PubSub.broadcast(Arena.PubSub, state.game_state.game_id, {:toggle_bots, encoded_msg})
+    state = Map.put(state, :bots_enabled?, not state.bots_enabled?)
+    PubSub.broadcast(Arena.PubSub, "BOTS_#{state.game_state.game_id}", {:enable_bots, state.bots_enabled?})
```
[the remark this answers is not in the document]
Nico-Sanchez: Done.

mFragaBA (changes requested):
I played a game with kenzu and everything went well. Then I tried to play a game with valtimer and got:

image

I couldn't replicate later though.

It also happened to me that the bots stopped mid game:

image

Apparently they shut down, but no errors appeared whatsoever. I was able to consistently replicate this by:

- launching a game, waiting for a while and exiting
- right away starting a new game
- wait and at some point the bots will shutdown

I think the game end message is being broadcasted to all bots instead of just the ones of the game

https://github.com/user-attachments/assets/9d0d99ab-37ca-48a7-a72b-3943e632a961

Nico-Sanchez:
@mFragaBA you were right. I missed the `restart: :transient` option in the genservers. By default is :permanent. I'll explain this:
- :permanent: The supervisor restarts the children, even when stoping them ({:stop, reason, state} as we do).
- :transient. The supervisor restarts the children ONLY when they crash. Thus they won't be restarted when they're being stopped (match ended).

If you want to debug this and check yourself, I just made the matches end quickly (edited zone params to shrink faster and do more damage) and printed in terminal the children length of the BotSupervisor: `DynamicSupervisor.which_children(Arena.Bots.BotSupervisor) |> length`

Also, if you want to check their restart option or params, you can always do `:sys.get_state(Arena.Bots.BotSupervisor)`. That's how I noticed that I missed that option.

Thanks a lot for your review, it was nicely spotted, and this may fix a huge error in prod and loadtests.

PS:
- about :restart values -> https://hexdocs.pm/elixir/Supervisor.html#module-restart-values-restart
- about exit reasons -> https://hexdocs.pm/elixir/Supervisor.html#module-exit-reasons-and-restarts

mFragaBA (approved):
LGTM! Tested locally a few matches and I was not able to replicate the previous issue. Awesome work @Nico-Sanchez

PR #1147, opened 2025-04-03
Update docu

ManuelAlonso1:
## Motivation

Updated the Mirra backend setup documentation. Added detailed steps for installing Nix, Devenv, Protobuf, and dependencies for Elixir and JavaScript Protobuf, along with instructions to properly start the services. >

### DIFF
--- a/apps/configurator/README.md
+++ b/apps/configurator/README.md
@@ -35,37 +35,6 @@ You can:
 - Assign skills (see skills config) to each character: basic, ultimate, dash.
 - Tune character params: speed, health, stamina, etc.
 
-Its mandatory to activate a character to have a basic skin assigned, for example, if you want to add Uren you should add this this lines of code:
-
-- Set `active` to `true` in `seeds.exs`
-charactername_params = %{
-  name: "valtimer",
-  active: true,
-  ...
-  ...
-}
-
-- 4.2 Create a default skin (if needed)
-charactername_basic_params = %{
-  is_default: true,
-  name: "Basic",
-  character_id: Enum.find(characters, fn c -> c.name == "charactername" end).id
-}
-- Add `charactername_basic_params` to the skins list
-Find the **Insert skins** section and add the new character's basic skin:
-
-Insert skins
-[
-  h4ck_fenix_params,
-  h4ck_basic_params,
-  valtimer_frostimer_params,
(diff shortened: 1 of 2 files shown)


apps/configurator/README.md
```diff
@@ -35,6 +35,37 @@ You can:
...
 
+Its mandatory to activate a character to have a basic skin assigned, for example, if you want to add Uren you should add this this lines of code:
+
+- Set `active` to `true` in `seeds.exs`
+charactername_params = %{
+  name: "valtimer",
+  active: true,
+  ...
+  ...
+}
```
Nico-Sanchez, line 46: A few things here:
- This README is for the Configurator app. It's documentation to add/edit characters via the forms in `localhost:4100`, not the seeds.
- You can either add a character via Configurator or by code using seeds.
The `active` param is the one that enables and disables characters, but it's already been said above:
> - Change it's `active` value to enable/disable characters in game.
- Also, you talk about Uren but the params are for Valtimer.

apps/configurator/README.md
```diff
@@ -35,6 +35,37 @@ You can:
...
+[
+  h4ck_fenix_params,
+  h4ck_basic_params,
+  valtimer_frostimer_params,
+  valtimer_basic_params,
+  kenzu_black_lotus_params,
+  kenzu_basic_params,
+  otix_corrupt_underground,
+  otix_basic_params
+]
```
Nico-Sanchez, line 67: Maybe we can replace all this by just saying that if you create a new character, you have to assign a default skin to it or it won't work, then show how to do it (like you do here).
Also, it would be nice to mention that this should be in the Configurator (again, these are Configurator's app docs), because now you can't create skins from it. If you add that, create an issue and link it here as well.

Nico-Sanchez (changes requested):

Nico-Sanchez:
> Key steps include:
> 
> Adding the character name in CharactersManager.cs.
> Activating the character in the configurator.
> Enabling the character's portrait in Unity.
> Setting up default skins in the backend if needed.

Also please update your PR description, because the steps you mention doesn't apply to this PR

Nico-Sanchez (changes requested):

mFragaBA (approved):
LGTM!

PR #1165, opened 2025-04-09
1155 add base attack and scale that and hp based of level

### LINKED ISSUE #1155: [FEATURE] Champion Progression - Stats per user & champion
When Levelling up, hero's will have their base HP and Power/Damage stat increased. We should change the current implementation so the hero stats can vary for each user.

mFragaBA:
> [!NOTE]
> To review this you'll need to use the branch from [1165 update messages proto and schemas #2764](https://github.com/lambdaclass/champions_of_mirra/pull/2764)

## Motivation

In order to add levelling up to a player, we need a way to improve the base stats of characters based on the level corresponding to the user. 

Closes #1155

## Summary of changes

We already had the `Unit` schema that coupled the user and the champion. It also contained the level field already.

- started using / introduced a `base_attack` that scales of the champion's level
- made health also scale off the champion's level
- used the `base_attack` to calculate the damage made by skills 
- modified the get_unit endpoint to also get the champion level for that champion and the provided user.
- started sending the champion level to the client. This was technically not required but happened as a consequence of handpassing data from the queue to the game updater. Considering we'll want this anyways in follow up PRs I decided to leave this.

Also changed the way damage is calculated: instead of it being 

$$
(1 + \mathit{powerUpDamageModifier} * \mathit{powerUpAmount} + \mathit{bonusDamage}) * \mathit{skill.damage}
$$ 

the 1 is replaced by the `base_attack` so it ends up being:

$$
(\mathit{baseAttack} + \mathit{powerUpDamageModifier} * \mathit{powerUpAmount} + \mathit{bonusDamage}) * \mathit{skill.damage}
$$

Also note that in this context, `base_attack` is not the one defined for each champion but instead is equal to $\mathit{champion.baseAttack} * (1 + \mathit{statIncreasePerLevel} * (\mathit{level} - 1))$

## How to test it?

Run the backend locally. Remember to start fresh by doing a `make start` or `devenv up` accordingly, based on which one you use.

By default, all champions start with base_attack = 1. To simulate "levelling up", you can change the base attack in the configurator for a champion you want.

Alternatively, you can run this on a iEx session to increase your character's level:

```elixir
GameBackend.Users.get_users_sorted_by_total_unit_prestige
```

should show something like:

```elixir
[
  %{
    username: "User_1",
    user_id: "20d01fe3-b4d1-4a3a-85dc-aa0bc8df3dbf",
    prestige: 22
  }
]
```

If there are more, locate the one corresponding to your user

Then get the unit you want to "level up" with:

```elixir
unit = GameBackend.Units.get_units("20d01fe3-b4d1-4a3a-85dc-aa0bc8df3dbf") |> Enum.find(fn unit -> unit.character.name == "kenzu" end)
```

And update the unit level with:

```elixir
unit |> Ecto.Changeset.change(%{level: 2}) |> GameBackend.Repo.update()
```

with level 10 you should be doing close to double damage

## Checklist


apps/arena/lib/arena/entities.ex
```diff
@@ -46,9 +50,11 @@ defmodule Arena.Entities do
...
       aditional_info: %{
         team: team,
-        health: character.base_health,
-        base_health: character.base_health,
-        max_health: character.base_health,
+        health: (character.base_health + character.base_health * @stat_increase_per_level * (level - 1)) |> round(),
+        base_health:
+          (character.base_health + character.base_health * @stat_increase_per_level * (level - 1)) |> round(),
+        max_health: (character.base_health + character.base_health * @stat_increase_per_level * (level - 1)) |> round(),
+        base_attack: character.base_attack + character.base_attack * @stat_increase_per_level * (level - 1),
```
Nico-Sanchez, line 57: Create a function to do this instead of doing it 4 times.

apps/arena/lib/arena/matchmaking/game_launcher.ex
```diff
@@ -115,6 +116,8 @@ defmodule Arena.Matchmaking.GameLauncher do
         client_id: client_id,
         skin_name: "Basic",
         character_name: Enum.random(characters).name,
+        # TODO: what do we do with bot levels? Average over all real player's levels
+        character_level: 1,
```
Nico-Sanchez, line 120: Nice question. Let's create an issue and link it here.

priv/repo/seeds.exs
```diff
@@ -1287,6 +1294,7 @@ uren_params = %{
   base_speed: 0.68,
   base_size: 100.0,
   base_health: 400,
+  base_attack: 1,
```
Nico-Sanchez, line 1297: Why do the characters need this parameter at all if its always 1?

Nico-Sanchez (changes requested):
I left some comments. Also, update arena's version due to change in protobuf messages.

priv/repo/seeds.exs
mFragaBA: now that you mention it we don't need this right now. I was adding the field so that tanks would have a smaller base attack and damage dealers would have a higher base attack but left it as 1 to be the same as main and avoid having to do balance changes.

apps/arena/lib/arena/matchmaking/game_launcher.ex
mFragaBA: created #1170

apps/arena/lib/arena/entities.ex
mFragaBA: done in [953d11f](https://github.com/lambdaclass/mirra_backend/pull/1165/commits/953d11f226c19c113e51ac4463f34b196326fe29)

mFragaBA:
> I left some comments. Also, update arena's version due to change in protobuf messages.

done in [6f7820c](https://github.com/lambdaclass/mirra_backend/pull/1165/commits/6f7820cbcf9b1c2a56c9627a0aec5bdd14b5c033)

apps/arena/lib/arena/entities.ex
```diff
@@ -50,11 +51,10 @@ defmodule Arena.Entities do
...
         team: team,
-        health: (character.base_health + character.base_health * @stat_increase_per_level * (level - 1)) |> round(),
-        base_health:
-          (character.base_health + character.base_health * @stat_increase_per_level * (level - 1)) |> round(),
-        max_health: (character.base_health + character.base_health * @stat_increase_per_level * (level - 1)) |> round(),
-        base_attack: character.base_attack + character.base_attack * @stat_increase_per_level * (level - 1),
+        health: scale_with_level(character.base_health, level) |> round(),
+        base_health: scale_with_level(character.base_health, level) |> round(),
+        max_health: scale_with_level(character.base_health, level) |> round(),
+        base_attack: scale_with_level(@character_base_attack, level),
```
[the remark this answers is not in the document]
mFragaBA: it was removed from the character, but it's used in game to calculate the damage. During the match it varies based on the champion's level so it's not a hardcoded value (like the `base_attack = 1` from `seeds.exs`)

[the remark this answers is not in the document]
Nico-Sanchez: Oh I see thank you.

apps/serialization/messages.proto
```diff
@@ -222,21 +222,22 @@ message Player {
...
+  optional bool forced_movement = 15;
+  optional bool bounty_completed = 16;
+  optional uint64 mana = 17;
+  optional uint32 current_basic_animation = 18;
+  optional uint32 match_position = 19;
+  optional uint32 team = 20;
+  optional uint64 max_health = 21;
+  map<uint32, Item> inventory = 22;
+  optional bool blocked_actions = 23;
+  optional string skin_name = 24;
```
Nico-Sanchez, line 240: Let's add at the end the character_level (24). It's a bad practice to move everything to make place for it.
https://protobuf.dev/best-practices/dos-donts/

Nico-Sanchez (changes requested):
Let's remove everything related to base_attack left in other files.

Nico-Sanchez (approved):

manucamejo (approved):

PR #1169, opened 2025-04-11
feat: add level up function in units module

### LINKED ISSUE #1156: [Feature] Champion Progression - Levelling Up
Implement hero levelling. For this we need to:

- provide an endpoint to level up
- provide a form on the configurator to change the cost and stat increase of each level
- define linear stat increase (5-10% increase per level over the base stat should be fine)

mFragaBA:
## Motivation

*Explain why not how.*
Closes #1156

> [!NOTE]
> Test this along side <TODO: create PR and link> from the client

## Summary of changes

*List your changes here.*

## How to test it?

## Checklist

### DIFF
--- a/apps/game_backend/lib/game_backend/units.ex
+++ b/apps/game_backend/lib/game_backend/units.ex
@@ -13,10 +13,14 @@ defmodule GameBackend.Units do
   import Ecto.Query
 
   alias Ecto.Multi
+  alias GameBackend.Utils
   alias GameBackend.Repo
+  alias GameBackend.Transaction
   alias GameBackend.Units.Unit
   alias GameBackend.Units.Characters.Character
   alias GameBackend.Units.UnitSkin
+  alias GameBackend.Users.Currencies
+  alias GameBackend.Users.Currencies.CurrencyCost
 
   @doc """
   Inserts a unit.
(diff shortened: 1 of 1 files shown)


FengariLuna (commented):
Tried the following steps and I cannot purchase the skin.
Since I get multiple issues on the backend, i add this as a comment instead of a request changes, since it could be a ME issue
Ive player naturally without touching the configurator

https://github.com/user-attachments/assets/eca51675-da0f-41a4-ad79-fa0dba7e3325

FengariLuna (dismissed):
tested the frontend and did not find any issues here nor there

apps/configurator/lib/configurator_web/live/level_up/form.html.heex
```diff
@@ -0,0 +1,29 @@
...
+    <h2>Level <%= level_info_f.data.level %></h2>
+    <.input field={level_info_f[:level]} type="hidden" value={level_info_f.data.level} />
+    <.input field={level_info_f[:stat_increase_percentage]} type="number" label="StatIncreasePercentage" />
+    <.inputs_for :let={currency_cost_form} field={level_info_f[:currency_costs]}>
+      <h3>Currency Cost</h3>
+      <.input field={currency_cost_form[:amount]} type="number" label="Amount" step="any" />
+      <.input field={currency_cost_form[:currency_id]} type="select" label="Currency" options={@currency_options} />
+    </.inputs_for>
+  </.inputs_for>
+  <!-- <.effect_form form={f} field={:effect} /> -->
```
Nico-Sanchez, line 19: What's this? Can we remove it?

apps/configurator/priv/repo/migrations/20250414200744_add_level_up_configuration.exs
```diff
@@ -0,0 +1,6 @@
+defmodule Configurator.Repo.Migrations.AddLevelUpConfiguration do
+  use Ecto.Migration
+
+  def change do
+  end
+end
```
Nico-Sanchez, line 6: What's this ghost migration?

apps/game_backend/lib/game_backend/units.ex
```diff
@@ -305,4 +308,92 @@ defmodule GameBackend.Units do
...
+
+        {:ok, %{unit: unit, user_currency: user_currency}} ->
+          {:ok, %{unit: unit, user_currency: user_currency}}
+      end
+    else
+      {:unit, {:error, :not_found}} -> {:error, :not_found}
+      {:can_level_up, false} -> {:error, :no_more_levels}
+      {:unit_owned, false} -> {:error, :not_found}
+      {:can_afford, false} -> {:error, :cant_afford}
+    end
```
Nico-Sanchez, line 352: This is pretty hard to follow. I can't see what's going on here.
Please refactor this to be like other controllers. A `with` `case` and `else` it's a worse case than the antipattern showed here: https://hexdocs.pm/elixir/code-anti-patterns.html#complex-else-clauses-in-with

apps/gateway/lib/gateway/controllers/curse_of_mirra/character_controller.ex
```diff
@@ -21,4 +21,35 @@ defmodule Gateway.Controllers.CurseOfMirra.CharacterController do
...
+          new_currency_balance: new_balances
+        })
+      )
+    else
+      {:error, :no_more_levels} ->
+        send_resp(conn, 400, "Cannot further level up")
+
+      {:error, :cant_afford} ->
+        send_resp(conn, 400, "Cannot afford levelling up")
+    end
```
Nico-Sanchez, line 49: Can we use the fallback controller instead of this? I don't know when those happen.
Also, we already have there an error clause for "cant_afford" cases.
I think a more understandable solution should be to check if the user has the enough currency to do this request, if that's ok, then it can level up.
That's how is being done everywhere, and that's why we use `with` + FallbackController, so we can split the requests in sequences of condition-checks then trigger the proper error accordingly.

Nico-Sanchez (changes requested):
The feature is working properly.
However, some parts of the code are hard to follow or need to be removed.

apps/gateway/lib/gateway/controllers/curse_of_mirra/character_controller.ex
mFragaBA: done! By this do you mean avoid doing the request from the frontend since it already knows you don't have enough money?

apps/configurator/lib/configurator_web/live/level_up/form.html.heex
mFragaBA: removed

apps/configurator/priv/repo/migrations/20250414200744_add_level_up_configuration.exs
mFragaBA: I think it got created with `mix ecto.gen migration` and I added it by mistake

apps/game_backend/lib/game_backend/units.ex
```diff
@@ -317,38 +317,35 @@ defmodule GameBackend.Units do
...
-           {:can_afford, Currencies.can_afford(user_id, costs)} do
-      result =
-        Multi.new()
-        |> Multi.run(:unit, fn _, _ -> add_level(unit) end)
-        |> Multi.run(:user_currency, fn _, _ ->
-          Currencies.substract_currencies(user_id, costs)
-        end)
-        |> Transaction.run()
+         {:ok, true} <- if(Currencies.can_afford(user_id, costs), do: {:ok, true}, else: {:error, :cant_afford}) do
+      Multi.new()
```
Nico-Sanchez, line 327: Let's use the same syntax as other endpoints. We always do:
```elixir
{:can_afford, true} <- {:can_afford, Currencies.can_afford(user_id, stuff_to_Buy)}
```

Nico-Sanchez (changes requested):

FengariLuna (approved):

PR #1176, opened 2025-04-16
1161 add buy skin behavior

### LINKED ISSUE #1161: [STORE] Add buy skin behavior

Nico-Sanchez:
> [!Warning]
> Merge alongside https://github.com/lambdaclass/champions_of_mirra/pull/2780

## Motivation

We need a way for users to unlock and buy skins
Closes #1161 

## Summary of changes

- [Add endpoint to buy a skin](https://github.com/lambdaclass/mirra_backend/commit/b9626a7e253a8a69cf05c3a3476cd47368edc855)
- [Add purchase costs to skins and update them in seeds](https://github.com/lambdaclass/mirra_backend/commit/eeed98348104f2cbf994fcfcaa7dd7cac620d51d)
- [Add endpoint to fetch skins from store](https://github.com/lambdaclass/mirra_backend/commit/3d327c0e5d6b8de35ed90ff547bf592b793c58f4)
- [Update functions for new buy_skin endpoint behavior](https://github.com/lambdaclass/mirra_backend/commit/f99702a46c4879ef014d238862a85ba036c29927)
- [Insert default skins only instead of all of them for users to buy them](https://github.com/lambdaclass/mirra_backend/commit/4c777677c6ab805544b2e424d115ba592183e0b3)
- [Update endpoint to send the updated user info instead of the bought skin id](https://github.com/lambdaclass/mirra_backend/commit/f70009ea558440f36d32043d2a8985880bcfb177) 

## How to test it?

1. Reset the DB and test if you can equip or buy a skin.
  - You shouldn't because you don't have the skin or currency enough to buy it.
2. Earn money, buy any skin.
3. Now you should be able to equip it.

To quick test this, you can edit the `quests_descriptions.json` to make them give you tons of money.

## Checklist

### DIFF
--- a/apps/game_backend/lib/game_backend/units/characters.ex
+++ b/apps/game_backend/lib/game_backend/units/characters.ex
@@ -235,7 +235,7 @@ defmodule GameBackend.Units.Characters do
 
   def get_skin_purchase_cost_by_currency(skin, currency_id) do
     purchase_cost =
-      Map.get(skin, :purchase_costs)
+      Map.get(skin, :purchase_costs, [])
       |> Enum.find(fn purchase_cost -> purchase_cost.currency_id == currency_id end)
 
     case purchase_cost do
--- a/apps/game_backend/lib/game_backend/units/characters.ex
+++ b/apps/game_backend/lib/game_backend/units/characters.ex
@@ -214,4 +214,20 @@ defmodule GameBackend.Units.Characters do
     |> Skin.changeset(attrs)
     |> Repo.insert()
   end
+
+  @doc """
+  Gets a Skin.
+  ## Examples
+      iex> insert_skin(%{field: value})
+      {:ok, %Skin{}}
+      iex> insert_skin(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+  """
+  def get_skin(id) do
+    Repo.one(Skin, id)
+    |> case do
+      nil -> {:error, :not_found}
+      skin -> {:ok, skin}
+    end
+  end
 end
--- a/apps/game_backend/lib/game_backend/units/characters.ex
+++ b/apps/game_backend/lib/game_backend/units/characters.ex
@@ -86,7 +86,7 @@ defmodule GameBackend.Units.Characters do
 
   ## Examples
 
-      iex> get_character(item_template_id)
+      iex> get_character(id)
       {:ok, %Character{}}
 
       iex> get_character(wrong_id)
@@ -220,17 +220,44 @@ defmodule GameBackend.Units.Characters do
   @doc """
   Gets a Skin.
   ## Examples
-      iex> insert_skin(%{field: value})
+      iex> get_skin_by_name(id)
       {:ok, %Skin{}}
-      iex> insert_skin(%{field: bad_value})
-      {:error, %Ecto.Changeset{}}
+      iex> get_skin_by_name(wrong_id)
+      {:error, :not_found}
   """
-  def get_skin(id) do
-    Repo.one(Skin, id)
+  def get_skin_by_name(skin_name) do
+    IO.inspect(skin_name, label: :aver_name)
+
+    Repo.get_by(Skin, name: skin_name)
     |> case do
       nil -> {:error, :not_found}
       skin -> {:ok, skin}
     end
+  end
+
+  def get_skin_purchase_cost_by_currency(skin, currency_id) do
+    purchase_cost =
+      Map.get(skin, :purchase_costs)
+      |> Enum.find(fn purchase_cost -> purchase_cost.currency_id == currency_id end)
+
+    case purchase_cost do
+      nil -> {:error, :not_found}
+      purchase_cost -> {:ok, purchase_cost}
+    end
+  end
+
+  @doc """
+  Receives a user_id, a skin_id and a list of CurrencyCosts.
+  Inserts new UserSkin from given Skin for given User.
+  Substract the amount of Currency to User by given params.
+  Returns {:ok, map_of_ran_operations} in case of success.
+  Returns {:error, failed_operation, failed_value, changes_so_far} if one of the operations fail.
+  """
+  def buy_skin(user_id, skin_id, purchase_costs_list) do
+    Multi.new()
+    |> Multi.run(:unit_skin, fn _, _ -> insert_unit_skin(%{user_id: user_id, skin_id: skin_id}) end)
+    |> Multi.run(:currencies, fn _, _ -> Currencies.substract_currencies(user_id, purchase_costs_list) end)
+    |> Repo.transaction()
   end
 
   def insert_unit_skin(attrs) do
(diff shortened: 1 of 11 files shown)


apps/game_backend/lib/game_backend/units/characters.ex
```diff
@@ -214,4 +216,72 @@ defmodule GameBackend.Units.Characters do
...
+    Repo.get_by(Skin, name: skin_name)
+    |> case do
+      nil -> {:error, :not_found}
+      skin -> {:ok, skin}
+    end
+  end
+
+  def get_skin_purchase_cost_by_currency(skin, currency_id) do
+    purchase_cost =
+      Map.get(skin, :purchase_costs)
```
mFragaBA, line 238: is `purchase_costs` always defined? if so we can just do `skin.purchase_costs`. If not, I think `Enum.find(nil, ...)` will fail so change it for `Map.get(skin, :purchase_costs, [])`

apps/game_backend/lib/game_backend/units/characters.ex
```diff
@@ -214,4 +216,72 @@ defmodule GameBackend.Units.Characters do
...
+  """
+  def buy_skin(%{user_id: user_id, skin_id: skin_id, unit_id: unit_id}, purchase_costs_list) do
+    curse_id = GameBackend.Utils.get_game_id(:curse_of_mirra)
+
+    Multi.new()
+    |> Multi.run(:unit_skin, fn _, _ -> insert_unit_skin(%{user_id: user_id, skin_id: skin_id, unit_id: unit_id}) end)
+    |> Multi.run(:currencies, fn _, _ -> Currencies.substract_currencies(user_id, purchase_costs_list) end)
+    |> Multi.run(:updated_user, fn _, _ -> GameBackend.Users.get_user_by_id_and_game_id(user_id, curse_id) end)
+    |> Repo.transaction()
+  end
```
mFragaBA, line 262: we should ensure the user does not already have the skin, I was able to double buy Kenzu's skin

apps/game_backend/lib/game_backend/units/characters.ex
mFragaBA, line 1: I'm noticing the characters are tied to a version but we never filter by version in these queries. Is this ok? Should we create a ticket to track this? No need to fix it in this PR because it seems rather big

mFragaBA (changes requested):
Should we do something when copying versions so we don't lose the skins with the prices?

apps/game_backend/lib/game_backend/units/characters.ex
Nico-Sanchez: Nice catch. Done!

apps/game_backend/lib/game_backend/units/characters.ex
Nico-Sanchez: You're totally right. Done!

apps/game_backend/lib/game_backend/units/characters.ex
mFragaBA: but on some queries we fetch from the `Character`/`Skin` schema (e.g. `list_skins_with_prices`, although we were doing it before this PR as well)

mFragaBA: I did a quick test by copying the starting version. After that:

- for previously existing users I didn't see any issues. I think this is because we can still fetch skins from previous versions
- when I try to login with a new guest user I get an error and there is an unhandled error on the backend:

```elixir
[error] #PID<0.4004.0> running Gateway.Endpoint (connection #PID<0.3740.0>, stream id 11) terminated
Server: localhost:4001 (http)
Request: POST /curse/users
** (exit) an exception was raised:
    ** (KeyError) key :skin not found in: nil

...
```

image

apps/gateway/lib/gateway/controllers/fallback_controller.ex
```diff
@@ -52,6 +52,14 @@ defmodule Gateway.Controllers.FallbackController do
...
   end
 
+  def call(conn, {:can_select_skin?, false}) do
+    send_resp(conn, 400, Jason.encode!(%{"error" => "user cannot select the skin"}))
+  end
+
+  def call(conn, {:already_bought_skin?, true}) do
+    send_resp(conn, 400, Jason.encode!(%{"error" => "user has the skin already"}))
+  end
+
```
mFragaBA, line 62: Do you have any opinions on having this controller handle all posible errors? I think these kinds of errors can be already caught in the corresponding controller (using the with-else clause) and we're just adding an indirection

apps/game_backend/lib/game_backend/units/characters.ex
mFragaBA: (I don't consider this blocking, we can further discuss in a separate issue / offline)

mFragaBA (changes requested):
I'm getting the wrong error I think:

image

apps/game_backend/lib/game_backend/units/characters.ex
Nico-Sanchez: This is already broken in main.

Nico-Sanchez:
> I'm getting the wrong error I think:
> 
> image

That's for the next steps issue https://github.com/lambdaclass/champions_of_mirra/issues/2781

apps/game_backend/lib/game_backend/units/characters.ex
mFragaBA: yes yes, that's why I mentioned it's an overall problem of how we handled skins/unit skins and not part of this PR and wanted to follow up the conversation on an issue

apps/gateway/lib/gateway/controllers/fallback_controller.ex
mFragaBA: yes, but why do we put all error explanations here instead of having each controller handle their errors? E.g. for `buy_skin` you could do:

```elixir
  def buy_skin(conn, params) do
    curse_of_mirra_id = GameBackend.Utils.get_game_id(:curse_of_mirra)

    with {:ok, skin} <- Characters.get_skin_by_name(params["skin_name"]),
         {:ok, unit} <- Units.get_unit_by_character_id(params["user_id"], skin.character_id),
         {:already_bought_skin?, false} <- {:already_bought_skin?, Units.has_skin?(unit, params["skin_name"])},
         {:ok, currency} <-
           Currencies.get_currency_by_name_and_game(params["currency_name"], curse_of_mirra_id),
         {:ok, purchase_cost} <-
           Characters.get_skin_purchase_cost_by_currency(skin, currency.id),
         {:can_afford, true} <- {:can_afford, Currencies.can_afford(params["user_id"], [purchase_cost])},
         {:ok, %{updated_user: updated_user}} <-
           Characters.buy_skin(%{user_id: params["user_id"], skin_id: skin.id, unit_id: unit.id}, [purchase_cost]) do
      send_resp(conn, 200, Jason.encode!(updated_user))
    else
      {:already_bought_skin?, true} -> 
        send_resp(conn, 400, Jason.encode!(%{"error" => "user has the skin already"}))
    end
  end
```

mFragaBA (approved):
LGTM!

let's create issues for the remaining things and continue discussion there

manucamejo (approved):

PR #1206, opened 2025-05-09
Shutdown websocket connection after game ended

### LINKED ISSUE #1205: Web socket connections not closed after game finishes
After a game ends, `Arena.GameSocketHandler` should terminate and the websocket connection should be ended. For real users, this happens as they exit the end game screen and that closes the connection. However, if you never exit, the socket connection is maintained. This is mostly significant since it will keep the socket connections for load test clients even after games end. It is also causing inconsistencies in metrics gathering/reporting as it's reporting more clients than there actually are. And it could cause problems if we want to do long running load tests.

mFragaBA:
## Motivation

We need to show in grafana more metrics, OS metrics, like CPU or memory usage.
Closes #1205 

## Summary of changes

- [Add os_mon dependency to monitor OS values](https://github.com/lambdaclass/mirra_backend/pull/1206/commits/5be55e95dc2ec5c0961a2b62f0304c8571854314)
- [Add new telemetry events for OS memory and cpu metrics](https://github.com/lambdaclass/mirra_backend/pull/1206/commits/14fd00412fdd59526b5c2839bf2e3679303a1652)
- [Update grafana dashboard: add cpu and memory panels](https://github.com/lambdaclass/mirra_backend/pull/1206/commits/297d3458d6c17f91ee5976791df4467332d46cb4)

When doing this, we spotted a bug where the players count wasn't working properly, it was never subtracting the disconnected players.
That is because the websocket connection was never closed until the client does so.
For that, we'll close the ws connection when (10 seconds after) the game ends, so we count the players accurately and the loadtests client can reconnect after finishing a match:
[Update grafana dashboard: add cpu and memory panels](https://github.com/lambdaclass/mirra_backend/pull/1206/commits/297d3458d6c17f91ee5976791df4467332d46cb4)

## How to test it?

Start the game and go to grafana, export the new dashboard.
You should start seeing OS cpu and memory metrics.
Also, the players count should be fixed.

## Checklist


Nico-Sanchez (approved):

mFragaBA (commented):
LGTM!

PR #1208, opened 2025-05-12
Update workflows and scripts due to grafana and loadtests

Nico-Sanchez:
## Motivation

We did a lot of changes in our infraestructure due to Grafana and Prometheus additions.
We need to update our Makefile and write some docs about it.

## Summary of changes

- [Update Makefile commands to setup prometheus + Grafana properly](https://github.com/lambdaclass/mirra_backend/commit/6c2974649ca664e91f36bd15cca25e85bc620ec9)
- [Add docs explaining the key-differences between Arena and Central servers, explaining Grafana and Prometheus responsibilities](https://github.com/lambdaclass/mirra_backend/commit/fb19c097a46b586cbb437afb2a0d98aac0d94108) 

## How to test it?

There's no need to test this because it was already done in our servers.
We can just try them out when needed and do the proper fixes there.
If you still want to do it, just run the makefile commands in a server. Be careful, it may break existing behavior.

## Checklist
- [x] This change requires new documentation.
  - [x] Documentation has been added/updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,9 +76,11 @@ server-specs:
 	./devops/server_specs.sh
 
 ## New server. Setup dependencies.
-## Run these as admin user
+## Run one of these (arena or central) as admin user
 
-admin-setup-arena-server: debian-install-deps setup-caddy create-env-file
+admin-setup-arena-server: debian-install-deps setup-caddy-arena create-env-file
+
+admin-setup-central-server: debian-install-deps debian-setup-grafana setup-caddy-central setup-prometheus create-env-file
 
 debian-install-deps:
 	sudo apt update -y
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -131,6 +123,81 @@ create-env-file:
...
+setup-central-caddy:
+	sudo ufw allow 80
+	sudo ufw allow 443
+	sudo truncate -s 0 /etc/caddy/Caddyfile && \
+	echo "$$(hostname).championsofmirra.com {" | sudo tee -a /etc/caddy/Caddyfile > /dev/null && \
+	echo "  reverse_proxy localhost:4000" | sudo tee -a /etc/caddy/Caddyfile > /dev/null && \
+	echo "}" | sudo tee -a /etc/caddy/Caddyfile > /dev/null && \
+	sudo sh -c 'echo "" >> /etc/caddy/Caddyfile'
+	echo "$$(hostname).grafana.championsofmirra.com {" | sudo tee -a /etc/caddy/Caddyfile > /dev/null && \
+	echo "  reverse_proxy localhost:4000" | sudo tee -a /etc/caddy/Caddyfile > /dev/null && \
```
mFragaBA, line 181: doesn't this have to be something different than port 4000? (on central testing europe I see port 3000 being used)

Makefile
```diff
@@ -131,6 +123,81 @@ create-env-file:
...
+	echo "    basic_auth:" >> /etc/prometheus/prometheus.yml && \
+	echo "      username: $$(PROMETHEUS_USER)" >> /etc/prometheus/prometheus.yml && \
+	echo "      password: $$(PROMETHEUS_PASSWORD)" >> /etc/prometheus/prometheus.yml && \
+	echo "    static_configs:" >> /etc/prometheus/prometheus.yml && \
+	echo "      - targets:" >> /etc/prometheus/prometheus.yml && \
+	cat /tmp/targets.txt >> /etc/prometheus/prometheus.yml && \
+	rm /tmp/targets.txt
+	sudo systemctl restart prometheus
+
+setup-central-caddy:
```
mFragaBA, line 172: aren't we missing the configurator reverse proxy? On central testing I see one

mFragaBA (changes requested):
Awesome! I left two questions but documentation is always nice to have

Makefile
Nico-Sanchez: You're right it's 3000

Makefile
Nico-Sanchez: Yup! I'll add it.

PR #1211, opened 2025-05-13
Add new broadcast update size metric and give Grafana panels some love

Nico-Sanchez:
## Motivation

Some Grafana panels in our dashboard are hard to understand.
The game tick panel isn't showing the right information.

## Summary of changes

- [Add new telemetry event to track size of broadcasted updates](https://github.com/lambdaclass/mirra_backend/commit/dbe4a6c2d3007d0898b9b897c85e6ed07b91db9f)
- [Fix game update tick metrics event](https://github.com/lambdaclass/mirra_backend/commit/1fc9ac05d3e0671332bb5dcf163236e38d14fc2a)
- [Update Grafana dashboard: show tick panel properly and add units for every panel](https://github.com/lambdaclass/mirra_backend/commit/71447bfcc310ab3394295dc9e7aa60d11a3a8515) 

## How to test it?

Just play a game and open your grafana dashboard (re-import the new one)!

## Checklist

### DIFF
--- a/apps/arena/lib/arena/prom_ex_plugin.ex
+++ b/apps/arena/lib/arena/prom_ex_plugin.ex
@@ -13,6 +13,7 @@ defmodule Arena.PromExPlugin do
         sum("arena.game.count", description: "Number of games in progress"),
         distribution("arena.game.tick.duration",
           description: "Time spent on running a game update tick",
+          unit: {:native, :millisecond},
           reporter_options: [buckets: [0, 5, 10, 30, 100, 1_000]]
         ),
         sum("arena.clients.count", description: "Number of clients (websockets) connected"),
--- a/apps/arena/lib/arena/prom_ex_plugin.ex
+++ b/apps/arena/lib/arena/prom_ex_plugin.ex
@@ -11,16 +11,9 @@ defmodule Arena.PromExPlugin do
     [
       Event.build(:game_metrics, [
         sum("arena.game.count", description: "Number of games in progress"),
-        ## TODO: this metric is an attempt to gather data to properly set the buckets for the distribution metric below
-        last_value("arena.game.tick.duration_measure",
-          description: "Last game tick duration",
-          unit: {:native, :nanosecond}
-        ),
-        ## TODO: Buckets probably need to be redefined, currently they all fall under the first bucket
         distribution("arena.game.tick.duration",
-          description: "Time spent on running a game tick",
-          unit: {:native, :nanosecond},
-          reporter_options: [buckets: [7_500_000.0, 15_000_000.0, 22_500_000.0]]
+          description: "Time spent on running a game update tick",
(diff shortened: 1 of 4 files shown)


apps/arena/lib/arena/prom_ex_plugin.ex
```diff
@@ -11,19 +11,16 @@ defmodule Arena.PromExPlugin do
...
-          description: "Last game tick duration",
-          unit: {:native, :nanosecond}
-        ),
-        ## TODO: Buckets probably need to be redefined, currently they all fall under the first bucket
         distribution("arena.game.tick.duration",
-          description: "Time spent on running a game tick",
-          unit: {:native, :nanosecond},
-          reporter_options: [buckets: [7_500_000.0, 15_000_000.0, 22_500_000.0]]
+          description: "Time spent on running a game update tick",
+          reporter_options: [buckets: [0, 5, 10, 30, 100, 1_000]]
```
mFragaBA, line 16: let's mention the unit here

grafana_arena_dashboard.json
```diff
@@ -1261,30 +1257,30 @@
...
         {
-          "datasource": {
-            "type": "prometheus",
-            "uid": "${DS_PROMETHEUS}"
-          },
           "disableTextWrap": false,
-          "editorMode": "builder",
-          "expr": "avg by(le) (arena_game_tick_duration_bucket)",
+          "editorMode": "code",
+          "expr": "rate(game_updater_broadcast_binary_size_sum[5m]) / rate(game_updater_broadcast_binary_size_count[5m]) * 30\n",
```
mFragaBA, line 1262: based on the name of the panel and its description I would expect this to be the estimate of the size of a single update, not an estimate of the amount of data sent per second (to each client)

mFragaBA (changes requested):
awesome work! Some tiny tweaks and we can merge

apps/arena/lib/arena/prom_ex_plugin.ex
Nico-Sanchez: Done

grafana_arena_dashboard.json
Nico-Sanchez: Ugh the per second thing is already in the panel. What name do you suggest?
Screenshot 2025-05-13 at 6 22 57 PM

mFragaBA: The data is an estimate bc we multiply by 30 (at least that's what I assumed from the code) the binary size of an update (assuming we always reach 30 updates per second) so I would mention that it's an estimate either in the name or description.

Alternatively, we can just emit the event on each update and show a last value of that (or the buckets one)

Nico-Sanchez: > Alternatively, we can just emit the event on each update and show a last value of that (or the buckets one)
If you do this, you only get 1 value while there are N >=0 active games.

mFragaBA (approved):

